Issue #377, opened 2022-09-28
Explore implementation of `num_updates` in scikit-learn with a one-by-one sample approach


srcansiz:
In GitLab by @sharkovsky on Sep 28, 2022, 11:40

As suggested by @pandrey-fr:
> This is indeed a limitation with scikit-learn, that occurs in most cases since most scikit-learn models run are designed to handle the entire training procedure based on a single fit call. This can however be avoided the specific case of SgdClassifier / SgdRegressor classes, because it is possible to set a constant arbitrary learning rate, feed the partial_fit method with a single data sample at a time, and reset its weights after each call. This way, it is possible to compute sample-wise, and therefore batch-wise, gradients. This is what we do in declearn, so that the same optimizer code (and plug-in system) can be used with these models as with torch and tensorflow.

PR #696, opened 2023-05-30
Draft: num updates

### LINKED ISSUE #369: Change optimizer to use num updates instead of epochs
In GitLab by @sharkovsky on Sep 22, 2022, 11:16

- [x] finalize decision on scikit-learn (see #371)
- [x] finalize discussion on batch_maxnum
- ~[ ] remove usage of epochs, we should only use `num_updates`~ We decided to also keep epochs for backward compatibility
- [x] Implementation for scikit-learn (see #377)

srcansiz:
In GitLab by @sharkovsky on Nov 23, 2022, 15:52

Closes #369 and #377.

This MR introduces a new training argument: `num_updates`. 
It is equivalent to the number of iterations in the training loop. For mini-batch SGD, this is equivalent to the number of model/optimizer updates, hence the name. 

For backward compatibility we still support using `epochs`, however we encourage using `num_updates` whenever possible. For this reason I have changed some notebooks, but not all. 

From a user point of view, I highly recommend reviewing the associated documentation MR fedbiomed.gitlabpages.inria.fr!66.

### **Major** changes in DataLoaders

DataLoaders have significantly different semantics now. In particular, both torch and numpy dataloaders now **iterate infinitely** over the dataset, building batches that are always of the same size. This is different from the previous behaviour, where dataloaders raised `StopIteration` at the end of one epoch, and the last batch could potentially be smaller than the specified batch size. 

You may find more information on the motivation and implementation details [here](https://notes.inria.fr/Bp7HeDM3QR6ftMuXiZK-Rw#).

Furthermore, I have factored out DataLoader-related implementation (for both torch and sklearn) in a separate `fedbiomed.common.data.loaders` module.

### Minor changes in TrainingArgs

We have decided that the following training arguments should have a default value of `None`. This is done to ensure two things:
1. we respect the original idea of TrainingArgs to assign a default value to all possible training arguments that a researcher might specify
2. we can keep track of whether the researcher specified a value for epochs, or for num updates

For more details, see this [Discord thread](https://discord.com/channels/762577487287812156/1044538782675910736).

### Minor changes in the training loop (TrainingPlan)

For both torch and scikit-learn, the training loop is no longer a nested loop (first over epochs, then over batches), but a single loop over the number of updates. This is explained more in detail [here](https://notes.inria.fr/Bp7HeDM3QR6ftMuXiZK-Rw#). 

#### batch maxnum

If the researcher specified epochs, they are allowed to specify the legacy training argument `batch_maxnum`. The meaning of this argument is "number of batches per epoch". For example, epochs=2 and batch_maxnum=3 leads to exactly 6 iterations, regardless of the dataset size.

#### Precedence rules

If the user specifies both epochs and num updates: 
1. num updates takes precedence
2. epochs will be ignored

For example, epochs=2, batch_maxnum=3, num_updates=17 leads to exactly 17 iterations, because num_updates takes precedence.

**Guidelines for MR review**

General:

* give a glance to [DoD](https://fedbiomed.gitlabpages.inria.fr/latest/developer/Fed-BioMed_DoD.pdf)
* check [coding rules and coding style](https://fedbiomed.gitlabpages.inria.fr/latest/developer/usage_and_tools/#coding-style)

Specific to some cases:


srcansiz:
In GitLab by @ybouilla on Nov 24, 2022, 14:49

Commented on [fedbiomed/common/data/loaders/utils.py line 7](https://github.com/fedbiomed/fedbiomed/compare/b6c3a4a46fe884b045b2f5b3db1ecc6dd2ba67f3..476d112b6f8f4773a09ff285717637ad08b05a56#diff-b32bdbdebf2c86fd73e90d1e7476e7b4R7)



regarding typing, you can use `Generator[Any]` from typing, which is more accurate than `Any`

srcansiz:
In GitLab by @ybouilla on Nov 24, 2022, 15:09

Commented on [fedbiomed/common/training_plans/_base_training_plan.py line 520](https://github.com/fedbiomed/fedbiomed/compare/b6c3a4a46fe884b045b2f5b3db1ecc6dd2ba67f3..476d112b6f8f4773a09ff285717637ad08b05a56#diff-feaa0c5898769613f5f0f30f4d2150fdR520)



It could be nice to detail what are the error raised and why in docstring

srcansiz:
In GitLab by @ybouilla on Nov 24, 2022, 15:12

Commented on [fedbiomed/common/training_plans/_base_training_plan.py line 542](https://github.com/fedbiomed/fedbiomed/compare/b6c3a4a46fe884b045b2f5b3db1ecc6dd2ba67f3..476d112b6f8f4773a09ff285717637ad08b05a56#diff-feaa0c5898769613f5f0f30f4d2150fdR542)



Have we decided what to do for that use case? in my MR, i was making sure num_updates would have been the same regardless of the dataset size on each node, if only `epochs` argument is provided. This is needed for aggregators like scaffold

srcansiz:
In GitLab by @sharkovsky on Nov 25, 2022, 19:04

Commented on [fedbiomed/common/training_plans/_base_training_plan.py line 542](https://github.com/fedbiomed/fedbiomed/compare/b6c3a4a46fe884b045b2f5b3db1ecc6dd2ba67f3..476d112b6f8f4773a09ff285717637ad08b05a56#diff-feaa0c5898769613f5f0f30f4d2150fdR542)



Now that I am reviewing your MR, I like your solution better. 
I will update this after we merge your MR (soon).

srcansiz:
In GitLab by @sharkovsky on Dec 21, 2022, 12:14

added 316 commits

<ul><li>c1582396...b6c3a4a4 - 308 commits from branch <code>develop</code></li><li>9fb9ad62 - Merge remote-tracking branch &#39;origin/develop&#39; into feature/369-num-updates</li><li>d044294b - Clean up num_updates handling in TrainingArgs</li><li>fa28b6e8 - Revisit computation of num_updates when epochs was specified</li><li>d5c47746 - Change semantics of num_updates when epochs is given</li><li>a6f60b73 - WIP add tests for MedicalFolderDataset-style shape</li><li>0fc0daa1 - Merge remote-tracking branch &#39;origin/develop&#39; into feature/369-num-updates</li><li>09807fed - Merge remote-tracking branch &#39;origin/develop&#39; into feature/369-num-updates</li><li>476d112b - misc improve comments</li></ul>

[Compare with previous version](/fedbiomed/fedbiomed/-/merge_requests/158/diffs?diff_id=223143&start_sha=c1582396a95a877167805daf897d759d47e9c054)

srcansiz:
In GitLab by @sharkovsky on Dec 21, 2022, 18:10

There is unfortunately a major source of incompatibility between Opacus and the current implementation. 

The issue is that Opacus bases a lot of their internal calculations on the assumption that iterations will be made on an epoch basis, and a new iterator will be instantiated at every new epoch. One notable example: they compute their sampling rate as `1/num_batches_per_epoch`.

This is not the case for the current proposed implementation, where we perform iterations based on the number of updates. This leads to all sorts of inconsistencies: for example, with their data loader it is impossible to perform more than `num_batches_per_epoch` updates, while we do not wish to impose such a limitation. Additionally, I strongly suspect our use of their tools may lead to wrong results (but not necessarily failures!) in the case of num_updates. 

So.... to be discussed. I am starting to think that the benefits of my proposed implementation (i.e. perfect handling of batch size even in-between epochs) may be outweighed by the issue of having to deal with the rest of the world still thinking in terms of epochs... I'm wondering whether I should revert these changes and implement something much simpler instead.

srcansiz:
In GitLab by @mvesin on Dec 21, 2022, 18:18

Important find you made @sharkovsky though it is unpleasant for now ...
At least we have a strong immediate reason to continue supporting epochs.

Maybe a minimal (ugly) implementation would be to impose using epochs when using opacus (then it means incompatibility with scaffold, for example).

Another option is the one you suggest (reverting the perfect fit of batch_size between epochs).

A third option might be to have 2 different modes for num_updates (the one existing, another one more compatible with epoch-based functions). Yet some functions (scaffold, opacus, ...) might not support all modes.

srcansiz:
In GitLab by @sharkovsky on Dec 21, 2022, 18:25

All options are valid, but they all make me :sob: 

I would be in favour of reverting my proposed implementation. 
Even though I liked it because I thought it was a smart idea, I don't think it's worth the trouble we are having... 

If I revert the changes, we will simply have to convert num_updates into epochs + a remainder, and learn to live with the fact that the last batch of each epoch may be incomplete. This corresponds to the "first idea" in my [notes](https://notes.inria.fr/Bp7HeDM3QR6ftMuXiZK-Rw#).

Issue #447, opened 2023-01-06
Use declearn optimizer on the researcher side (aggregation)


srcansiz:
In GitLab by @ybouilla on Jan 6, 2023, 18:48

Declearn optimizer provides a Researcher side optimizer (not provided in FedBioMed yet). Goals

Discuss with team the following points:
- **how to integrate Researcher optimizer in FedBioMed. What would be the API?**

In the poc, we have the following:
```python
from fedbiomed.common.optimizer import Optimizer
from fedbiomed.common.optimizer.modules import AdamModule, MomentumModule


researcher_opt = Optimizer(
    lrate=1.0,
    modules=[MomentumModule()],
)
exp = Experiment(
    tags=tags,
    model_args=model_args,
    training_plan=training_plan,  # assuming training_plan has been defined previously
    training_args=training_args,
    round_limit=rounds,
    aggregator=FedAverage(researcher_opt),  # pass the opt
    node_selection_strategy=None,
)
```
- **do we provide a way to add a PyTorch optimizer too on Researcher side?** : can be done thanks to `declearn` new feature
- **how to set `Scaffold` or other aggregators?**
in the poc, it is done this way:
```python
from fedbiomed.common.optimizer import Optimizer
from fedbiomed.common.optimizer.modules import ScaffoldServerModule, MomentumModule


researcher_opt = Optimizer(
    lrate=1.0,
    modules=[ScaffoldServerModule(), MomentumModule()],
)
exp = Experiment(
    tags=tags,
    model_args=model_args,
    training_plan=training_plan,  # assuming training_plan has been defined previously
    training_args=training_args,
    round_limit=rounds,
    aggregator=FedAverage(researcher_opt),  # pass the opt
    node_selection_strategy=None,
)

```
This syntax may be confusing because aggregator defined is `FedAverage`, and user wants to use `Scaffold`.
We can provide a `DeclearnScaffold` that inherits from `FedAvg` with pre-defined Optimizer

**IMPORTANT: make sure breakpoints feature is still working** -> add `get_state` and `save_state` methods

Issue #467, opened 2023-01-19
Implementing hyperparameters exchange  for declearn optimizers


srcansiz:
In GitLab by @ybouilla on Jan 19, 2023, 10:30

This task is about finding out an appropriate way to deal with declearn optimizers hyperparameters sent through the network.
In declearn, optimizer hyperparameters are sent through `collect_aux_var` method, and received / parsed through `process_aux_var` 

In FedBioMed, there are 2 ways to sent hyperparameters (more generally speaking information) between Nodes and Researcher:
1. through MQTT message broker (JSON format): this is for small messages/ piece of information, like sharing a few values 
2. through file system exchange: this is for large messages / piece of information, such as model parameters or scaffold 's correction state

The decision of how to send messages in FedBioMed should be defined by the developer somehow - it should be hard-coded, and not just based on the size of the piece of information exchanged

The goal of this task is to:
 - [x] decide which object has the responsibility to send such information (Aggregator class in Fed-BioMed?)
 - [x] find a way to detail in Fed-BioMed how parameters required by declearn optimizers should be sent - it could be a Enum class that specify which parameters are required, that is consulted before sending messages for instance
 - [x] create interface / wrapper to use declearn 's optimizers `collect_aux_var` and `process_aux_var`  methods
  - ~~[ ] make hyper-parameters exchange compatible with Secure Aggregation
Assignee can detail in this issue proposals that can be discussed and acknowledged by Fed-BioMed team~~ replaced by #837

ybouilla:
**About secure aggregation**:  auxiliary variables of declearn are exchanged decrypted, with may be an issue if user is using `Secure Aggregation`   or `DP`:

- for `Secure Aggregation`, a idea could be to create a declearn Scaffold warpper, that passes `Aggregator` as an argument (deltas sent back by `Nodes` to `Researcher` will be averaged)
- for `DP`, we should state clearly that for now, DP is only used on model weights (maybe add a warning message). 

For `Secure Aggregation`, if we go for a Scaffold-declearn warpper in Fed-BioMed, then we will have to choose with the normal way in declearn to use scaffold:

```
Optimizer(lr=.01, modules=[ScaffoldClientModule()])  # or the other way using strings
```
Will this be raising an error or work anyway (with/without encrypting auxiliary variables)

ybouilla:
Work for making Optimizers that rely on auxiliary variables that are compatible with secagg :

1. bout keys :
[...]

Issue #540, opened 2023-05-22
Refactoring Aggregator


srcansiz:
In GitLab by @ybouilla on May 22, 2023, 16:50

`Aggregator` class may need some refactoring: indeed, `Aggregator` responsibility is not very clear.
In this proposal, we commit that `Aggregator` responsibility is only about aggregating parameters received from Nodes on the Researcher. There is no optimization steps involved. An example of `Aggregator` is `FedAvg`, whereas fedprox and scaffold are not `Aggregator`, but rather `Optimizers` built upon`FedAvg`.

In the following issue, we propose a way to specify in the `Experiment` the `Aggregator` and the Researcher `Optimizer`:

- introduce a new class `AggregationPolicy` that will be responsible of  the `Aggregator` , the secure aggregation, and the `Optimizer`. Hence

```
aggregation_policy = AggregationPolicy(
          aggregation_method=FedAverage(),
          secagg=True,
          optimizer= Optimizer()

)

exp = Experiment(...,
        aggregator=aggregaton_policy,
                 ....)
       
```

- updates `Experiment` class wrt the changes

Please check !220 - https://gitlab.inria.fr/fedbiomed/fedbiomed/-/merge_requests/220#note_836500 for more details

**strategy refactoring**: please see
- #176
- #224

Issue #170, opened 2021-11-22
design Experiment() results query interface


srcansiz:
In GitLab by @mvesin on Nov 22, 2021, 09:43

Experiment results should include all what is needed to analyze an experiment after it completed.

Currently we access an experiment's results by browsing these data structures:
- `exp.aggregated_params` for researcher-computed results
- `exp.training_replies` for node-returned results
- `exp._rounds`, `exp._model_path` for experiment definition (not really part of the experiment results, but probably needed to use the experiment results)

What is lacking:
- no clear definition of what the results of an experiment include
- no class design

Proposed improvements
- [ ] define what scenarios we want to support for an experiment that impact content of results
  - eg: change `Strategy`, model, `Aggregator` during experiment ? (and keep record of the one used for each round)
- [ ] define precisely what is included in the results of an experiment
  - add from previous point
  - misc: add initial value of aggregated parameters ? add experiment call prototype ?
- [ ] define access conditions for an experiment result
  - eg: access by an `Experiment.results()` in a notebook
  - eg: later reload of results after `Experiment` is finished and deleted, maybe without instantiating an experiment. Eg.
```
res = Results(path_to_results_of this experiment)
res.results()
```
- [ ] define methods/interfaces to `Experiment` for querying results
- [ ] (optional) macro-choice for implementation. Eg how to store the results:
  - option 1: `Experiment` queries a `Strategy.results()` 
  - option 2: `Experiment` queries a `Results.strategy_results()`
  - etc.

PR #757, opened 2023-05-30
Feature/447 Add researcher-side (optional) Optimizer

srcansiz:
In GitLab by @pandrey-fr on May 12, 2023, 13:51

This MR adds the possibility to set up an Optimizer on the researcher-side, so as to refine aggregated model updates received from nodes prior to applying them to the global model. See related issue #447.

Adding an `Optimizer` on the researcher side enables the following features to be plugged into an Experiment:
- Set up some momentum or an adaptive optimizer, effectively implementing the so-called FedAvgM or FedOpt algorithms.
- Set up a server-side learning rate, some weight decay, or any of the previous algorithms in a modular way. Most notably, the choice of using such refinements is decoupled from the choice of aggregation rule, so that any combination of recipes inspired from the literature may be set up.
- Set up the Scaffold algorithm via the declearn backend:
  - Plug a `ScaffoldClient` module into the `Optimizer` used by nodes (defined in the experiment's `TraininPlan`).
  - Plug a `ScaffoldServer` module into the `Optimizer` used by the researcher (directly in `Experiment`).
  - This improves over the legacy implementation of Scaffold in the following ways:
      - It covers both scikit-learn and torch (and would cover tensorflow and jax/haiku if Fed-BioMed supported them).
      - It is more generic (notably enabling nodes to run distinct numbers of local optimization steps).
      - It is decoupled from the choice of aggregation rule (so one could for example use GradientMaskedAveraging and Scaffold together).
      - Its backend mechanisms may be re-used to implement other state-synchronization-based algorithms, such as FedDyn.
  - *Warning*: This does not work yet as this MR does not implement the sharing of Optimizer auxiliary variables (see issue #467).

This MR goes for the following implementation:
- Add an optional `researcher_optimizer` instantiation parameter to `Experiment`. By default, leave to None.
- Add the associate getter and setter methods.
- When a researcher optimizer is set, add the associated steps as part of the `Experiment.run_once` method.
- Include the researcher optimizer as part of the `Experiment` breakpoint system.

Closes issue #447


srcansiz:
In GitLab by @ybouilla on May 12, 2023, 18:08

Commented on [fedbiomed/researcher/experiment.py line 934](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01R934)



Here, I guess we should test the compatibility between node and researcher optimizer: they must be declearn optimizers

srcansiz:
In GitLab by @pandrey-fr on May 15, 2023, 08:43

Commented on [fedbiomed/researcher/experiment.py line 934](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01R934)



Not necessarily; this is only required when the researcher want to set up algorithms that require a pair of information-exchanging plug-ins (typically, Scaffold). Otherwise we can support using a declearn Optimizer on the researcher side (e.g. to set up a learning rate and/or momentum) and a framework-specific one on the node side.

That being said, I look forward to a point in the future when we change the API to set up the node-side optimizer, so that researchers (a) specify it at the same level as the researcher-side one (b) only use a (declearn-based) `Optimizer` (and can plug a framework-specific one into it if they really feel the urge for it - see our [`TorchOptiModule`](https://magnet.gitlabpages.inria.fr/declearn/docs/2.2/api-reference/model/torch/TorchOptiModule/)). Regarding the security model, it would be possible to have nodes review the configuration of their Optimizer, since it can be serialized into a human-readable JSON-compatible dict describing the hyper-parameters and plug-ins used.

srcansiz:
In GitLab by @pandrey-fr on May 16, 2023, 09:35

Commented on [fedbiomed/researcher/experiment.py line 134](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01R134)



Hello Sergen,

I agree with you that this question is not obvious, and should be discussed with the rest of the team.

What decided me to implement things this way, rather than going back to updating the `Aggregator` API, is the following:

- Updates' aggregation and application are two distinct, successive operations. In the original POC, we ended up with the optimizer-related code being defined at the ABC level, while sub-classes were only about defining the aggregation step. To me, it feels like shipping two bits of code in one, which is not optimal in terms of code readability (both for end-users and developers) nor maintainability (because it means that revising either feature is eager to have side effect on the other, for no good reason).

- If we were to refactor the two operations in a shared code structure that would then be plugged into `Experiment`, I would be in favor of using some composition, _i.e._ having some bit of code that receives (a) an `Aggregator` (b) an `Optimizer` and then operates on both, as a client of their respective APIs. This could probably be done as a mere backend refactoring effort to lighten up `Experiment` from a developer perspective, without altering the API for end-users.

- I think the `Aggregator` API would gain to undergo some revision, notably when we deprecate the legacy Scaffold implementation (which makes a confusion between aggregation and optimization), to remove unrequired bits and perhaps use the declearn `Vector` to abstract framework-dependent code. To be honest, I think it might be worth considering switching to the declearn `Aggregator` API - but that is of course an option, not an obligation. At any rate, in my opinion, revising `Aggregator` is going to be simpler if it truly is "only" about aggregation. Note that it might be worth re-thinking the articulation with SecAgg features, but given how unlearned I am about these, I cannot formulate a proper opinion, let alone a concrete design proposition.

- `Experiment` is indeed quite a huge chunk of code, the revision of which would probably be a nice effort towards developers' sanity (to be honest, every time I open it I want to cry - and the current tests make me want to quit programming altogether). Notwithstanding the discussion about whether getters/setters are utterly desirable (which I am not going to dive into; I would just note that when these getters/setters are properly written, they usually end up being very unitary and short bits of code, which means that they are easy to maintain and test), the `Experiment` backend is quite long, and sometimes quite convoluted. I think that part of it is inevitable (after all, it is the main entry-point for researcher end-users), but that is can mostly be improved, with some refactoring and/or rewriting efforts. That being said, the added bits of code as part of the present MR are quite short and unitary, hence although they do not solve the overall mess, I think they do not add to it, and might be illustrative of how refactoring things could make that mess more orderly, if still long or complex. This does not hold for unit tests, which I think should be rewritten entirely at some point, with an articulation of properly-mocked/patched unit tests, and integration tests.

Anyway, thank you for having a look at the code; I am happy to discuss more about the implementation choices, and just as open to rewriting things another way if the team decides on it in the end :smile:

srcansiz:
In GitLab by @pandrey-fr on May 16, 2023, 11:14

Commented on [fedbiomed/researcher/experiment.py line 134](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01R134)



Hi @srcansiz,

Then we are on the same page! I agree that implementing some composition container(s) that refactor `Experiment` into distinct bricks would be nice, and that grouping aggregation / secagg / optimization would make sense.

For some of the bricks they might be used to merely refactor backend code (preserving front-end init and run signatures), but for that specific one I agree with you that it could be worth making it public - notably because it would make end-user code easier to read through. I would even go further and say that in the longer term, we could consider moving the definition of the node-side optimizer from the TrainingPlan to such a facility.

srcansiz:
In GitLab by @srcansiz on May 16, 2023, 11:26

Commented on [fedbiomed/researcher/experiment.py line 134](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01R134)



There is one more thing that i want to propose if we decide to use the way that you've implemented currently. I would be favor changing the name `researcher_optimizer` to something else that does not refers to `researcher`, a name that indicates that it is a optimization step after the aggregation. I don't have a better idea but  more like `fed_optimizer`, or just `optimizer` or variant of those can be clearer. I don't know what kind of terminology other libraries use, we might check those as well to be coherent.

srcansiz:
In GitLab by @pandrey-fr on May 17, 2023, 13:27

added 114 commits

<ul><li>1e48af5a...305c3dc2 - 110 commits from branch <code>develop</code></li><li>4ed9e6d5 - Add researcher-side (opt.) optimizer.</li><li>157e51e9 - Add unit tests for researcher-side Optimizer use.</li><li>b1bae092 - Update &#39;FakeExperiment&#39; test support.</li><li>1ccf05d8 - Enable checkpointing the researcher-side Optimizer.</li></ul>

[Compare with previous version](/fedbiomed/fedbiomed/-/merge_requests/220/diffs?diff_id=254270&start_sha=1e48af5a86426ab98f22fd12c8785a2cd799d873)

srcansiz:
In GitLab by @pandrey-fr on May 17, 2023, 13:40

Commented on [fedbiomed/researcher/experiment.py line 134](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01R134)



I agree, `researcher_optimizer` is a long name, and may not be a very obvious one. I lack suggestions for a better one though; your `fed_optimizer` proposal makes sense to me.

- In declearn we use `client_opt` and `server_opt` to define the client-side and server-side optimizers (client/server being our equivalent of fbm's node/researcher). We think that this makes sense and is readable-enough as they are defined at the same level (and together with an aggregator).
- In flower the API is very different: optimization details are implemented either as part of the client-side training routines or shared `Strategy.aggregate_fit` method, which groups aggregation and post-processing. Hence there is no equivalent of providing a slot for an optimizer.
- PySyft's documentation is currently unreadable. I do not know how they organize, let alone name, things.

srcansiz:
In GitLab by @ybouilla on May 17, 2023, 16:14

Commented on [fedbiomed/researcher/experiment.py line 1560](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01L1560)



Here I am a bit confused: are you updating the global model coefs too? I mean the model in the training plan

Plus just to make sure, `step` call returns only `- gradient * learning rate`, correct?

for clarity sake, i would rename the first `agg_updates` to `gradients` or something similar

srcansiz:
In GitLab by @ybouilla on May 17, 2023, 16:14

I have a remark about user Fedbiomed Optimizer rather than generic_optimizer: We are using generic_optimizer on node side, I dont know but I feel it can make sense to use it on Researcher side too. I think doing so will ease users creating their own aggregators, so it looks like a code similar to pytorch, with a `step` and `zero_grad` method. What do you guys think?


Regarding tests: i think one that checks if optimizer is correctly saved as a breakpoint is missing) and can be reloaded with the same optimizer modules and optimizer parameters (learning rate, ...) from a breakpoint file.

srcansiz:
In GitLab by @ybouilla on May 17, 2023, 16:14

Commented on [fedbiomed/researcher/experiment.py line 2190](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01R2190)



I dont know if it is a good idea to give a random number to the optimizer file state on the researcher side, for saving a breakpoint

srcansiz:
In GitLab by @ybouilla on May 17, 2023, 16:14

Commented on [fedbiomed/researcher/experiment.py line 1560](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01L1560)



I guess Researcher optimization should belong to another function, so it makes `run_once` less messy

srcansiz:
In GitLab by @ybouilla on May 17, 2023, 16:14

Commented on [fedbiomed/researcher/experiment.py line 2189](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01R2189)



I think we should also save auxiliary variables that the optimizer contains (i understood it is not returned by the `get_state` method)

srcansiz:
In GitLab by @ybouilla on May 17, 2023, 16:14

Commented on [tests/test_experiment.py line 552](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-52d705e1ae8090413914aebb38308971R552)



the test you wrote is very nice, however, we can go one step forward by checking if optimizer ids are the same and that they have same parameters (learning rate, weight decay, ...)

srcansiz:
In GitLab by @pandrey-fr on May 17, 2023, 17:50

Commented on [tests/test_experiment.py line 1263](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-52d705e1ae8090413914aebb38308971R1263)



> what about a test that also tests gradient descent is applied accordingly

`Optimizer.step` already has its own tests that verify the correctness of the applied formula. Otherwise, you are talking about an integration test, which IMO is not in the scope of the current MR.

> and another one that check that if parameters have not been updated from one round to another, parameters are left unchanged

If parameters have not been updated, you are computing updates from zero-valued gradients. Depending on the `Optimizer` used, this may not always result in zero-valued updates (because of momentum and/or weight decay). At any rate, this is not in the scope of what this MR adds, but rather in the scope of `Optimizer` and its plug-in algorithms being correct - which is tested in other places.

srcansiz:
In GitLab by @pandrey-fr on May 17, 2023, 17:51

Commented on [fedbiomed/researcher/experiment.py line 1560](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01L1560)



> Plus just to make sure, `step` call returns only `- gradient * learning rate`, correct?

[As documented](https://gitlab.inria.fr/fedbiomed/fedbiomed/-/blob/feature/447-researcher-optimizer/fedbiomed/common/optimizers/optimizer.py#L112), it can do quite a few more things (add weight decay and/or run any plug-in algorithms, e.g. to add momentum), but overall, yes. And to answer your question: it does not have a side effect on the inputs, hence the model coefficients are not updated as part of this code.

> for clarity sake, i would rename the first `agg_updates` to `gradients` or something similar

The thing is, we treat the aggregated updates as though it were a gradient (and it is effectively an approximation of a gradient), hence my using this name. But I am okay with renaming to either `gradients` or `agg_gradients` depending on which you prefer.

srcansiz:
In GitLab by @pandrey-fr on May 17, 2023, 17:51

Commented on [fedbiomed/researcher/experiment.py line 2190](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01R2190)



As far as I understand, this is what is done for the model weight files (a link to which is then created as part of breakpointing), hence my doing that. But I am happy to change that if you have a suggestions for naming these more appropriately!

srcansiz:
In GitLab by @pandrey-fr on May 17, 2023, 17:51

Commented on [fedbiomed/researcher/experiment.py line 2189](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01R2189)



As I have already stated numerous times: absolutely not. Internal state variables are returned by `get_state`, while calling `get_aux` will (a) have undesired side effects (b) emit information that cannot be ingested back (as they are addressed to nodes).

srcansiz:
In GitLab by @pandrey-fr on May 17, 2023, 17:51

Commented on [tests/test_experiment.py line 552](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-52d705e1ae8090413914aebb38308971R552)



`self.assertIs(self.test_exp.researcher_optimizer(), optim)` - what more do you want?

srcansiz:
In GitLab by @pandrey-fr on May 17, 2023, 18:07

> We are using generic_optimizer on node side, I dont know but I feel it can make sense to use it on Researcher side too.

As we already discussed, I believe that this is a bad idea, that will only add complexity for the sake of no tangible feature gain - especially as the researcher-side optimizer does not operate on gradients produced by the model it could monitor, but on approximations of these gradients obtained from the aggregation of node updates - hence the Torch implementation would be very complex, and have even more overhead than what the `declearn.model.torch.TorchOptiModule` plug-in (which is compatible with the current implementation) generates.

> I think doing so will ease users creating their own aggregators, so it looks like a code similar to pytorch, with a `step` and `zero_grad` method.

This makes no sense to me. Aggregators do not have a step, nor gradients: they define a rule to aggregate a set of model weights. This is already extendable via the `Aggregator` API (although it is somewhat messy). And if you want a generic one that operates on Vector instances we could switch to the declearn `Aggregator` instead.

> Regarding tests: i think one that checks if optimizer is correctly saved as a breakpoint is missing) and can be reloaded with the same optimizer modules and optimizer parameters (learning rate, ...) from a breakpoint file.

This is true, and it would be nice to have such tests. But the thing is: the current tests for the breakpoint file are terrible in every way (they are both hard to read and very bad at actually testing anything, due to the abundance of patching), so I decidedly avoided spending time adding more mess to them for the sake of hacking the coverage metric.

srcansiz:
In GitLab by @mvesin on May 22, 2023, 11:35

Commented on [fedbiomed/researcher/experiment.py line 134](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01R134)



Thanks all for the discussion and arguments !

=> I also agree to merge the current implementation.

I share @pandrey-fr arguments that aggregation and optimization are independant, and that the initially proposed API (`Experiment( ... aggregator=FedAvg(optimizer=...) ...)`) is a tentative to partially refactor the API, while it should be later treated globally (not as part of this MR or milestone).

@pandrey-fr @srcansiz nice idea also to have a (name to find) `MasterAggregator` (`LearningPolicy` ?) in that future refactor. This could be a milestone by its own, that might include:
- researcher API redesign (experiment/strategy/aggregator), starting from a user's perspective (interact with researchers)
- backend redesign, including rewriting `Job` (part of the problems of `Experiment` ...), and maybe modularizing the breakpoint part out to an `ExperimentBreakpoint`. Another pending point was creating an `ExperimentResult` #207 #170 to handle the results save/load/format of an experiment.

srcansiz:
In GitLab by @mvesin on May 22, 2023, 11:48

>> Regarding tests: i think one that checks if optimizer is correctly saved as a breakpoint is missing) and can be reloaded with the same optimizer modules and optimizer parameters (learning rate, ...) from a breakpoint file.
> This is true, and it would be nice to have such tests. But the thing is: the current tests for the breakpoint file are terrible in every way (they are both hard to read and very bad at actually testing anything, due to the abundance of patching), so I decidedly avoided spending time adding more mess to them for the sake of hacking the coverage metric.

Though I agree with @pandrey-fr about all the limits of our current unit tests, especially in the case of breakpoints, I think like @ybouilla it would be nice to have unit test for that, as we agreed to at least unit test coverage for new code.

srcansiz:
In GitLab by @ybouilla on May 22, 2023, 14:29

Commented on [fedbiomed/researcher/experiment.py line 1560](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01L1560)



> > Plus just to make sure, `step` call returns only `- gradient * learning rate`, correct?

> [As documented](https://gitlab.inria.fr/fedbiomed/fedbiomed/-/blob/feature/447-researcher-optimizer/fedbiomed/common/optimizers/optimizer.py#L112), it can do quite a few more things (add weight decay and/or run any plug-in algorithms, e.g. to add momentum), but overall, yes. And to answer your question: it does not have a side effect on the inputs, hence the model coefficients are not updated as part of this code.

Ok so this is where I am confused: shouldn't it be
`aggregated_params = (init_params + agg_updates).coefs` then?

> The thing is, we treat the aggregated updates as though it were a gradient (and it is effectively an approximation of a gradient), hence my using this name. But I am okay with renaming to either `gradients` or `agg_gradients` depending on which you prefer.

I see, but since it is an approximation of a gradient what about naming it `approx_gradients` ? otherwise `agg_gradients` seems to be a good idea!

srcansiz:
In GitLab by @ybouilla on May 22, 2023, 14:29

Commented on [fedbiomed/researcher/experiment.py line 2189](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01R2189)



I see, you can get back the auxiliary variables just by knowing the states you are saving so it might be not necessary.
The initial idea of saving auxiliary variables in the breakpoint was that (a) we committed to save every piece of information before sending it to Nodes

srcansiz:
In GitLab by @ybouilla on May 22, 2023, 14:29

Commented on [fedbiomed/researcher/experiment.py line 2190](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01R2190)



I think for now we can keep as it is, since it wont introduce any bugs, but actually the idea behind random strings was more for model parameters, that were meant to be sent on the Node side. For this optimizer, since it is going to be kept on Researcher side, I think it doesnot need that IMHO

srcansiz:
In GitLab by @ybouilla on May 22, 2023, 14:29

Commented on [tests/test_experiment.py line 552](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-52d705e1ae8090413914aebb38308971R552)



Ok then, i was thinking of checking parameters that they havenot changed before and after passing optimizer through getter/setter, but maybe it is useless to test those

srcansiz:
In GitLab by @ybouilla on May 22, 2023, 14:29

> This makes no sense to me. Aggregators do not have a step, nor gradients: they define a rule to aggregate a set of model weights. This is already extendable via the `Aggregator` API

Actually, I don't think we all committed to this definition of `Aggregator`: but i think you are right about going towards that definition. Anyway, my remark was more about using `generic_optimizer` over Fed-BioMed `Optimizer`: we could create a new method in `generic_optimizer` that updates the optimizer on `Researcher` side (something like `researcher_step`). Actually, it is more about writing the following in the `generic_optimizers`:
```
init_params = Vector.build(self._global_model)
agg_updates = init_params - Vector.build(aggregated_params)
agg_updates = self._global_optim.step(agg_updates, init_params)
aggregated_params = (init_params - agg_updates).coefs
```
In addition i think everything related to optimization should be handled by `generic_optimizer` class

srcansiz:
In GitLab by @srcansiz on May 22, 2023, 15:08

Hi @ybouilla,

For me taking the code snippet that you shared and putting it in `GenericOptimizer` as a method does not make any difference. However, IF `GenericOptimizer` by itself avoids building model parameters as `declearn.Vector` again while it can be done in the model class (meaning that if this part is an unnecessary duplication), using `GenericOptimizer` might be useful in terms of simplicity.

srcansiz:
In GitLab by @ybouilla on May 22, 2023, 15:18

> For me taking the code snippet that you shared and putting it in `GenericOptimizer` as a method does not make any difference.

This is more about responsibility: for me it makes sense that generic Optimizer handles optimization, regardless if it happens on node or researcher side

> However, IF `GenericOptimizer` by itself avoids building model parameters as `declearn.Vector` again while it can be done in the model class (meaning that if this part is an unnecessary duplication), using `GenericOptimizer` might be useful in terms of simplicity.

We are already building the model, since we call `self._global_model = self._job.training_plan.get_model_params()`. so i don't think there is any redundancies

srcansiz:
In GitLab by @mvesin on May 22, 2023, 15:42

Commented on [fedbiomed/researcher/experiment.py line 2190](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01R2190)



I would also be in favor of keeping @pandrey-fr implementation with random string in the file name.

The reason for including random string is to avoid any file conflict. This is needed in many situations, including pure researcher-side files.

In this particular case, with current implementation, it looks like: we use 1 breakpoint dir per experiment and per round, we have only 1 optimizer per experiment at a given round. So there should not be conflict. Nevertheless, using random number makes implementation more robust and future proof (eg: if we re-implement breakpoint files layout).

srcansiz:
In GitLab by @ybouilla on May 22, 2023, 16:20

Commented on [fedbiomed/researcher/experiment.py line 1560](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01L1560)



By the way, should we specify `self._global_optim.init_round()`?
I think it is useless because it doesnot make sense to have `Regularizers` on Researcher side, but just wanted to make sure that you agree before moving forward

srcansiz:
In GitLab by @pandrey-fr on May 22, 2023, 17:38

Commented on [fedbiomed/researcher/experiment.py line 1560](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01L1560)



> Ok so this is where I am confused: shouldn't it be `aggregated_params = (init_params + agg_updates).coefs` then?

~~Nope - and I went out of my way to explain why (unless I did some very basic maths error somewhere, which is always a possibility, so thank you for proof-reading what follows).~~

Yes, indeed. My bad for forgetting that `Optimizer.step` already switches the sign of inputs to output updates that are to be added to the initial weights rather than substracted... Sorry. Here is, however, a long explanation on the maths at hand, that might still be useful:

For each node $`i`$ , you have (schematically, i.e. notwithstanding plug-in modifications to the vanilla stochastic SGD) : $`\theta_i^{t+1} = \theta^t - \sum_{k=1}^K \eta_{i,t,k} \nabla_{i,t,k}`$, where $`\eta_{i,t}`$ is a learning rate and $`\nabla_{i,t,k}`$ is a simplified way to denote the (stochastic) gradients of a loss function wrt the model weights, computed at a given step based on the current local weights and a batch of samples. I introduce the notation $`\hat{\nabla}_i^t := \sum_{k=1}^K \eta_{i,t,k} \nabla_{i,t,k}`$ for the sake of simplifying what follows.

In Federated Learning, we classically use $`\frac{1}{I} \sum_{i=1}^I \hat{\nabla}_i^t`$ as an estimator of the gradients of a loss function wrt the shared weights $`\theta^t`$ - and then use these stochastic gradients to update these weights, similar to a stochastic SGD step. In vanilla FedAvg, the learning rate is 1.0 and there is no further refining of the updates: $`\theta^{t+1} = \theta^t - \frac{1}{I} \sum_{i=1}^I \hat{\nabla}_i^t`$. This is what Fed-BioMed currently does, taking advantage of the fact that $`\theta^t - \frac{1}{I} \sum_{i=1}^I \hat{\nabla}_i^t = \sum_{i=1}^I (\theta^t - \hat{\nabla}_i^t) = \sum_{i=1}^I \theta_i^{t+1}`$. Hence, for this specific case, the updated global weights are the average of the node-wise ones. Note that I let apart the weighting of nodes' updates for simplicity - it would require adding some terms into the formula, but would not alter the points that are of interest for this discussion.

A more generic version of what precedes is to start back from $`\theta^{t+1} = \theta^t - \frac{1}{I} \sum_{i=1}^I \hat{\nabla}_i^t`$, and allow to plug into this formula any modification that builds on vanilla SGD. For example, one might add a learning rate, some momentum (FedAvgM), or even an adaptive algorithm (FedOpt in its various declinations: FedAdagrad, FedAdam, FedYogi...). This is precisely what the use of the declearn-based `Optimizer` enables: but for that, its specification requires that the optimizer instance receives only $`\frac{1}{I} \sum_{i=1}^I \hat{\nabla}_i^t`$ - which may in fact be replace by any other aggregation rule (for example a median rather than mean; or an algorithm that scales the outputs' magnitude based on consensus; etc.). Hence my point that these are two very distinct steps: first there is an aggregation rule to compute $`\hat{\nabla}^t := agg(\{\hat{\nabla_i^t}\}_{i=1}^I)`$, then there is an optimization rule to compute the updates to the global weights : $`\theta^{t+1} = \theta^t + opt(\hat{\nabla}^t)`$. **Edit**: I corrected the sign in the last formula, because in our case `opt` is designed to switch it.

Hence, in my (edited) implementation, I do the following:
- let existing code compute $`\textnormal{aggregated\_params} = \theta^t - \hat{\nabla}^t`$ (in fact always using some averaging: $`\frac{1}{I} \sum_{i=1}^I (\theta^t - \hat{\nabla}_i^t)`$)
- compute $`\hat{\nabla}^t = \theta^t - \textnormal{aggregated\_params}`$
- compute $`\textnormal{agg\_updates} = opt(\hat{\nabla}^t)`$
- go back to $`\theta^t + opt(\hat{\nabla}^t)`$ to let existing code apply the update

Note that I would be in favor of revising the `Aggregator` API so that it receives only the node-wise updates rather than their weights - notably because some aggregations rules would not be applicable in the current versions. At which point some of steps I take could be removed; but for the time being, I tried to make my modifications as minimal as I could.

> I see, but since it is an approximation of a gradient what about naming it `approx_gradients` ? otherwise `agg_gradients` seems to be a good idea!

I would rather stick to `agg_gradients`, because the important point to understand the logic of this step if that aggregation has already occurred.

> By the way, should we specify `self._global_optim.init_round()`? I think it is useless because it does not make sense to have `Regularizers` on Researcher side, but just wanted to make sure that you agree before moving forward

You are right, although it is highly unlikely for regularizers to be used, this method should be called. I am going to add that.

srcansiz:
In GitLab by @pandrey-fr on May 22, 2023, 17:40

Commented on [fedbiomed/researcher/experiment.py line 2189](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01R2189)



For the n-th time: I know, and every bit of information _is_ saved in the current version. The specs for checkpoints is to enable resuming an experiment at the very start of a given round, returning to the state things were at the very and of the previous ones. The current implementation for `Optimizer` and for this MR abides by this.

srcansiz:
In GitLab by @pandrey-fr on May 22, 2023, 17:46

Commented on [tests/test_experiment.py line 552](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-52d705e1ae8090413914aebb38308971R552)



Okay, this makes more sense to me. I think it is entirely superfluous (it means writing tests that are way longer than a couple of methods that are literally about assigning to a variable and accessing it), but I can do it.

srcansiz:
In GitLab by @pandrey-fr on May 23, 2023, 10:14

Commented on [fedbiomed/researcher/experiment.py line 1560](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01L1560)



@ybouilla My bad, I had the maths wrong regarding the updates' sign... Thank you for your vigilance! I have now corrected the error, and refactored the code, introducing your suggested changes.

srcansiz:
In GitLab by @pandrey-fr on May 23, 2023, 10:24

Commented on [tests/test_experiment.py line 1263](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-52d705e1ae8090413914aebb38308971R1263)



> True ( i was thinking about a simple optimizer with no w_decay or momentum) but here the idea is more to test the correct update of Researcher optimizer, which has not been tested yet I think

I understand, but to me testing this specific edge case does not provide actual value, and would just add some runtime and maintenance cost to the tests. To me, if we truly want to check that the maths are correct (in addition to testing that the code runs, which is already tested), we should rely on an integration test, using some known dataset and learning task to verify overall convergence of the process.

> This is arguable, i understand your point but I think it is important to be tested nevertheless, to check the correct computation of the "gradients" (Reseracher wise). Note that this might be tested separately if we go to a separate method to compute researcher updates

My point is that:
- `test_experiment_27_run_once_with_optimizer` already tests that the code can be properly run, and results in the expected call to `Optimizer.step`.
- `Optimizer.step` has its own tests that verify that the maths are correct.

At this point, the only additional thing we could do is to hack further into mock objects so that we can verify that the assigned new weights are equal to our expected formula. But the thing is, this will only be equivalent to writing our code a second time: it would not provide a guarantee that the maths are correct. In my opinion, the only way to check that the maths are correct is to go for an integration test. Otherwise, the unit test can only be about verifying that inputs run properly through the expected bits of code - which they do in their current form.

All of that being said, if you have a suggestion for a better version of the current test, feel free to write it!

srcansiz:
In GitLab by @pandrey-fr on May 23, 2023, 10:29

Commented on [tests/test_experiment.py line 552](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-52d705e1ae8090413914aebb38308971R552)



Oh, in fact, no, I cannot do it. This is a Mock object, not an actual Optimizer. That being said, there is no way to modify an Optimizer instance's configuration without accessing its private attributes (or replacing it with another instance - which we know not to happen via the `assertIs` check).

srcansiz:
In GitLab by @pandrey-fr on May 23, 2023, 10:32

added 1 commit

<ul><li>9f8dce17 - Refactor researcher-side Optimizer backend code and fix the maths.</li></ul>

[Compare with previous version](/fedbiomed/fedbiomed/-/merge_requests/220/diffs?diff_id=254681&start_sha=961b0afbebf9e3d3dc21fe58b8825feec94593a8)

srcansiz:
In GitLab by @ybouilla on May 23, 2023, 10:48

Commented on [tests/test_experiment.py line 552](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-52d705e1ae8090413914aebb38308971R552)



If you feel it is unnecessary  then it is ok to leave it as it is. I don't want you to do extra work, especially if you think it does not add too much

srcansiz:
In GitLab by @ybouilla on May 23, 2023, 10:48

Commented on [tests/test_experiment.py line 1263](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-52d705e1ae8090413914aebb38308971R1263)



OK I will try to write it! I was thinking about adding a new test. Let me know if it pleases you

srcansiz:
In GitLab by @ybouilla on May 23, 2023, 10:48

Commented on [fedbiomed/researcher/experiment.py line 134](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-d668ccb0192379eee259e5881add9c01R134)



just opened issue #540 so we keep track of this relevant discussion somewhere else

srcansiz:
In GitLab by @ybouilla on May 23, 2023, 11:05

Commented on [tests/test_experiment.py line 1598](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-52d705e1ae8090413914aebb38308971L1598)



In this test, we are testing complete saving of a breakpoint. I think it could be a good idea to update this test by adding an entry for Researcher Optimizer

srcansiz:
In GitLab by @pandrey-fr on May 23, 2023, 11:14

Commented on [tests/test_experiment.py line 1598](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-52d705e1ae8090413914aebb38308971L1598)



This test is almost entirely fake, because nearly all points of the saving process is patched or mocked. You are merely testing that the collection happens as expected on private attributes (which feels awkward to me), and that the standard lib JSON works for the values you set... with no guarantees that they match real-life values. But okay, I will add up to the mess and cover the global optimizer.

srcansiz:
In GitLab by @pandrey-fr on May 23, 2023, 11:34

added 1 commit

<ul><li>364af7a5 - Add test for researcher optimizer breakpointing.</li></ul>

[Compare with previous version](/fedbiomed/fedbiomed/-/merge_requests/220/diffs?diff_id=254742&start_sha=9f8dce1767b99620c173102f2c388f1257ad5b89)

srcansiz:
In GitLab by @pandrey-fr on May 23, 2023, 11:36

Commented on [tests/test_experiment.py line 1598](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-52d705e1ae8090413914aebb38308971L1598)



Done.

srcansiz:
In GitLab by @ybouilla on May 23, 2023, 18:02

added 21 commits

<ul><li>364af7a5...2e3d5a5f - 11 commits from branch <code>develop</code></li><li>0b24f788 - Add researcher-side (opt.) optimizer.</li><li>cca13b30 - Add unit tests for researcher-side Optimizer use.</li><li>5727f53b - Update &#39;FakeExperiment&#39; test support.</li><li>1e48af5a - Enable checkpointing the researcher-side Optimizer.</li><li>5a23fb6c - Merge branch &#39;feature/447-researcher-optimizer&#39; of...</li><li>6d6d60e8 - Merge branch &#39;develop&#39; of https://gitlab.inria.fr/fedbiomed/fedbiomed into...</li><li>a5cbd01d - Merge branch &#39;feature/447-researcher-optimizer&#39; of...</li><li>1d5577c1 - Merge branch &#39;develop&#39; of https://gitlab.inria.fr/fedbiomed/fedbiomed into...</li><li>b0c11418 - adding test that tests correct update of Optimizer on Researcher side</li><li>23ba6f4b - Merge branch &#39;feature/447-researcher-optimizer&#39; of...</li></ul>

[Compare with previous version](/fedbiomed/fedbiomed/-/merge_requests/220/diffs?diff_id=254993&start_sha=364af7a52df57d1c6dd21e777edb4748b78ca15e)

srcansiz:
In GitLab by @ybouilla on May 23, 2023, 18:03

Commented on [tests/test_experiment.py line 1263](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-52d705e1ae8090413914aebb38308971R1263)



done! Let me know if you like it or want me to revert it back

srcansiz:
In GitLab by @pandrey-fr on May 24, 2023, 09:07

Commented on [tests/test_experiment.py line 1263](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-52d705e1ae8090413914aebb38308971R1263)



Thanks Yannick! I have mixed feelings. On the one hand, I understand what you want to achieve, and the test seems correct. On the other, it is very contextual / does not feel like a robust unit test to me, in the sense that you are testing a private method, meaning that any purely-backend revision is likely to break the test regardless of whether the behavior of the public methods relying on it remain correct.

I feel like we could do without this test, mostly because the correctness of the Optimizer-wrapped operations is already assessed elsewhere. But we can also do with it, and I agree that there is some sense in assessing the maths' correctness. Hence I would be in favor of keeping the code you wrote for now, and reconsider it once we further revise Experiment. Does that seem good to you?

srcansiz:
In GitLab by @ybouilla on May 24, 2023, 10:03

Commented on [tests/test_experiment.py line 1263](https://github.com/fedbiomed/fedbiomed/compare/1aae048fccd39bf6f98949d827dba9fb20754cfa..65e3a49587b350a8c418eb027f5bb367389ac971#diff-52d705e1ae8090413914aebb38308971R1263)



I also share the same feelings! Like you, I dont really like the idea of testing private methods, but it is nice to have a test that checks the maths! My first intention was to test `Experiment.run_once` method, but then I remembered that `Experiment` should be refactored, so I went for the easiest solution: testing only this private method (I commented out my feelings about that in the code).

srcansiz:
In GitLab by @ybouilla on May 24, 2023, 10:03

Hey Paul, thanks again for this MR ! 
I think we covered everything, there are still some refactoring problems but they go beyond the scope of this MR.

So I think we can merge your work. Please let me know if you want to add some changes, and "un-draft" the MR, before moving forward

srcansiz:
In GitLab by @pandrey-fr on May 24, 2023, 10:34

resolved all threads

srcansiz:
In GitLab by @pandrey-fr on May 24, 2023, 10:35

Thank you Yannick! I am going to "un-draft" then, and we can delay the addition of auxiliary variables exchange to a distinct MR.

srcansiz:
In GitLab by @pandrey-fr on May 24, 2023, 10:35

marked this merge request as **ready**

srcansiz:
In GitLab by @ybouilla on May 24, 2023, 17:26

added 3 commits

<ul><li>23ba6f4b...1aae048f - 2 commits from branch <code>develop</code></li><li>65e3a495 - Merge branch &#39;develop&#39; of https://gitlab.inria.fr/fedbiomed/fedbiomed into...</li></ul>

[Compare with previous version](/fedbiomed/fedbiomed/-/merge_requests/220/diffs?diff_id=255256&start_sha=23ba6f4b8d498716f2faaa17d994a7ee3d62fa6a)

srcansiz:
In GitLab by @mvesin on May 24, 2023, 17:28

Jenkins please retry a build

srcansiz:
In GitLab by @ybouilla on May 25, 2023, 09:43

Just checked that current MR is working with secure aggregation!

ybouilla:
**Update**: according to a poll done within the dev team, we should rename `researcher_optimizer` into `agg_optimizer` 
Result of the poll:

- `fed_optimizer` 2 votes
- `global_model_optimizer`: 2 votes
- `agg_optimizer`: 5 votes

pandrey-fr:
> we should rename researcher_optimizer into agg_optimizer

Done!

fedbiomed/researcher/experiment.py
```diff
@@ -1589,6 +1643,48 @@ def run_once(self, increase: bool = False, test_after: bool = False) -> int:
...
+            self._job.training_plan.get_model_params(only_trainable=True)
+        )
+        init_params = Vector.build(
+            {k: v for k, v in self._global_model.items() if k in names}
+        )
+        agg_gradients = init_params - Vector.build(
+            {k: v for k, v in aggregated_params.items() if k in names}
+        )
+        # Take an Optimizer step to compute the updates.
+        # When using vanilla SGD: agg_updates = - lrate * agg_gradients
```
ybouilla, line 1682: Nice addition! But I think this could also be tested to make sure non-trainable layers are not updated during the optimization. I can take care of that if you want

ybouilla: Great! Thanks for the test, I think it covered well the use case depicted!

PR #788, opened 2023-06-06
Feature/467 Implement exchange of Optimizer auxiliary variables

pandrey-fr:
This PR tackles issue #467 by implementing the (optional) exchange of auxiliary variables between the node-side and researcher-side Optimizer instances (provided they both exist).

The implementation was done incrementally, adding (in this order):
- Collection and uploading of auxiliary variables by Experiment / Job
- Downloading and processing of the former information by Node / Round
- Collection and uploading of auxiliary variables by Round
- Processing of the former information by Experiment (received as part of training Responses).

Task-list:
- [x] Implement the full cycle of Optimizer auxiliary variables
- [x] Patch code and/or existing unit tests so that they pass
- [ ] Run some integration tests for Scaffold using Optimizer (decide whether to publish them)
- [x] Write (or extend) unit tests for the new features
  - [x] Experiment
  - [x] Job
  - [x] Node & TrainRequest
  - [x] Round
- [ ] Ensure SecAgg use can be added in the near future, and related information is collected in a dedicated issue


ybouilla:
Hi Paul, thanks for this Pull Request!

I have seen in the code that auxiliary variables are serialized in MsgPack whatever its nature is (heavy or light parameters size - wise). In the previous work when integrating Scaffold, we tried to abide with the following principle:
- if parameter is a scalar or  vector, send parameter through MQTT
- otherwise if it is rather heavy (like correction term in scaffold), use a file transfer

I agree that for a first version we may want to take the easiest solution  as you did, which converts everything into MsgPack when it comes to auxiliary variables

fedbiomed/researcher/job.py
```diff
@@ -475,22 +495,107 @@ def start_nodes_training_round(self,
...
+        Returns:
+            url_shared: url of a file containing auxiliary variables shared
+                across all nodes, or None (in the absence of such information).
+            url_bynode: dict mapping urls of files containing node-specific
+                auxiliary variables to the nodes' id (a missing `nodes` key
+                indicates that this node has no such information to receive).
+        """
+        # Split the information between shared and node-wise dictionaries.
+        aux_shared, aux_bynode = self._prepare_agg_optimizer_aux_var(
+            optimizer=optimizer, nodes=list(self._nodes)
```
ybouilla, line 536: I got a bit confused here with `aux_shared` and `aux_bynode`, is `aux_shared` really needed here? 
Is it possible to use only `aux_bynode`, and if a aux_var needs to be  broadcasted to all Nodes, we assign same variable to all node_id in the dictionary.
That way code will be slightly convoluted (unless there is something I havenot seen)

pandrey-fr: The idea is that the auxiliary variables sent by the researcher may either be shared across all nodes (e.g. if we implement a plug-in to adjust the node-side learning rate based on the global (aggregated) loss, or if we share the global state $c_t$ in Scaffold), or specific to each node (e.g. if we share the local states $c_{i,t}$ or the correction terms $\delta_{i,t}$ in Scaffold).

In the latter case (node-specific), we want to upload them in distinct files, so that each node downloads only their own information (both to avoid unnecessary communication costs and to prevent privacy leakage).

In the former case (shared across nodes), my implementation isolates all of this information in a dedicated file, that is uploaded only once by the researcher, and downloaded by each and every node. As a result, we save on communication costs related to the file upload.

That being said, we could do something that is more costly, but simpler, which is to merely upload one file per node, regardless of the redundancy of information between these files.

pandrey-fr:
> I think we can add this task to this PR as discussed:
> 
>     * [  ]: find a way to check synchronicity between `Researcher` and `Nodes` `Optimizers`

At the moment, in case of discrepancy, an error will be raised at some point:
- at the start of the training round by `Node` is auxiliary variables were received but the node either does not hold an `Optimizer`, or holds an `Optimizer` that is not aligned with the received information
- at the end of the training round by `Experiment`, if the node was able to run its round, but sent auxiliary variables that could not be properly processed, again either because no `Optimizer` is defined on the researcher side, or the one in use is not aligned with the received information

Typically, in the current implementation of Scaffold, if something is misspecified an error will be raised at the start of the first training round.

pandrey-fr:
> Hi Paul, thanks for this Pull Request!
> 
> I have seen in the code that auxiliary variables are serialized in MsgPack whatever its nature is (heavy or light parameters size - wise). In the previous work when integrating Scaffold, we tried to abide with the following principle:
> 
>     * if parameter is a scalar or  vector, send parameter through MQTT
> 
>     * otherwise if it is rather heavy (like correction term in scaffold), use a file transfer
> 
> 
> I agree that for a first version we may want to take the easiest solution as you did, which converts everything into MsgPack when it comes to auxiliary variables

Auxiliary variables is always a nested dict. It may contain lightweight information (as per your definition), however I think that trying to parse through it to decide how to share it rather than simply using a file upload is bound to add complexity (and bugs) to the code, while the current version makes it so that we leave the inspection of the contents to `Optimizer` and its plug-ins. We also make sure that the MQTT messages can only be marginally different depending on the auxiliary variables (one field is either None or a list of 1 or 2 strings containing urls), which would not be the case otherwise (because you could end up packing many scalar information that sum up to something non-lightweight).

Note that in the case of Scaffold (which, at the moment, is the only plug-in that actively uses this system - others are on the line, but have not been implemented yet), the information is "heavyweight": it is made of `Vector` instances storing correction states.

ybouilla:
> > Hi Paul, thanks for this Pull Request!
> > I have seen in the code that auxiliary variables are serialized in MsgPack whatever its nature is (heavy or light parameters size - wise). In the previous work when integrating Scaffold, we tried to abide with the following principle:
> > ```
> > * if parameter is a scalar or  vector, send parameter through MQTT
> > 
> > * otherwise if it is rather heavy (like correction term in scaffold), use a file transfer
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I agree that for a first version we may want to take the easiest solution as you did, which converts everything into MsgPack when it comes to auxiliary variables
> 
> Auxiliary variables is always a nested dict. It may contain lightweight information (as per your definition), however I think that trying to parse through it to decide how to share it rather than simply using a file upload is bound to add complexity (and bugs) to the code, while the current version makes it so that we leave the inspection of the contents to `Optimizer` and its plug-ins. We also make sure that the MQTT messages can only be marginally different depending on the auxiliary variables (one field is either None or a list of 1 or 2 strings containing urls), which would not be the case otherwise (because you could end up packing many scalar information that sum up to something non-lightweight).
> 
> Note that in the case of Scaffold (which, at the moment, is the only plug-in that actively uses this system - others are on the line, but have not been implemented yet), the information is "heavyweight": it is made of `Vector` instances storing correction states.

Yes I agree: but I  think we should go further and separate somehow "light weight" and "heavy" auxiliary variables (this piece of information can be hardcoded in the Optimizer), because it is not very optimized for light weight auxiliary variables. I am ok for leaving as you did for this Pull Request, and think about how to implement such thing in a future PR

fedbiomed/researcher/job.py
ybouilla: Yes I got your idea, I was just saying that I don't see why you have these 2 dictionaries:
because  :

let's assume that i want to share the same file across all my nodes. I upload this file on the url `https:/my/url`

then we can have a variable `aux_by_nodes` = `{'node_id-1': 'https:/my/url', 'node_id-2': 'https:/my/url', 'node_id-3': 'https:/my/url'} `  if i have 3 nodes 'node_id-1', 'node_id-2', 'node_id-3' ,  and this `aux_by_nodes` is equivalent to your `aux_shared`

but again maybe there is something I am missing here

fedbiomed/researcher/experiment.py
```diff
@@ -1643,6 +1647,36 @@ def run_once(self, increase: bool = False, test_after: bool = False) -> int:
...
+        aux_var = {}  # type: Dict[str, Dict[str, Dict[str, Any]]]
+        for reply in self._job.training_replies[self._round_current]:
+            node_id = reply["node_id"]
+            aux_var = reply.get("optim_aux_var", {})
+            for module, params in aux_var.items():
+                aux_var.setdefault(module, {})[node_id] = params
+        # If an Optimizer is used, pass it the auxiliary variables (if any).
+        if self._agg_optimizer is not None:
+            self._agg_optimizer.set_aux(aux_var)
+        # If no Optimizer is used but auxiliary variables were received, raise.
```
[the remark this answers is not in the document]
pandrey-fr: declearn already handles that for Vector-encapsulated data, based on GPU availability and the global device policy - which we should handle in Fed-BioMed (using `declearn.utils.set_device_policy` in the parts that already take care of selecting whether to use CPU or GPU).

fedbiomed/researcher/job.py
pandrey-fr: I think that you are missing the point that there can be both shared _and_ node-specific information _at the same time_. Hence the use of two distinct variables, because in the end a node can receive between zero and two urls of files to download.

So we can have `aux_shared = {module_a: {key: val}, module_b: {key: val}}` and `aux_nodes = {node_1: {module_c: {key: val}}, node_2: {module_c: {key: val}}}` that will result in `aux_url_shared = "url_to_shared_auxvar"` and `aux_url_bynode = {"node_1": "url_to_node1_auxvar", "node_2": "url_to_node2_auxvar"}`, and eventually node 1 will receive a list with both URLs to the shared and their own aux vars; similary for node 2; etc.

ybouilla: Oh I see! But then, why could you  send up to only 2 auxiliary variables for each OptiModule, and not more? I thought you could sent as many url links (in a list for instance) in `aux_nodes ` (this is what we are currently doing in Fed-BioMed).

I do think we should do the same when dealing with declearn optimodules, in order to make things as generic as possible (if in declearn you plan to implement more OptiModule that uses more than 2 auxiliary variables or if we want to implement ours)

fedbiomed/researcher/experiment.py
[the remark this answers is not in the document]
ybouilla: OK then I guess we may want to open a separate issue on that, I feel it can be interesting to have but doesnot belong to this Pull Request

ybouilla:
> at the end of the training round by Experiment, if the node was able to run its round, but sent auxiliary variables that could not be properly processed, again either because no Optimizer is defined on the researcher side, or the one in use is not aligned with the received information

Actually, I think  we should not let the User  waiting for a complete round to see an error raised

fedbiomed/researcher/job.py
pandrey-fr: > That being said, I still don't see why you could not append the shared content into the file that has the node specific content

You definitely could; but it would be less efficient in terms of communication costs, because you would send some information (the shared one) `n_nodes` times rather than once. I believe that the additional code complexity is worth it, but if you do not, I can simplify the logic at the cost of that researcher-upload overhead.

fedbiomed/researcher/experiment.py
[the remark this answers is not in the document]
pandrey-fr: I agree - and I think that this issue (dealing with device placement when using declearn) should be our next priority after the current MR. Note that I also think that it will be straightforward to address thanks to `declearn.utils.set_device_policy`.

fedbiomed/researcher/experiment.py
```diff
@@ -1643,6 +1647,36 @@ def run_once(self, increase: bool = False, test_after: bool = False) -> int:
...
+    ) -> None:
+        """Process any Optimizer auxiliary variables received during a round.
+
+        Raises:
+            FedbiomedExperimentError: if auxiliary variables were received,
+                but `agg_optimizer` is None and thus cannot process them.
+            FedbiomedOptimizerError: if the received auxiliary variables do
+                not match the expectations of the `agg_optimizer` Optimizer.
+        """
+        # Restructure the received auxiliary variables (if any).
```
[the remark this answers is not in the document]
pandrey-fr: I agree on the principle. I think part of it is inevitable with the current partition of the workload between `Job` and `Experiment`: `Job` is in charge of collecting and sending information, then gathering replies, while `Experiment` is in charge of conducting their post-processing (aggregation, optimization and application of updates); but I also think that I can improve my implementation to better stick to that partition.

I think it would make sense to:
- move the re-organization of received auxiliary variables to `Job`, so that `Experiment._process_optim_aux_var` can be shrinked to merely passing the reply-contained aux vars to the Optimizer (and handling incoherence issues).
- change `Job.upload_agg_optimizer_aux_var` signature so that instead of `agg_optimizer: Optional[Optimizer]` it expects `optim_aux_var: Optional[Dict[str, Dict[str, Any]]]`, and have `Experiment` perform the explicit call to `Optimizer.get_aux`

This way the structuration of information (for upload and at download) would all be part of `Job`, while the collection and processing at the level of the `Optimizer` would both be performed directly by `Experiment`.

Would that suit you?

pandrey-fr:
> > at the end of the training round by Experiment, if the node was able to run its round, but sent auxiliary variables that could not be properly processed, again either because no Optimizer is defined on the researcher side, or the one in use is not aligned with the received information
> 
> Actually, I think we should not let the User waiting for a complete round to see an error raised

I agree on principle, but in practice doing some verification without harmful side effects is going to be a pain. We have been wanting to implement this kind of thing in declearn for some time (both for our Model and Optimizer), but are yet to come up with an efficient design. The simplest thing I can think of is doing some kind of isolated simulation:

1. Instantiate copies of the node-side and researcher-side Optimizers (if any);
2. Collect aux vars from the researcher-side Optimizer (if any) and check that they can be processed by the node-side one:
3. Collect aux vars from the node-side Optimizer (if any) and check that they can be processed by the researcher-side one.

The will not catch all issues, and may emit some undue warnings due to the absence of gradients-processing between the second and third steps, but I think that (in the current state of things) it would be harmful to pass arbitrary-shape fake gradients, and difficult to gather the required information to pass proper-shape fake gradients.

fedbiomed/researcher/experiment.py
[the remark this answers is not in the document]
pandrey-fr: Update: I implemented the changes suggested above based on your remark. Let me know if that suits you :)

pandrey-fr:
I have started extending unit tests, so that the code added by this PR on the Researcher side is covered. I am still to do the same for the Node side code.

fedbiomed/researcher/experiment.py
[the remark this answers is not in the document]
ybouilla: > This way the structuration of information (for upload and at download) would all be part of Job, while the collection and processing at the level of the Optimizer would both be performed directly by Experiment.

> Would that suit you?

Yes this is what I think we should do!

fedbiomed/researcher/experiment.py
```diff
@@ -1574,12 +1574,18 @@ def run_once(self, increase: bool = False, test_after: bool = False) -> int:
...
+            round_=self._round_current,
+            aggregator_args_thr_msg=aggr_args_thr_msg,
+            aggregator_args_thr_files=aggr_args_thr_file,
+            do_training=True,
+            secagg_arguments=secagg_arguments,
+            optim_aux_var=optim_aux_var,
+        )
 
         # refining/normalizing model weights received from nodes
         model_params, weights, total_sample_size, encryption_factors = self._node_selection_strategy.refine(
```
ybouilla, line 1591: I know this is not your doing but for the line `aggregated_params: Dict[str, Union['torch.tensor', 'nd.ndarray']] = \`
should it be `np.ndarray` instead?

fedbiomed/researcher/job.py
```diff
@@ -475,22 +494,131 @@ def start_nodes_training_round(self,
...
+                if node_aux := mod_info.get(str(node_id)):
+                    aux_bynode[node_id][mod_name] = node_aux
+                # Case of global information shared with all nodes.
+                elif mod_name not in aux_shared:
+                    aux_shared[mod_name] = mod_info
+        # Return the restructured auxiliary variables dicts.
+        return aux_shared, aux_bynode
+
+    def get_received_optimizer_aux_var_from_round(
+        self,
```
ybouilla, line 602: I really like what you did: I am just a bit concerned by the name of the function: I like the name `get_received_optimizer_aux_var_from_round`, but should we really have to  specify `_from_round` ?

ybouilla:
> > > at the end of the training round by Experiment, if the node was able to run its round, but sent auxiliary variables that could not be properly processed, again either because no Optimizer is defined on the researcher side, or the one in use is not aligned with the received information
> > 
> > 
> > Actually, I think we should not let the User waiting for a complete round to see an error raised
> 
> I agree on principle, but in practice doing some verification without harmful side effects is going to be a pain. We have been wanting to implement this kind of thing in declearn for some time (both for our Model and Optimizer), but are yet to come up with an efficient design. The simplest thing I can think of is doing some kind of isolated simulation:
> 
>     1. Instantiate copies of the node-side and researcher-side Optimizers (if any);
> 
>     2. Collect aux vars from the researcher-side Optimizer (if any) and check that they can be processed by the node-side one:
> 
>     3. Collect aux vars from the node-side Optimizer (if any) and check that they can be processed by the researcher-side one.
> 
> 
> The will not catch all issues, and may emit some undue warnings due to the absence of gradients-processing between the second and third steps, but I think that (in the current state of things) it would be harmful to pass arbitrary-shape fake gradients, and difficult to gather the required information to pass proper-shape fake gradients.

I was thinking about the same way of checking that modules  on `Nodes` and `Researcher` sides are synchronized. I think it will catch most of the issues.

If it raises warning this might be an issue though. If too complicated to implement, we can still work on that synchronicity checking later on in an other Pull Request

fedbiomed/researcher/experiment.py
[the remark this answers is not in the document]
ybouilla: *Update**: I checked and I think and the way you re-design things is way better in term of responsibility! 
Only one small detail: ``get_received_optimizer_aux_var` could also be in `Experiment` (but I guess you did that in order to avoid appending it to Experiment-that-really-need-some-refactoring :) )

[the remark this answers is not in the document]
ybouilla: This is merely a suggestion, please resolve the conversation if you think it is doesnot make sense

fedbiomed/researcher/job.py
pandrey-fr: I am glad that you are convinced!

To me it is already documented, both by the docstring and inline comments of `Job.upload_agg_optimizer_aux_var`. That function's inline comments, that explain the flow along the code, are the following four lines:
```
# Split the information between shared and node-wise dictionaries.
# Upload the shared information that all nodes will download.
# Upload the node-specific information, with node-specific urls.
# Return the urls of the uploaded files.
```

fedbiomed/researcher/experiment.py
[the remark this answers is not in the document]
pandrey-fr: > Only one small detail: `get_received_optimizer_aux_var` could also be in `Experiment`

Well, in fact it _was_ in `Experiment` (as part of `_process_optim_aux_var`) prior to the refactoring...

I think I like it better where it is, because this way the bits of code that handle the re-structuring of aux var in both directions (researcher <-> nodes) are grouped together within `Job`. But if you disagree, I can revert that part of the changes and change the unit tests, so let me know.

fedbiomed/researcher/experiment.py
pandrey-fr: It should really be `Dict[str, Union[torch.Tensor, np.ndarray]]`. There is absolutely no value in not importing numpy and torch here, since they are automatically imported by `fedbiomed.common.training_plans`. Indeed, this is pre-existing, but I am all in favor of Uncle Bob's Boy Scout Rule, so I will happily change that.

fedbiomed/researcher/job.py
pandrey-fr: To me, `process` implies some side effects, while here this _is_ a getter, with some minor-overhead restructuring but no side effects. Hence my choosing `get`. That being said, I would be okay with something in the vein of `access` or `extract`, which might emphasize that this has some (limited) cost.

fedbiomed/researcher/experiment.py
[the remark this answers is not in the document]
ybouilla: Ok then it makes even more sense to keep it in `Job`!
Ok let's keep it in `Job` then!

fedbiomed/researcher/job.py
ybouilla: Yes sorry if i wasnot clear enough, I mean we can specify why we are doing that -> to avoid possible overhead in the communication

tests/test_job.py
```diff
@@ -443,8 +443,47 @@ def test_job_11_start_training_round(self,
         self.assertListEqual(nodes, ['node-1'])
         self.assertEqual(serialize_load_patch.call_count, 1)
 
-    def test_job_12_update_parameters_with_invalid_arguments(self):
-        """ Testing update_parameters method with invalid arguments.s"""
+    def test_job_12_start_nodes_training_round_optim_aux_var(self):
+        """Test that 'optim_aux_var' is properly used in 'start_nodes_training_round'."""
+        mock_aux_var = {"module": {"key": "val"}}
+        # General setup: skip requests sending and replies processing.
```
ybouilla, line 449: I have a very minor concern about the name `mock_aux_var`: it can be a bit confusing since it is not really a `Mock` object

fedbiomed/researcher/job.py
```diff
@@ -363,33 +366,48 @@ def start_nodes_training_round(self,
                 aggregator_args_thr_msg .
             secagg_arguments: Secure aggregation ServerKey context id
             do_training: if False, skip training in this round (do only validation). Defaults to True.
+            optim_aux_var: Auxiliary variables of the researcher-side Optimizer, if any.
         """
```
ybouilla, line 370: I would add that `optim_aux_var` is for `declearn` optimizers only (not to be confused with `aggregator_args_thr_msg`  and `aggregator_args_thr_files`  that convey auxiliary variables too)

tests/test_job.py
```diff
@@ -483,8 +522,108 @@ def test_job_14_update_parameters_from_file(self):
...
+        url_shared, url_bynode = self.job.upload_agg_optimizer_aux_var(aux_var)
+        # Verify that results and mock calls match expectations.
+        self.assertEqual(url_shared, "url")
+        self.assertDictEqual(url_bynode, {})
+        patch_serializer_dump.assert_called_once_with(
+            aux_var, "dir/aux_var_shared_uuid.mpk"
+        )
+        patch_repository_upload.assert_called_once_with(
+            "dir/aux_var_shared_uuid.mpk"
+        )
```
ybouilla, line 589: Just a (very) minor remark: we could assign `"url"` to a variable, making the test less bug prone: I mean:
```
url_link = "url"
...
self.assertEqual(url_shared, url_link)
```
same for `"dir"`

fedbiomed/researcher/job.py
```diff
@@ -475,22 +494,131 @@ def start_nodes_training_round(self,
...
+            Serializer.dump(aux_shared, path)
+            url_shared = self.repo.upload_file(path)["file"]
+        else:
+            url_shared = None
+        # Upload the node-specific information, with node-specific urls.
+        url_bynode = {}  # type: Dict[uuid.UUID, str]
+        for node_id, node_aux in aux_bynode.items():
+            if not node_aux:
+                continue
+            path = os.path.join(
```
ybouilla, line 552: I dont really understand that part: what if `node_aux = {}`? should it be considered as an error instead? Or do you see any specific setting where it can happen

tests/test_experiment.py
```diff
@@ -1251,20 +1259,53 @@ def test_experiment_27_run_once_with_optimizer(self):
...
-        optim.init_round.assert_called_once()
-        optim.step.assert_called_once()
+        self.assertEqual(result, 1, "run_once did not successfully run the round")
+        self.assertListEqual(
+            [name for name, *_ in optim.method_calls],
+            ["get_aux", "set_aux", "init_round", "step"],
+            "aggregates Optimizer did not receive expected ordered calls"
+        )
+
+    def test_experiment_28_run_once_fails_from_unexpected_nodes_aux_var(self):
```
ybouilla, line 1274: I guess we are missing a test that tests run_once with an optimizer that received `aux_var` and another that check that loading optimizer from breakpoint with `aux_var` gives the correct `aux_var`

fedbiomed/researcher/job.py
ybouilla: I think `extract` is a good alternative: let s call it : `extract_received_optimizer_aux_var_from_round` !

tests/test_experiment.py
```diff
@@ -1247,24 +1255,59 @@ def test_experiment_27_run_once_with_optimizer(self):
...
+        mock_agg.aggregator_name = "mock"
+        mock_agg.create_aggregator_args.return_value = ({}, {})
+        self.test_exp.set_aggregator(mock_agg)
+        # Disable breakpointing.
+        self.test_exp.set_save_breakpoints(False)
+        # Call run once, which should fail due to the lack of Optimizer
+        # to process the auxiliary variables received from nodes.
+        with patch.object(FedbiomedExperimentError, "__init__") as patch_exc:
+            patch_exc.return_value = None  # __init__ must return None
+            self.assertRaises(SystemExit, self.test_exp.run_once)
```
ybouilla, line 1296: wow this is very nice! I think we can even change other pieces of the test where we are testing that some method triggers `SystemExit`, but never the underlying exception !

tests/test_job.py
pandrey-fr: OK, I renamed it to `fake_aux_var`.

tests/test_job.py
pandrey-fr: Indeed; I corrected that. Thanks!

tests/test_experiment.py
pandrey-fr: > I guess we are missing a test that tests run_once with an optimizer that received aux_var (could be appended to already existing test_experiment_27)

This is already tested in `test_experiment_27_run_once_with_optimizer`, using an autospec mock `Optimizer`: we test that the methods corresponding to collecting/sending, and later unpacking/processing auxiliary variables are each called once, and in the expected order.

> and another that check that loading optimizer from breakpoint with aux_var gives the correct aux_var

We already test that loading and Optimizer restores its internal states. As aux_var are emitted deterministically based on these states, there is no more thing to test.

To me, there are no missing things in the generic/general unit tests. There would however be value in setting up an integration test using Optimizer-based Scaffold - which would be a way to conduct these verification in practice.

srcansiz:
Hi @pandrey-fr and @ybouilla, 
Just a quick reminder (even tough it is a bit late): `Draft` tag does not work as it works in GitLab. To make the pull request draft and avoid executing checks, you need to convert the pull request to draft from the right panel > reviewers > convert to draft.

pandrey-fr:
I have now added some tests for `Node` (nearly nothing) and `Round` (quite a few tests, for which I turned some backend methods public). The only thing that are not properly tested in `Round` are (a) the overall sequence in `run_model_training` (i.e. check that all aux-var-related operations happen, and in the proper order) and (b) the exception catching around `collect_aux_var` as part of that same `run_model_training`.

As `run_model_training` is a very complex method (that I would favor refactoring as an upcoming effort later this sprint or the next one), I still need a bit of time to figure out a clever scheme as to what to patch/mock and what to actually run; but the existing tests should help. In the meanwhile, the rest of the new code is now properly covered!

fedbiomed/researcher/job.py
pandrey-fr: Okay, I added a note to the docstring.

fedbiomed/researcher/job.py
```diff
@@ -475,22 +497,136 @@ def start_nodes_training_round(self,
...
+            aux_bynode[node_id] = {}
+            for mod_name, mod_info in aux_var.items():
+                # Case of node-specfic information.
+                if node_aux := mod_info.get(str(node_id)):
+                    aux_bynode[node_id][mod_name] = node_aux
+                # Case of global information shared with all nodes.
+                elif mod_name not in aux_shared:
+                    aux_shared[mod_name] = mod_info
+        # Return the restructured auxiliary variables dicts.
+        return aux_shared, aux_bynode
```
ybouilla, line 607: I was wondering here what would happen if user is using same `OptiModule` twice. Let 's say he is using an optimizer such as `Optimizer(lr=.123, modules=[ScaffoldServer(), ScaffoldServer()])` (which doesnot make sense, but is allowed by declearn I guess). Would the above work ?

pandrey-fr: No, it would not. In general, you are allowed to use multiple instances of the same plug-in, but for auxiliary variables there will be a concurrence, i.e. the last on in the chain will overwrite the variables emitted by others, and be the only one to process received ones.

This may change in the future if we see a use case for it, but at the moment there is none - as you said, implementing two Scaffold controllers at once makes no sense. It is bound to raise an error at some point (that point being the first call to `process_aux_var`), and for us it is really up to the users not to write up things that are mathematically senseless.

pandrey-fr: When we implement a tool to do some early control of configuration correctness, this is typically something that will be caught. In the meanwhile, we may add some check and verbose exception in DecLearn 2.4 - I forgot to do it because in early versions of declearn 2, it was not possible to have multiple instances of the same module.

ybouilla: So should we raise an error here ? or should we raise it when creating `Optimizer`? Or should we leave it as it is?

ybouilla:
> I have now added some tests for Node (nearly nothing) and Round (quite a few tests, for which I turned some backend methods public). The only thing that are not properly tested in Round are (a) the overall sequence in run_model_training (i.e. check that all aux-var-related operations happen, and in the proper order) and (b) the exception catching around collect_aux_var as part of that same run_model_training.

As run_model_training is a very complex method (that I would favor refactoring as an upcoming effort later this sprint or the next one), I am having trouble figuring out a clever scheme as to what to patch/mock and what to actually run. @ybouilla would you mind giving it a look and give me some advice on how to do it/whether you think it is relevant to do it in the current state of the code?

In the meanwhile, the rest of the new code is now properly covered!

I will try to implement some specific test and see about this `run_model_training ` . I guess it is ok to not test the exception is raised if too complicated to do so

fedbiomed/researcher/job.py
pandrey-fr: I think we should leave it as it is, and tackle it as part of a future task about setting up a controller that would run coherence checks / ad-hoc tests on the Researcher side prior to starting training.

tests/test_round.py
```diff
@@ -871,5 +873,207 @@ class FakeModule:
...
+        msg = rnd.process_optim_aux_var()
+        self.assertTrue(fake_error in msg)
+        mock_optim.set_aux.assert_called_once_with(fake_aux_var)
+
+    def test_round_22_collect_optim_aux_var(self):
+        """Test that 'collect_optim_aux_var' works properly with an Optimizer."""
+        # Set up a Round with an attached mock Optimizer.
+        rnd = Round()
+        mock_optim = create_autospec(Optimizer, instance=True)
+        mock_b_opt = create_autospec(BaseOptimizer, instance=True)
```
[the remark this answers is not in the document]
pandrey-fr: I do not think so. I mean, in practice, in the current state of things, it would be a `DecLearnOptimizer`, but here we are using mock objects, and by the signatures of the tested objects, it is enough to pass a `BaseOptimizer`-like mock, the `.optimizer` of which returns an `Optimizer`.

fedbiomed/researcher/job.py
ybouilla: OK ! I am going to write a little comment about that, but I am ok to leave it as it is

tests/test_round.py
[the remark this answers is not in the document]
ybouilla: I see, but such auxiliary variable exchange is specific to `DecLearnOptimizer` , so for me it would make more sense to use it instead of doing something more generic. But since the tests are passing, i am ok to leave it as it is

ybouilla:
In this PR, the team decided not to include Secure Aggregation on auxiliary variables for now. This feature will be introduced latter on in another PR.

Thus, one will be sending non-encrypted auxiliary variables when using Secure Aggregation with Scaffold in the current implementation:  therefore a new task will be added: create a warning when sending auxiliary variables, with secure aggregation turned on

**Regarding declearn's SCAFFOLD compatibility with the original SCAFFOLD implementation**:

Since the re-definition / clarification of the `Aggregator` part, original SCAFFOLD implementation will be deprecated in future FedBioMed verison, in favor of the declearn's one

ybouilla:
Regarding native scaffold:  we cannot use native scaffold with declearn one

fedbiomed/researcher/job.py
ybouilla: added

ybouilla:
Regarding this PR: I am ok to merge it, there is still an issue with a test, where a path name validation fails - while working on my local machine. I dont know how to fix that (please check logs).

tests/test_experiment.py
ybouilla: > We already test that loading and Optimizer restores its internal states. As aux_var are emitted deterministically based on these states, there is no more thing to test.

But we don't make sure that Optimizer saved through a breakpoint file is loaded correctly, since it involves a few files (mainly Optimizer content file, and breakpoint state): my feeling would be to test that in a "integration test"

ybouilla: added

ybouilla:
This Pull Request can be merged, but there are still issues to be considered:

1. make SecAgg compatible with scaffold
2. force user to use `num_updates` in the training args

PR #794, opened 2023-06-07
Feature/79 Set random seed

### LINKED ISSUE #79: Add random seed for model parameters initialization + data shuffling when applying gradient descent
In GitLab by @ybouilla on Aug 19, 2021, 15:15

random seeds are  numbers used to initialize pseudo random generator. Setting a random seed  grants model reproducibility, and is widely used by data scientist / statisticians to achieve same results given a model. It can be applied wherever a random generator is needed, thus:
1) for model parameters initialization 
2) for data shuffling before computing gradient descent.

sharkovsky:
Closes #79 .

This PR introduces the option to set the random seed on the node at the beginning of the round.

## Design choice: how to deal with impossibility to save random state between rounds on the node

There was a design choice to be made, related to the fact that Fed-BioMed does not store any state on the node between rounds. In fact, the constraint is even stricter: a Fed-BioMed node may even be restarted between two rounds!

The options I considered were:

###  Reset the random seed at the beginning of every round to the same value

Pros: it is clean. Even though it can lead to counterintuitive behaviour, once you are aware of this quirk it is very easy to understand what's happening and to compare different runs.
Cons: when using `exp.run()`, the same seed will be set at every round. This invalidates the users' expectations w.r.t. to randomization effects like shuffling the dataset, since it will always be shuffled in the same way

### Reset the random seed at the beginning of every round to a value provided automatically by the experiment class

Pros: different rounds have different random behaviour
Cons: it becomes more difficult to achieve reproducibility, because the user must remember that they cannot compare two runs with the same random seed directly, and instead they need to know how the experiment changed the random seed in each run. 

### Both of the above, with a switch in the experiment class

Pros: maximum flexibility
Cons: requires changing the experiment (i.e. the user) interface

### Implement a separate message to request setting the random seed on the node

Pros: avoids the confusion of resetting the random seed at every round
Cons: the researcher does not know what happens between rounds. It may happen that the machine is restarted, and then the random seed would be lost.

I preferred the option of `resetting the random seed at every round to the same value` because:
1) once documented, at least the user always knows clearly what is going on
2) there is a simple workaround, which is to use `run_once` and change the seed at every round
3) I did not want to change the experiment interface, adding yet-another option

### DIFF
--- a/fedbiomed/common/training_args.py
+++ b/fedbiomed/common/training_args.py
@@ -300,7 +300,7 @@ class TrainingArgs:
         | fedprox_mu | set the value of mu and enable FedProx correction |
         | dp_args | arguments for Differential Privacy |
         | share_persistent_buffers | toggle whether nodes share the full state_dict (when True) or only trainable parameters (False) in a TorchTrainingPlan |
-        | random_seed | set random seed at the beginning of each round |
+        | random_seed | set random seed on the node at the beginning of each round |
 
         """
         return {
--- a/fedbiomed/common/training_args.py
+++ b/fedbiomed/common/training_args.py
@@ -4,9 +4,8 @@
 """
 Provide a way to easily to manage training arguments.
 """
-
 from copy import deepcopy
-from typing import Any, Dict, TypeVar, Union, Tuple, Callable
+from typing import Any, Dict, Type, TypeVar, Union, Tuple, Callable
 
 from fedbiomed.common.constants import ErrorNumbers
 from fedbiomed.common.exceptions import FedbiomedUserInputError
@@ -256,6 +255,25 @@ class TrainingArgs:
 
         return True
 
+    @staticmethod
+    def optional_type(typespec: Union[Type, Tuple[Type, ...]], argname: str):
+        """Utility factory function to generate functions that check for an optional type(s).
+
+        Args:
+            typespec: type specification which will be passed to the `isinstance` function
+            argname: the name of the training argument for outputting meaningful error messages
+
+        Returns:
+            type_check: a callable that takes a single argument and checks whether it is either None
+                or the required type(s)
+        """
+        @validator_decorator
+        def type_check(v):
+            if v is not None and not isinstance(v, typespec):
+                return False, f"Invalid type: {argname} must be {typespec} or None"
+            return True
+        return type_check
+
     @classmethod
     def default_scheme(cls) -> Dict:
         """
@@ -282,6 +300,7 @@ class TrainingArgs:
         | fedprox_mu | set the value of mu and enable FedProx correction |
         | dp_args | arguments for Differential Privacy |
         | share_persistent_buffers | toggle whether nodes share the full state_dict (when True) or only trainable parameters (False) in a TorchTrainingPlan |
+        | random_seed | set random seed at the beginning of each round |
 
         """
         return {
@@ -334,6 +353,9 @@ class TrainingArgs:
             },
             "share_persistent_buffers": {
                 "rules": [bool], "required": False, "default": True
+            },
+            "random_seed": {
+                "rules": [cls.optional_type(typespec=int, argname='random_seed')], "required": True, "default": None
             }
         }
 
--- a/fedbiomed/common/training_args.py
+++ b/fedbiomed/common/training_args.py
@@ -300,7 +300,7 @@ class TrainingArgs:
         | fedprox_mu | set the value of mu and enable FedProx correction |
         | dp_args | arguments for Differential Privacy |
         | share_persistent_buffers | toggle whether nodes share the full state_dict (when True) or only trainable parameters (False) in a TorchTrainingPlan |
-        | random_seed | set random seed on the node at the beginning of each round |
+        | random_seed | set random seed at the beginning of each round |
 
         """
         return {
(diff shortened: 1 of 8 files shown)


srcansiz:
Hi @sharkovsky, 
`Draft:` prefix is not working on GitHub pull request. After you create your pull request you can convert it to draft form the right panel > reviewers  > "Still in progress? convert to draft"

fedbiomed/common/training_args.py
```diff
@@ -334,6 +353,9 @@ def default_scheme(cls) -> Dict:
             },
             "share_persistent_buffers": {
                 "rules": [bool], "required": False, "default": True
+            },
+            "random_seed": {
+                "rules": [cls.optional_type(typespec=int, argname='random_seed')], "required": True, "default": None
```
srcansiz, line 358: Hi @sharkovsky, 

I didn't understand why you did need create `optional_type` method. Isn't `required=False` for optional arguments?

sharkovsky: Sadly no 😢 

There are two confusing things
## 1. `required` does not mean the user is required to provide the value

The way it is implemented, `required=True|False` means that it is required during the execution of the round, not that the user is required to provide it. 
There is an interaction with the `default` field as well, and the fact that inside the round we initialize the TrainingArgs using `TrainingArgs(self.training_kwargs, only_required=False)`. 

I will give some examples to illustrate:

| `required` | `default` exists? | user provided the value? | behaviour as implemented | 
| ---  | --- | --- | --- |
| True | yes | no | default value is substituted |
| True | no | no | error |
| False | yes | no | default value is substituted |
| False | no | no | nothing is done |

So basically if I put `required=True`, then even if the user doesn't specify a value the default one will be provided (so `required` does not mean it is required from the user)

## 2. `rules` are independent of `required`
Even if `required` had the expected meaning, the `rules` would still be checked: if I put `"rules": int, ..., "default": None` it will fail because `None` is not an `int`.

fedbiomed/common/training_plans/_base_training_plan.py
```diff
@@ -86,6 +85,11 @@ def post_init(
             aggregator_args: Arguments managed by and shared with the
                 researcher-side aggregator.
         """
+        # Set random seed
+        rseed = training_args['random_seed']
+        if rseed is not None:
+            random.seed(rseed)
```
srcansiz, line 91: Hi @sharkovsky,

As I understand random.seed for mainly for; 
- model initialization 
- data shuffling 

However, model initialization is never done on the node side. Node always gets initial parameters from the researcher. So since it is possible for researcher set random seed in his python runtime i consider that it is more useful for training data. 

However, random seed is set in `post_init` so it will impact every method that uses randomness.  My concern is that how it will  impact other parts of training for example DP. 

-

sharkovsky: My purpose with `random_seed` was to maximize reproducibility. That is why I tried to set the random seed as soon as possible and with a global method. I am happy if it also affects DP. 

Maybe I can make this clearer in the documentation, since I only spoke about model and data there. Really, the purpose is to enable:
- reproducible results
- debugging

srcansiz (changes requested):

sharkovsky:
Converting to draft because:
- we want to better investigate if the node should have a mechanism for disabling random seeds from the researcher
- According to [some](https://pytorch.org/docs/stable/data.html#randomness-in-multi-process-data-loading) [sources](https://gist.github.com/ihoromi4/b681a9088f348942b01711f251e5f964?permalink_comment_id=4585244#gistcomment-4585244), setting the seed for dataloaders may need to be done manually and separately.

srcansiz:
> Converting to draft because:
> 
>     * we want to better investigate if the node should have a mechanism for disabling random seeds from the researcher
> 
>     * According to [some](https://pytorch.org/docs/stable/data.html#randomness-in-multi-process-data-loading) [sources](https://gist.github.com/ihoromi4/b681a9088f348942b01711f251e5f964?permalink_comment_id=4585244#gistcomment-4585244), setting the seed for dataloaders may need to be done manually and separately.

Hi @sharkovsky,

I've checked `opacus` library and search for the places where [randomness](https://github.com/search?q=repo%3Apytorch%2Fopacus+manual_seed&type=code) is used. I think setting seeds does not affect opacus implementation. However, there are some utility function where random seed is set as hard coded but I didn't see those functions are used  in the library. Also, `opacus` let users to set seed manually. I consider it as there is no problem of using random seed  from DP point of you.  

For me it is okay to merge it. Let me know do you want to add extra configuration for `dataloaders` and `cuda`.

sharkovsky:
Hi @srcansiz, 

I have confirmed that our implementation correctly also influences dataloaders (and I assume cuda too, since we use `torch.manual_seed`). So in theory it is also ready for merge on my side.

However, I would ask a final review because I made a slight change in the implementation:
- before: we only set random seed if the researcher used the `random_seed` training argument
- now: we always set random seed, but we rely on built-in functionalities to set it to a random value if the researcher did not specify `random_seed` in the training arguments

The reason for this change is that, since we now always explicitly reset the random seed at the beginning of the round, there is less risk for interaction between different experiments. 
In the previous implementation, the random seed set by one experiment would affect the ones after it. 

There was a corner case where we run one round of experiment A with random seed N, one round of experiment B without random seed, another round of experiment A with the same random seed N, and another round of experiment B without random seed. The expectation would be that in experiment A, in both rounds we read the exact same samples in the same order, while in experiment B we read them in random order. However, in practice what would happen is that we would read them in the same order also in experiment B. But with the new implementation, this is avoided because a new random seed would be set (with a random value!).

srcansiz (approved):

Issue #532, opened 2023-05-04
Create additional CI slaves for MacOS


srcansiz:
In GitLab by @mvesin on May 4, 2023, 10:16

- [x] test MacOS CI slave (on CloudStack ?)
- [x] create MacOS CI slave (on catalysis-zie ?)
  - done: un-filter from Jenkins to catalysis-zie
- [x] add automatic CI tests for MacOS for `fedbiomed/fedbiomed`

PR #799, opened 2023-06-09
add a job strategy to test on macosx-M1

ErwanDemairy:
Part of #532 implementation.


ErwanDemairy:
already implemented when setting #803

Issue #448, opened 2023-01-09
Revise versions of packages included in Fed-BioMed


srcansiz:
In GitLab by @sharkovsky on Jan 9, 2023, 10:12

Revise versions of dependencies in Fed-BioMed:
- update conda environments `.yaml` files
- update node GUI yarn packages dependencies `packages.json` 
- upgrade prerequisite `docker-compose` (v1) to `docker compose` (v2) https://docs.docker.com/compose/migrate/ 
- update docker container files (base image, dependencies)
- replace deprecated `nosetests` by `pytest`

General direction is to try to stick to up-to-date dependency versions (*).
Exceptions:
- python 3.10 (because 3.11 is quite new and not yet supported by some dependencies)
- torch 1.x (because 2.0 is also new and not yet supported by some dependencies)
- boringtun 0.4.0 (because we could not have working setup with 0.5.x yet) 

Tasks:

- manually test on systems supported (fedora, ubuntu, mac, windows)
  - ~~[ ] Ubuntu 22.04 LTS~~ will be tested with CI
  - [x] Fedora 38
  - [x] MacOS X 12.6.6
    * dev: 101, sklearn perceptron, medical folder dataset
    * vpn: 101, medical folder dataset (w/ batch size 4 and no local/global test, because 8GB RAM limit on test laptop)
  - [x] Windows 11 (21H2) w/ WSL2 Ubuntu 22.04 LTS
    * dev: 101, medical folder dataset, secagg
    * vpn: 101, gui, (not tested medical folder dataset)
- udpdate/test CI slaves
  - [x] Ubuntu 22.04 LTS
  - ~~[ ] MacOS xxx~~ not yet active
- [x] python version (3.9 to 3.10 or 3.11) => 3.10
  - [x] consequence: replace nosetests
- [x] conda envs packages update
  - include: fedbiomed-{node,network,researcher,gui} fedbiomed-ci ; exclude: fedbiomed-{doc,doc-test} (will change with github migration)
  - [x] check/adapt: conda env can be created
  - [x] check/adapt: pass unit tests
  - [x] check/adapt: run notebooks manually (tested on Fedora 38 for: 101, declearn, flamby, breakpoint, gpu, secagg, monai classification, mednist, opacus mnist, celeba, sklearn perceptron + medical folder dataset w/ 2 nodes and training plan approval)
  - [x] check/adapt: pass CI builds
- [x] gui packages
  - upgrade `gui/ui/package.json` content
[...]

PR #802, opened 2023-06-19
Revise versions of packages #448

mvesin:
**PR description**

Implements #448 "Revise versions of packages included in Fed-BioMed"

Please see the issue for details of the branch's content.

Tests were already conducted under:
- Fedora 38 (used for dev of this branch)
- Ubuntu 22.04 LTS (CI)
- MacOS X 12.6.6 (could not have a MacOS X 13.x testing machine)
- Windows 11 (21H2) w/ WSL2 Ubuntu 22.04 LTS

The review can be and/or of:
- visual review of the changes ?
- additional tests ?

When merging this PR to `develop`, please advise developers/users in Discord as it may require on their laptops:
- upgrading some system packages (eg: `docker compose v2` replaces v1)
- cleaning out and rebuilding all their Fed-BioMed conda environments (though upgrade of the envs may function, it is probably safer/cleaner to do so) 
  ```
  for i in node researcher gui network ; do conda remove -y -n fedbiomed-$i --all ; done
  ./scripts/configure_conda
  ```
- cleaning out all their Fed-BioMed docker images and containers, plus GUI yarn packages. The simplest way is to clean also the configs with
  ```
  source ./scripts/fedbiomed_environment clean
  ./scripts/fedbiomed_vpn clean image
  ```

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:


.github/actions/unit-tests/action.yml
```diff
@@ -18,13 +18,14 @@ runs:
...
         source ~/.bashrc
+        conda clean -i -y
         ./scripts/configure_conda researcher
         ./scripts/fedbiomed_run network
       shell: bash
 
     - name: Run unit tests 
       run: |
-          PYTHONPATH=${PYTHONPATH:-$PWD} conda run -n fedbiomed-researcher nosetests -w ${{ inputs.test-dir }} --cover-xml --cover-erase --with-coverage --cover-package=fedbiomed --with-xunit -v --process-restartworker
+          PYTHONPATH=${PYTHONPATH:-$PWD} conda run -n fedbiomed-researcher pytest -v --cov=fedbiomed --cov-report term --cov-report xml:${{ inputs.test-dir }}/coverage.xml ${{ inputs.test-dir }}
```
srcansiz, line 28: Hi @mvesin, 

Thank you for the great effort! 

I got an error while running unittest tests with pytest. I think it is due to my environment but others may have the same error too. We may the change the way that we test `test_environ_01_initialize_common_variables_01`. 

```
tests/test_environ_common.py::TestEnviron::test_environ_01_initialize_common_variables_01 - AssertionError: '/user/scansiz/home/projects/fedbiomed-dev/fedbiomed' != '/home/scansiz/projects/fedbiomed-dev/fedbiomed'
```

mvesin: Thanks @srcansiz , I'm going to robustify this test.

Nevertheless, I suspect it might be due to the command you launch for test, not the migration to pytest.
Documented command in `tests/README.md` is:

```
cd tests
pytest -v --cov=fedbiomed --cov-report term --cov-report xml:coverage.xml
```

... which works fine for me, but if I run from `/home/mvesin/GIT/fedbiomed/fedbiomed/tests` the following then I get (and find it quite normal) the same error that you reported:

```
# use an alternate path to reach the dir
$ pytest -v --cov=fedbiomed --cov-report term --cov-report xml:coverage.xml /user/mvesin/home/mvesin/GIT/fedbiomed/fedbiomed/tests
...
FAILED ../../../../../../user/mvesin/home/GIT/fedbiomed/fedbiomed/tests/test_environ_common.py::TestEnviron::test_environ_01_initialize_common_variables_01 - AssertionError: '/home/mvesin/GIT/fedbiomed/fedbiomed' != '/user/mvesin/hom...

```

srcansiz: Thank you @mvesin, 

I also tested to run `pytest` in the `tests`  directory. It still fails on the same test.

scripts/fedbiomed_run
```diff
@@ -242,11 +242,11 @@ case $1 in
                 CONTAINER_UID=$(id -u) CONTAINER_GID=$(id -g) \
                              CONTAINER_USER=$(id -un | sed 's/[^[:alnum:]]/_/g') \
                              CONTAINER_GROUP=$(id -gn | sed 's/[^[:alnum:]]/_/g') \
-                             docker-compose build restful mqtt
+                             docker compose build restful mqtt
```
srcansiz, line 245: Hi @mvesin,

After merging this branch into develop/master the installations that have old version of docker won't work anymore. Do you think that it is good idea to add deprecation warning instead of failing. For example hte script can be modified to support following scenarios:

- user has old docker version  -> raise deprecation warning and ask for installing version and run `docker-compose`
- user has newer version -> use `docker compose`

I know that it is also the same for conda packages.  It would be nice to have versioning for conda yml files that we can check which version of conda env is in use. Because the users who will pull from latest and have already a older conda envs will face problems.  We do not have mechanism that checks the version of the conda env. Of course it may be overkill for this merge request. However, if we assume that the Fed-BioMed will distributed in package managers such as conda or pip it won't be problem since each time Fed-BioMed upgraded or installed the packages compatible with the version will be reinstalled or upgraded as well. 

However, to me, docker is more independent than conda environments considering that the user may also use docker for other applications. Conda is more for having isolated environment which is fine to ask user upgrade your conda environment.

.github/actions/unit-tests/action.yml
mvesin: @srcansiz : pushed a patch for this, can you confirm that it also fixes the issue for you ?

srcansiz: Hi @mvesin,

Yes it worked! thank you

envs/development/conda/fedbiomed-node.yaml
```diff
@@ -10,56 +10,55 @@ channels:
...
 dependencies:
   # minimal environment
-  - python >=3.9,<3.10
-  - pip
-  - jupyter
-  - ipython
+  # python 3.11 recently released 2022-11 and not yet supported by some deps including: torchvision
+  - python >=3.10,<3.11
+  - pip >= 23.0
+  - jupyter ~=1.0.0
```
srcansiz, line 16: Hi @mvesin,

I know that it was already there but do we need jupyter installed on the node side?

envs/development/conda/fedbiomed-researcher.yaml
```diff
@@ -10,63 +10,59 @@ channels:
 
 dependencies:
   # minimal environment
-  - python >=3.9,<3.10
-  - pip
-  - jupyter
-  - ipython
+  - python >=3.10,<3.11
+  - pip >= 23.0
+  - jupyter ~=1.0.0
```
srcansiz, line 15: Hi @mvesin,

It seems that Jupyter classic notebooks won't be maintained anymore. https://jupyter-notebook.readthedocs.io/en/latest/migrate_to_notebook7.html they recommend to migrate Notebook 7  and JupyterLab instead of notebooks. We may create another issue for that but I am adding this comment since this pull request is about upgrading packages.

envs/vpn/conda/fedbiomed-node.yaml
```diff
@@ -10,56 +10,55 @@ channels:
...
 dependencies:
   # minimal environment
-  - python >=3.9,<3.10
-  - pip
-  - jupyter
-  - ipython
+  # python 3.11 recently released 2022-11 and not yet supported by some deps including: torchvision
+  - python >=3.10,<3.11
+  - pip >= 23.0
+  - jupyter ~=1.0.0
```
srcansiz, line 16: Same comment for VPN conda envs, I think we can remove Jupyter installation so it may reduce the time consumed for installation.

scripts/fedbiomed_vpn
```diff
@@ -221,7 +232,7 @@ containers_clean() {
 
     for i in base basenode
     do
-        docker-compose rm -sf $i >/dev/null
+        docker compose rm -sf $i >/dev/null
```
srcansiz, line 235: We can also add a warning here regarding old `docker-compose`.

envs/vpn/conda/fedbiomed-node.yaml
```diff
@@ -10,56 +10,55 @@ channels:
...
   # git notebook striper
-  - nbstripout
-  - joblib >=1.0.1
-  # sklearn
-  #   scipy >= 1.9 from conda-forge needs recent GLIBC thus causes issue 389
-  #   with many current systems
-  #   another option is to install scipy from pip
-  - scipy >=1.8.0,<1.9.0
-  - scikit-learn >=1.0.0,<1.1.0
+  - nbstripout >=0.6.1,<0.7.0
```
srcansiz, line 34: Hi @mvesin, 

As I understand `nbsripout` used for converting `.ipynb` files to `.py`  python modules. It is only used for interation test so it may not need to be installed on the node side. Or it can be part of `fedbiomed-ci` environment only (in the future)

envs/vpn/conda/fedbiomed-node.yaml
```diff
@@ -10,56 +10,55 @@ channels:
...
-  - requests >=2.25.1,<3.0.0
-  - paho-mqtt >=1.5.1,<2.0.0
-  - validators >=0.18.2,<0.19.0
-  - tqdm >=4.59.0,<5.0.0
-  - git
-  - packaging >=23.0,<24.0
+  - requests ~=2.29.0
+  - paho-mqtt ~=1.6.1
+  - validators >=0.20.0,<0.21.0
+  - tqdm ~=4.65.0
```
srcansiz, line 27: I didn't see this module is used somewhere in Fed-BioMed. May it left from an old implementation?

envs/vpn/conda/fedbiomed-node.yaml
```diff
@@ -10,56 +10,55 @@ channels:
...
-  - GitPython >=3.1.14,<4.0.0
-  - requests >=2.25.1,<3.0.0
-  - paho-mqtt >=1.5.1,<2.0.0
-  - validators >=0.18.2,<0.19.0
-  - tqdm >=4.59.0,<5.0.0
-  - git
-  - packaging >=23.0,<24.0
+  - requests ~=2.29.0
+  - paho-mqtt ~=1.6.1
+  - validators >=0.20.0,<0.21.0
```
srcansiz, line 26: I saw that this module is used only for validating URLs. I think a regex implementation will be enough to validate a URL or built-in `urllib`. Using an  third-party module seemed unnecessary to me.  However, it is not in the scope of this PR. I am leaving this comment to leave a track. I may create a new issue for that.

envs/vpn/conda/fedbiomed-node.yaml
```diff
@@ -10,56 +10,55 @@ channels:
...
   # minimal environment
-  - python >=3.9,<3.10
-  - pip
-  - jupyter
-  - ipython
+  # python 3.11 recently released 2022-11 and not yet supported by some deps including: torchvision
+  - python >=3.10,<3.11
+  - pip >= 23.0
+  - jupyter ~=1.0.0
+  - ipython ~=8.13.2
```
srcansiz, line 17: Some comment for `ipython` i don't whether it is really necessary since it is never used on the node side.

scripts/fedbiomed_run
mvesin: Thank you @srcansiz for the comment & review:
- regarding versioning of conda envs: I did not implement yet (might not be that simple, raises questions about what we exactly want to do/check), but created issue #808 for it
- regarding `docker compose` migration smoothing for users, I added in (hopefully) all scripts that use the command
  * backward compatibility: if no `docker compose` is found, it raises a warning and uses `docker-compose`. I'm not sure it is a good idea as we mentionned that we don't support it but that may ease the short term life of users. Mid-term, we should probably remove this backward compatibility additions and keep only `docker compose` (and fail if not existing)
  * failure with error message if no `docker compose` or `docker-compose` is found. Exception: the `fedbiomed_environment clean` will just skip docker container cleaning, and continue with as much cleaning as possible to fit the spirit of this script

envs/development/conda/fedbiomed-node.yaml
mvesin: Thanks for noticing ... it is probably not needed. I remove it in all 3 node environments (dev/linux, dev/macos, vpn/linux)

envs/vpn/conda/fedbiomed-node.yaml
mvesin: OK, also removing `ipython` for the 3 node environments

envs/development/conda/fedbiomed-researcher.yaml
mvesin: Good point.
I noted this as a comment in the `yml` files, but will not treat it as a part of this PR (at least needs to redo testings).
My understanding is we also need to choose between classical notebooks v7 and migrating to jupyterlab.

Issue #809 is created for it.

envs/vpn/conda/fedbiomed-node.yaml
mvesin: `nbstripout` is for removing notebooks' output (not for converting to `.py`).

This is coupled to some `.gitattributes` action for auto removing notebooks' output when comnitting notebooks. But it required some developers' configuration of its clone to be effective (and I can't find any documentation for it ...). Moreover, I don't know if someone is still using it. Asking all the developers. If not, we may completely remove it.

In the meantime:
- removed from node envs (we don't use notebooks in node envs)
- added comment "consider removing nbstripout" in yml

envs/vpn/conda/fedbiomed-node.yaml
mvesin: Thanks for noticing.
I would favor using `urllib` over custom regexp.
I suspect the initial idea was that we would have to check many types of data, but as it is not the case, it is probably better to stick to a simpler alternative, as you suggest.

scripts/fedbiomed_vpn
mvesin: Done for all instances of `docker compose` in (hopefully) all scripts using it.

envs/vpn/conda/fedbiomed-node.yaml
mvesin: `tqdm` is used in `download_and_split_ixi.py` script, which is used in medical folder dataset notebook.

As discussed, we may remove from the `yml` file and replace with instructions for `pip install tqdm` in the notebooks + documentation. Doing this.

srcansiz:
Hi @mvesin,

Thanks again for the PR and the great effort of updating all the packages. It is okay for me to merge!

envs/vpn/conda/fedbiomed-node.yaml
mvesin: We introduced a regression with that one, see #803 (will be corrected through this PR)

Issue #849, opened 2023-09-06
False positive in `check_data_quality` function prevents me from running experiment


sharkovsky:
The `check_data_quality` function prevents me from running an experiment because of a mismatch in data types, even though as a researcher I am aware of this and I can easily fix it in the `training_data` function.

1) Due to the internal logic of `pandas`, the data types for a csv dataset **at the time of adding the dataset to the node** can be interpreted differently on different nodes (example: a column with only integers has int type, but the same column on another node with integers and NaN values becomes float)
2) Afterward, as a researcher when I build the experiment, I get an error because the data types are not consistent among nodes
3) This is not a real issue, because as a researcher I can easily fix the inconsistencies inside the `training_data` function. But I cannot change the way the dataset was loaded on the node (and therefore I am stuck with the default dtype logic of pandas)

This prevents me from running the experiment, even though it shouldn't.

mvesin:
This issue is related to #424

sharkovsky:
Fixed by #866

PR #844, opened 2023-08-21
fixes doc build error

srcansiz:
**PR description**

This PR fixes build error in documentation build.  Build was raising error due to non-updated template files. Therefore, documentation API uses default template files instead of customized ones. Customization is done in the style.css.

- Docs display private attributes ( except the one start with `!^__` double underscore). I though tahat it may be useful. However, I am leaving it to reviewer's decision. It is possible to hide private attrs and methods as it is  in the previous versions

**Developer Certificate Of Origin (DCO)**

By opening this merge request, you agree to the
[Developer Certificate of Origin (DCO)](https://github.com/fedbiomed/fedbiomed/blob/master/CONTRIBUTING.md#fed-biomed-developer-certificate-of-origin-dco)

This DCO essentially means that:

- you offer the changes under the same license agreement as the project, and
- you have the right to do that,
- you did not steal somebody else’s work.

**License**

Project code files should begin with these comment lines to help trace their origin:
```
# This file is originally part of Fed-BioMed
# SPDX-License-Identifier: Apache-2.0
```

Code files can be reused from another project with a compatible non-contaminating license.
They shall retain the original license and copyright mentions.
The `CREDIT.md` file and `credit/` directory shall be completed and updated accordingly.

**Guidelines for PR review**

General:

* give a glance to [DoD](https://fedbiomed.org/latest/developer/definition-of-done/)
* check [coding rules and coding style](https://fedbiomed.org/latest/developer/usage_and_tools/#coding-style)
* check docstrings (eg run `tests/docstrings/check_docstrings`)

Specific to some cases:

* update all conda envs consistently (`development` and `vpn`, Linux and MacOS)
* if modified researcher (eg new attributes in classes) check if breakpoint needs update (`breakpoint`/`load_breakpoint` in `Experiment()`, `save_state`/`load_state` in aggregators, strategies, secagg, etc.)
* if modified a component with versioning (config files, breakpoint, messaging protocol) then update the version following the rules in `common/utils/_versions.py`

PR #850, opened 2023-09-06
[HOTFIX] Introduce option to skip data quality check in `job`

sharkovsky:
This hotfix introduces the option to skip the data quality checks during initialization of `job`, because in some cases it can lead to false positives as explained in #849 . 

Introducing such option seemed like the simplest way to avoid the problem with minimal code changes (Personally, I would get rid of the data quality checks altogether). I am open to discuss other approaches.

**NOTE** This hotfix also includes the changes needed for building the documentation, as introduced in #844


mvesin:
Thank you @sharkovsky , is this PR intended to be merged
1. as a "normal" feature branch (in develop)
2. as a hotfix (in master and develop) because it is a major bug that cannot wait for next release.

My understanding is we are case 1. because this is a non critical "annoyment" so it does not need immediate fix in release.

As a reminder in case 1. (git-flow + our dev doc) branch should be named `feature/xxx` and in case 2. `hotfix/xxx` (this avoids my current question :-) ). Thank you !

sharkovsky:
Hi @mvesin , 
it is intended to be a hotfix because it is needed for the EUCAIM demonstrator, hence it needs to be linked to a (stable) numbered version (I would prefer not asking users to rely on develop). I can rename the branch to hotfix!

mvesin:
@sharkovsky thanks !
I suspect you also have to re-open a PR in that case (seems like it was auto-closed at branch rename). Sorry.

PR #865, opened 2023-09-15
Correctly import all training plan dependencies on researcher side while initializing Job

### LINKED ISSUE #864: `NameError` when declaring Experiment
A  `NameError` can occur when calling the Experiment constructor because of the following chain of events:
1) call Experiment init
2) this calls Job init
3) this instantiates a TrainingPlan, and calls `post_init`

When point 3 is reached, the `NameError` arises because the dependencies of the training plan have not been correctly imported by the Job init. Indeed, in other cases (for example, on the node side in Round) we import the training plan file before instantiating the training plan class, which solves the problem. 

A small example that reproduces the error is given below

```python
from fedbiomed.common.training_plans import TorchTrainingPlan
from fedbiomed.common.logger import logger
from fedbiomed.common.data import DataManager, MedicalFolderDataset
from fedbiomed.researcher.experiment import Experiment
from fedbiomed.researcher.aggregators.fedavg import FedAverage

class UNetTrainingPlan(TorchTrainingPlan):

    def init_model(self, model_args):
        model = DenseNet(spatial_dims=1,in_channels=1,out_channels=1)
        return model


    def init_optimizer(self):
        optimizer = AdamW(self.model().parameters())
        return optimizer

    def init_dependencies(self):
        # Here we define the custom dependencies that will be needed by our custom Dataloader
        deps = ['from monai.networks.nets import DenseNet',
                "import torch.nn as nn",
                'import torch.nn.functional as F',
                'from torch.optim import AdamW',]
        return deps

    def training_data(self):
    # The training_data creates the Dataloader to be used for training in the general class Torchnn of fedbiomed
        dataset = MedicalFolderDataset(
            root=self.dataset_path,
            data_modalities='T1',
            target_modalities='label',
            transform=None,
            target_transform=None,
            demographics_transform=None)
        loader_arguments = {'shuffle': True}
        return DataManager(dataset, **loader_arguments)


    def training_step(self, data, target):
        return 0

tags =  ['ixi-train']

exp = Experiment(tags=tags,
                 model_args={},
                 training_plan_class=UNetTrainingPlan,
                 training_args={},
                 round_limit=1,
                 aggregator=FedAverage()
                )
```

Which leads to the error
```bash
--------------------
Fed-BioMed researcher stopped due to unknown error:
name 'DenseNet' is not defined
More details in the backtrace extract below
--------------------
Traceback (most recent call last):
  File "/Users/fcremone/dev/fedbiomed/fedbiomed/researcher/experiment.py", line 72, in payload
    ret = function(*args, **kwargs)
  File "/Users/fcremone/dev/fedbiomed/fedbiomed/researcher/experiment.py", line 278, in __init__
    self.set_job()
  File "/Users/fcremone/dev/fedbiomed/fedbiomed/researcher/experiment.py", line 1430, in set_job
    self._job = Job(reqs=self._reqs,
  File "/Users/fcremone/dev/fedbiomed/fedbiomed/researcher/job.py", line 145, in __init__
    self._training_plan.post_init(model_args={} if self._model_args is None else self._model_args,
  File "/Users/fcremone/dev/fedbiomed/fedbiomed/common/training_plans/_torchnn.py", line 157, in post_init
    self._configure_model_and_optimizer()
NameError: name 'DenseNet' is not defined
--------------------
```

sharkovsky:
Fix #864.

This PR introduces the following change:
- **Before**: inside Job init, we instantiated a training plan class but it was "unusable" (i.e. `post_init` could fail) because we never imported the necessary modules
- **Now**: we import the modules necessary to enable the training plan class instance to actually execute `post_init`

Details:
1. The modules are imported in the same way that we do on the node, i.e. by saving a training plan file with the necessary import statements and then importing that file
2. inside job init, we need to instantiate the training plan class twice: the first time is just to be able to save the code. In this case, the instance is still in an "unusable" state. The second time is actually the good one with the correctly imported modules

### DIFF
--- a/fedbiomed/researcher/job.py
+++ b/fedbiomed/researcher/job.py
@@ -142,7 +142,7 @@ class Job:
 
         else:
             self._training_plan = self._training_plan_class
-        self._training_plan._configure_dependencies()
+        self._training_plan.configure_dependencies()
 
         # find the name of the class in any case
         # (it is `model` only in the case where `model` is not an instance)
--- a/fedbiomed/researcher/job.py
+++ b/fedbiomed/researcher/job.py
@@ -142,8 +142,7 @@ class Job:
 
         else:
             self._training_plan = self._training_plan_class
-        self._training_plan.post_init(model_args={} if self._model_args is None else self._model_args,
-                                      training_args=self._training_args)
+        self._training_plan._configure_dependencies()
 
         # find the name of the class in any case
         # (it is `model` only in the case where `model` is not an instance)
@@ -163,6 +162,9 @@ class Job:
 
         self._repository_args['training_plan_url'] = repo_response['file']
 
+        self._training_plan = self._load_training_plan_from_file()
+        self._training_plan.post_init(model_args={} if self._model_args is None else self._model_args,
+                                      training_args=self._training_args)
         # Save model parameters to a local file and upload it to the remote repository.
         # The filename and remote url are assigned to attributes through this call.
         try:
@@ -177,6 +179,13 @@ class Job:
         # Validate fields in each argument
         self.validate_minimal_arguments(self._repository_args,
                                         ['training_plan_url', 'training_plan_class', 'params_url'])
+
+    def _load_training_plan_from_file(self):
+        sys.path.insert(0, self._keep_files_dir)
+        module = importlib.import_module(os.path.split(self._training_plan_file)[1][:-3])
+        train_class = getattr(module, self._training_plan_name)
+        sys.path.pop(0)
+        return train_class()
 
     @staticmethod
     def validate_minimal_arguments(obj: dict, fields: Union[tuple, list]):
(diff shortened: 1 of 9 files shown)


mvesin:
`bugfix/864-name-error-in-job` is not a git-flow branch name and does not indicate clearly how to review this PR.
Please remember to name it as `feature/864-name-error-in-job` (as this is to be treated as a "normal" PR).

If it were a hotfix: `hotfix/864-name-error-in-job`

Thanks !

fedbiomed/researcher/job.py
```diff
@@ -178,6 +180,13 @@ def __init__(self,
         self.validate_minimal_arguments(self._repository_args,
                                         ['training_plan_url', 'training_plan_class', 'params_url'])
 
+    def _load_training_plan_from_file(self):
+        sys.path.insert(0, self._keep_files_dir)
+        module = importlib.import_module(os.path.split(self._training_plan_file)[1][:-3])
```
mvesin, line 185: Rather than compute `os.path.split(self._training_plan_file)[1][:-3]` (makes assumption on file name structure, thus difficult to maintain) I would propose to pass a `training_plan_module` variable.

=> I push a patch for this

PR #732, opened 2023-05-30
Draft: Feature/479 bug in integration test

srcansiz:
In GitLab by @ErwanDemairy on Mar 16, 2023, 12:22

**MR description**

TO_BE_FILLED_BY_MR_CREATOR

**Developer Certificate Of Origin (DCO)**

By opening this merge request, you agree the
[Developer Certificate of Origin (DCO)](https://gitlab.inria.fr/fedbiomed/fedbiomed/-/blob/develop/CONTRIBUTING.md#fed-biomed-developer-certificate-of-origin-dco)

This DCO essentially means that:

- you offer the changes under the same license agreement as the project, and
- you have the right to do that,
- you did not steal somebody else’s work.

**License**

Project code files should begin with these comment lines to help trace their origin:
```
# This file is originally part of Fed-BioMed
# SPDX-License-Identifier: Apache-2.0
```

Code files can be reused from another project with a compatible non-contaminating license.
They shall retain the original license and copyright mentions.
The `CREDIT.md` file and `credit/` directory shall be completed and updated accordingly.

**Guidelines for MR review**

General:

* give a glance to [DoD](https://fedbiomed.gitlabpages.inria.fr/latest/developer/Fed-BioMed_DoD.pdf)
* check [coding rules and coding style](https://fedbiomed.gitlabpages.inria.fr/latest/developer/usage_and_tools/#coding-style)
* check docstrings (eg run `tests/docstrings/check_docstrings`)

Specific to some cases:

* update all conda envs consistently (`development` and `vpn`, Linux and MacOS)
* if modified researcher (eg new attributes in classes) check if breakpoint needs update (`breakpoint`/`load_breakpoint` in `Experiment()`, `save_state`/`load_state` in aggregators, strategies, secagg, etc.)

PR #872, opened 2023-09-27
Feature/479 bug in integration test

ErwanDemairy:
For previous discussions, see: PR #732
This new PR is done in order to be able to invite @srcansiz as a reviewer, since the PR #732 was created by him, and one cannot be reviewer of its own PR.


srcansiz:
Hi @ErwanDemairy, 

Thanks for implementing End-2-End tests! 

I updated the `GRPC_SERVER_SETUP_TIMEOUT` as `env` variable that is only increased while running end to end tests. I also disabled running end to end test while pushing the current branch (feature/479...). It is only triggered once there is a push to master or develop branches, and every night at 01:30. 

After the checks are completed I am going to merge it in develop.   

Thanks again for the effort you spend while integrating end to tests.

Issue #886, opened 2023-10-10
update coding rules and DoD


mvesin:
Update coding rules and DoD regarding:
- arguments checking
- exceptions handling
- unit tests
- docstrings

The goal is to work more efficiently, balancing simplicity and quality of the code.

PR #888, opened 2023-10-16
update coding rules

mvesin:
**PR description**

Update coding rules after core dev team meeting #886

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:


CODING_RULES.md
```diff
@@ -11,90 +11,156 @@ to be completed
...
-  - except of the top level program, it is **forbidden** to trap all exceptions (with ```except:``` or ```except Exception```)
-
-
-- the **try:** block is as small as possible
-
-- force to read the documentation
-
+### exceptions handling
+
+- on the node: in general, node should not stop because of exceptions that occur during its operation
```
srcansiz, line 16: Hi @mvesin, 
I think it is better to specify that node should not stop while executing requests received from researcher rather than "its operation". For other operations, usually CLI operations, node can stop with an exception.

mvesin: Thank you, pushed that modification

CODING_RULES.md
```diff
@@ -11,90 +11,156 @@ to be completed
...
+
+    Do:
+    ```
+    try:
+      mycode()
+    except FedbiomedSomeError as e:
+      ...
+    except FedbiomedError as e:
+      ...
+    except Exception as e:
```
srcansiz, line 92: Hi @mvesin, 

We can also specify that if all other exceptions different than `FedbiomedError` are considered as undetermined error and needs to be cached at the top layer where the first request is executed. For example, `Node.task_manager.train_task`

mvesin: Thank you @srcansiz 
Can you please reformulate, I don't understand your comment ? (sentence seems unfinished)

srcansiz: What I meant is that `except Exception` cases are undetermined error cases that are usually not expected. It can be due to anything. Therefore,  we should consider those as `Undefined` error while sending back to researcher without attaching the exception message.  These exceptions should always be caught where the task is executed (top-layer: where `round.run_model_training` or `secagg.setup`  methods are executed) rather than  lower layers. This is of course not the case if developer wants to add some specificity to the undefined error message from the lower layers.

CODING_RULES.md
```diff
@@ -11,90 +11,156 @@ to be completed
...
+    Don't
+    ```
+    try:
+      mycode()
+    except:
+      ...
+    ```
+
+  - should separate FedbiomedError and other exceptions
+
```
ybouilla, line 83: ACtually it depends what the dev wants to do: he could just want to log a generic error or process wrt the error raised. I would complete the example:

```
    try:
      mycode()
    except FedbiomedSomeError as e:
       logger.warning("mycode failed with SomeError")
    except FedbiomedError as e:
       logger.critical("mycode failed with a FedBioMedError")
    except Exception as e:
       logger.critical(f"mycode failed with unknown error {e}")

```

CODING_RULES.md
```diff
@@ -11,90 +11,156 @@ to be completed
...
+    except Exception as e:
+      ...
+    ```
+
+- in general, a class shouldn't log in logger when raising or re-raising an exception. The class should log when catching and not re-raising a FedbiomedError exception. The class can log when catching a and not re-raising a python exception.
+
+  Do:
+  ```
+  try:
+    some_function()
```
ybouilla, line 109: maybe we can specify `from fedbiomed.common.logger import logger`

CODING_RULES.md
```diff
@@ -11,90 +11,156 @@ to be completed
...
+    ```
+    try:
+      some_code()
+    except Exception as e:
+      ...
+    ```
+
+- in general, a class shouldn't log in logger when raising or re-raising an exception. The class should log when catching and not re-raising a FedbiomedError exception. The class can log when catching a and not re-raising a python exception.
+
+  Do:
```
ybouilla, line 106: I would complete the above  with something like: 

" The idea is to not trigger exception so `Node` process is not killed"

CODING_RULES.md
```diff
@@ -11,90 +11,156 @@ to be completed
...
-
-except SomeError as e:
-
-    _msg = ErrorNumbers.FBxxx.value + ": the file " + filename + " does not exist"
-
-    logger.error(_msg)
-    raise OneOfFedbiomedError(_msg)
-```
+  ```
+  try:
```
ybouilla, line 142: we could specify the import

`from fediomed.common.constants import ErrorNumbers`

CODING_RULES.md
mvesin: Thank you @srcansiz 

Pushed a new version that (hopefully) addresses these points
- `except Exception:` in top layer
- `except Exception:` in lower layers
- node sending errors to researcher in the case of exception caught at top layer

CODING_RULES.md
```diff
@@ -11,90 +11,156 @@ to be completed
...
+      mycode()
+    except FedbiomedSomeError as e:
+      ...
+    except FedbiomedError as e:
+      ...
+    except Exception as e:
+      ...
+    ```
+
+    Don't (in case where the Exception can be a FedbiomedError)
```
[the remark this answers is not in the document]
mvesin: (this paragraph was removed in the while)

CODING_RULES.md
mvesin: Thank you @ybouilla 
I added imports for logger, error messages, and exceptions

CODING_RULES.md
mvesin: Added imports for error numbers

CODING_RULES.md
mvesin: Thank you @ybouilla 

I think this exists in this paragraph (at the beginning of the exception handling rules, because this is not specific to log/logger):

> on the node: in general, node should not stop because of exceptions that occur while executing requests received > from researcher. Top level layer code should catch and handle the exceptions, and can send an error message to 
> the researcher (but without full exception message to avoid leaking information).

CODING_RULES.md
mvesin: Thank you @ybouilla 

OK to add a mention like "so as to take distinct actions".
But I don't like the idea of writing `logger.xxx()` in the example because it can confuse the dev into thinking that he should always log message in such scenario (while it can also raise error, do none of the above). In other words, this rule is not about logging IMO.

ybouilla (commented):
Besides, i am surprised we did not mention about licensing at the beginning of each files of FedbBioMed

mvesin:
@ybouilla 

> Besides, i am surprised we did not mention about licensing at the beginning of each files of FedbBioMed

This is mentioned in the more general "usage & tools" document (which points to the coding rules for technical implementation details rules): 
https://fedbiomed.org/latest/developer/usage_and_tools/#license

CODING_RULES.md
```diff
@@ -11,90 +11,163 @@ to be completed
...
+- when a class raises an exception, it raises a FedbiomedSomethingError: use exceptions defined in **fedbiomed.common.exceptions** :
+
+  Do:
+  ```
+  from fedbiomed.common.exceptions import FedbiomedSometypeError
+  raise FedbiomedSometypeError()
+  ```
+
+  Don't:
+  ```
```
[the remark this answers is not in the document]
mvesin: This is detailed at the begining of the paragraph ("when a class raises an exception, it raises a FedbiomedSomethingError, not a python native error")

CODING_RULES.md
```diff
@@ -11,90 +11,163 @@ to be completed
...
-- force to read the documentation
-
+### exceptions handling
+
+- on the node: in general, node should not stop because of exceptions that occur while executing requests received from researcher. Top level layer code should catch and handle the exceptions, and can send an error message to the researcher (but without full exception message to avoid leaking information).
+
+- on the researcher: general behaviour is to propagate the exceptions to the top level layer, where they are transformed to a friendlier output. Researcher displays this output and stops.
+
+- when a class raises an exception, it raises a FedbiomedSomethingError: use exceptions defined in **fedbiomed.common.exceptions** :
+
```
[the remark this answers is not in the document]
mvesin: Thank you.

Added a mention "use exceptions defined in **fedbiomed.common.exceptions** or define new ones (eg: one per module)"

We often define one per module, but this is not a rule that this should be always be the granularity.

PR #895, opened 2023-10-25
Implementing fix for  loading Node state while testing

ybouilla:
**PR description**

A bug has been spotted regarding Node state saving facility:
it was no more possible to use validation feature on Node state.

The bug was due to the fact that each time a TrainRequest is issued, node state is saved.
However, testing on local updates is a specific TrainRequest, and doing such test updates  the NodeStateManager with a new `state_id`.

This fix intents to fix this bug by checking if :
- a training is done
- NodeStateAgent is not updated when performing testing on local updates

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/fedbiomed/node/round.py
+++ b/fedbiomed/node/round.py
@@ -115,21 +115,43 @@ class Round:
         self._optim_aux_var = {}  # type: Dict[str, Dict[str, Any]]
         self._node_state_manager = NodeStateManager(environ['DB_PATH'])
 
-    def initialize_validate_training_arguments(self) -> None:
+    def _initialize_validate_training_arguments(self) -> None:
         """Validates and separates training argument for experiment round"""
 
         self.training_arguments = TrainingArgs(self.training_kwargs, only_required=False)
         self.testing_arguments = self.training_arguments.testing_arguments()
         self.loader_arguments = self.training_arguments.loader_arguments()
 
-    def initialize_node_state_manager(self, previous_state_id: Optional[str] = None):
+    def initialize_node_state_manager(self, previous_state_id: Optional[str] = None,
+                                      test: Optional[bool] = False):
         """Initializes [`NodeStateManager`][fedbiomed.node.node_state_manager.NodeStateManager]. 
 
         Args:
             previous_state_id (optional): previous state_id from which to load `Node` state.
-            Defaults to None (which is the value for `Round` 0).
+                Defaults to None (which is the value for `Round` 0).
         """
-        self._node_state_manager.initialize(previous_state_id=previous_state_id)
+
+        self._node_state_manager.initialize(previous_state_id=previous_state_id,
+                                            testing=not self.training)
+    
+    def initialize_validate_training_arguments(self):
+        # Initialize and validate requested experiment/training arguments.
+        try:
+            self._initialize_validate_training_arguments()
+
+        except FedbiomedUserInputError as e:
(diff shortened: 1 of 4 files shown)


fedbiomed/node/round.py
```diff
@@ -442,7 +463,12 @@ def run_model_training(
                 except Exception as exc:
                     error_message = f"Cannot train model in round: {repr(exc)}"
                     return self._send_round_reply(success=False, message=error_message)
+                try:
```
mvesin, line 466: What is the reason for moving these lines here ?

Though it does not break anything, I don't really see a reason.

Moreover, IMO, it makes more sense to keep it where it is currently, aka just before *using* the node state (`results['state_id'] = self._node_state_manager.state_id`). This avoids saving states that won't be used, in case of error between the two.

But that's also OK for me to commit your change, if your think it makes sense to keep it.

fedbiomed/node/round.py
```diff
@@ -114,21 +114,55 @@ def __init__(
...
+        try:
+            self._initialize_validate_training_arguments()
+        except FedbiomedUserInputError as e:
+            return self._send_round_reply(success=False, message=repr(e))
+        except Exception as e:
+            msg = 'Unexpected error while validating training argument'
+            logger.debug(f"{msg}: {repr(e)}")
+            return self._send_round_reply(success=False, message=f'{msg}. Please contact system provider')
+
+    def initialize_arguments(self,
```
mvesin, line 150: Thank you @ybouilla for the nice fix !

As discussed, we may simplify a little bit the code by removing some very short functions:
- `initialize_node_state_manager` (merge code in `initialize_arguments`)
-  `_initialize_validate_training_arguments` (merge code in `initialize_validate_training_arguments`)

You may keep `initialize_validate_training_arguments` (and maybe make it private ?).

fedbiomed/node/round.py
mvesin: Discussed together the case where:
- training is True
- test ratio is 1

In this case, this change introduces a bug, as a `state_id` (not None) is created and returned to the researcher, but no state is saved for this `state_id`.

So the options are:
1. restore initial position of block
2. keep position of block but get it out of the `if self.training_plan.training_data_loader is not None:` test
3. add tests for the case where training is True and test_ratio is 1, so it doesn't read state.

IMO:
1. is the most simple and clear
2. is also OK
3. should really really be avoided, as it creates complexity for next-step developers both on node and researcher side: instead of a simple condition "when training is True, then a state is saved by the node", they have to handle the case where test ratio is 1. This is error prone.

fedbiomed/node/round.py
ybouilla: Regarding the points you spotted above, I 'd like to keep `initialize_validate_training_arguments` and make it private.
Concerning the `_save_round_state` method place, i undo the changes. Hence, if `test_ratio = 1`, Node state will still be saved even if no training is performed. We made this decision for simplicity sake.

fedbiomed/node/round.py
ybouilla: 1. has been done. changes undone.

fedbiomed/node/round.py
ybouilla: However, TrainRequest fails if `test_ratio=1` and `test_on_local_updates=True`. This bug is not that simple to fix and a issue has be opened #901

mvesin (approved):

PR #913, opened 2023-11-14
Feature/879 redesign request

srcansiz:

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:


fedbiomed/researcher/requests/_requests.py
```diff
@@ -0,0 +1,496 @@
...
+        """ Pings online nodes
+
+        Returns:
+            List of ID of up and running nodes
+        """
+        ping = ResearcherMessages.format_outgoing_message({
+            'researcher_id': environ["ID"],
+            'command': "ping"}
+        )
+        with self.send(ping) as federated_req:
```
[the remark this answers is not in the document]
srcansiz: Thanks Marc, yes I forgot to add it. I pushed a new commit to add timeout for ping.

mvesin:
Hi @srcansiz , as discussed I gave a look to the following 2 cases. Found no specific security/robustness issue, so it is OK for me to keep as it is. Nevertheless, this would be nice at some point to have (general) tighter type/value check for data received in the network `Messages`. Currently, asserting this creates no issue needs navigating through many classes ...

1. Node doesn't check that received `parties` are a `list[str]` (just checks it is a `list`)
2. `request_id` is optional in `RequestReply` messages, so Node may have to handle (eg) `TrainRequest` with `request_id: None`

fedbiomed/transport/node_agent.py
```diff
@@ -112,11 +155,14 @@ async def send(self, message: Message) -> None:
...
                             "as DISCONNECTED soon if no request received.")
 
-        try:
-            await self._queue.put(message)
-        except Exception as exp:
-            raise FedbiomedCommunicationError(
-                f"{ErrorNumbers.FB628}: Can't send message to the client. Exception: {exp}")
+        # Updates replies
+        async with self._replies_lock:
+            self._replies.update({
```
mvesin, line 160: Hi @srcansiz 
I wonder if we should add:
```
if message.request_id:
   self._replies.update(....)
```

This avoids running the `on_reply` if we send a `message` with a `request_id` of `None` or empty string.
Though this case would be a programmer error, this would be pretty annoying. For example, with `request_id = None` it would divert a LOG or SCALAR message, etc.
This case is not covered/secured by the type check in `GrpcServer.send()` which checks that message is ` Message` (that's fine).
Currently this case cannot happen as in current version, `GrpcServer.send()` is called only from `Request.send()` which adds a `request_id`.

Another option would be to add a check in `GrpcServer.send()` that `request_id` is not `None` / not empty string.

This would make sense for the sake of modularization between gRPC layer / upper layers.

What do you think ?

fedbiomed/researcher/requests/_requests.py
```diff
@@ -0,0 +1,495 @@
...
+
+        with self.send(message, nodes, policies=[DiscardOnTimeout(5)]) as federated_req:
+            errors = federated_req.errors()
+            replies = federated_req.replies()
+
+            # TODO: Loop over errors and replies
+            for node_id, error in errors.items():
+                logger.info(f"Node ({node_id}) has returned error {error.errnum}, {error.extra_msg}")
+
+        return replies
```
mvesin, line 481: Returned value is not what is expected  / described by prototype.
Pushing a fix for this.

mvesin: Moreover, `NodeAgent` is lacking a way to simply request the `node_id` (which is needed for this fix). Thus, one has no easy way to query a `FederatedRequests` nodes IDs (needed for the fix ...)
=> adding this

fedbiomed/researcher/requests/_requests.py
```diff
@@ -0,0 +1,495 @@
...
+        """Constructor of the class.
+
+        Args:
+            message: either a common `Message` to send to nodes or a dict of distinct message per node
+                indexed by the node ID
+            nodes: list of nodes that are sent the message
+            policy: list of policies for controlling the handling of the request
+        """
+
+        self.message = message
```
mvesin, line 136: Hi @srcansiz 
All variables from `FederatedRequests` should probably be private withe getter when needed (eg `self.policies` `self.requests`.

This would be nice implementation IMO as it prevents direct modification of the variables (which is not supposed to occur)

fedbiomed/researcher/requests/_requests.py
```diff
@@ -0,0 +1,495 @@
...
+    ) -> None:
+        """Single request for node
+
+        Args:
+            message: Message to send to the node
+            node: Node agent
+            request_id: unique ID of request
+            sem_pending: semaphore for signaling new pending reply
+        """
+        self.request_id = request_id if request_id else str(uuid.uuid4())
```
mvesin: hi @srcansiz 

Same remark as `FederatedRequests` : it would probably be nice to have private variables with getters when needed (and setter for `self.status` to allow policies to set it)

fedbiomed/transport/node_agent.py
```diff
@@ -74,31 +84,64 @@ def id(self) -> str:
...
-        async with self._status_lock:
+        Args:
+            request_id: request ID for which the replies should be flushed
+            stopped: the request was stopped during processing
+        """
 
-            # Inform user that node is online again
-            if self._status == NodeActiveStatus.DISCONNECTED:
-                logger.info(f"Node {self._id} is back online!")
+        async with self._replies_lock:
```
mvesin, line 95: Hi @srcansiz 
I don't really understand what the `self._stopped_request_ids` is for ?
(nevertheless, this does not introduce bug - so it is at worst useless - so it's minor issue).

1. it is only used to do a distinct `logger.warning()` in `on_reply` - useful ?
2. this `logger.warning()` does not seem to me very significant because the `self._stopped_request_ids` contains both (1) ID of the request if this node answered before a node stopped the request (2) ID of the request if this node didn't answered before a node stopped the request

IMO the `logger.warning()` would be different in case (1) and (2), which would require something like:
```
        async with self._replies_lock:
           if stopped and self._replies[request_id]['reply'] is None:
               async with self._stopped_request_ids_lock:
                    self._stopped_request_ids.append(request_id)

            self._replies.pop(request_id, None)
```

So finally for me, though it would gain to:
a. either remove completely `self._stopped_request_ids` and associated lock
b. or improve it (eg see above)

fedbiomed/transport/node_agent.py
srcansiz: Hi Marc, 

`LOG` and `SCALAR` messages has their own channel and service so request replies never confused with `Feedback` channel. If `request_id` is `None`. `NodeAgent` will create a key in `self._replies` as `None` like `{None: {on_reply: callback}}`. This can be dangerous because developer won't see any error. Therefore, I agree that we may add your modification.  

I was actually thinking about removing `GrpcServer.send` because it is not used anymore. All messages are send through `NodeAgent.send`. It is because each message from researcher to node is considered as request. I would suggest to remove it for now and add it when there is use-case for it.

fedbiomed/transport/node_agent.py
srcansiz: Hi Marc, 

I think naming was confusing for `self._stopped_request_ids`. It only saves the request ids that has not answered before stopping request. Please look `if stopped and self._replies[request_id]['reply'] is None:`. For example, let's assume that there is strategy for StopOnDisconnect

- Node 1 -> Starts processing request
- Node 2 -> Starts processing request
- Node 3 -> Starts processing request
Then, 

- Node 3 -> Disconnects 
- Federated request stops  

There Node 2 and Node 1 still process the request, however the reply callback is flushed due to stopped request. Then, Node 2 and Node 1 sends their replies. In that case, researcher will assume that a reply received with unexpected request ID.  However, it was a reply for federated request that is stopped.   `self._stopped_request_ids` keeps this in memory and print proper message for those situations.  

If you think that is not necessary I am okay with removing it  but I don't see any improvement.  The improvement should be on the node side where the jobs for canceled federated requests should be canceled on the node side as well.

fedbiomed/transport/node_agent.py
mvesin: > LOG and SCALAR messages has their own channel and service so request replies never confused with Feedback channel
You're right, thanks for clarifying.

> Therefore, I agree that we may add your modification.
OK.
Maybe I would favor implementation which checks in GrpcServer.send() that request_id is not None / not empty string (to have all checks in same place) - but it's up to you.

> I was actually thinking about removing GrpcServer.send
OK for me

fedbiomed/researcher/requests/_requests.py
srcansiz: `request_id` can be private but creating setter for `self.status` is not needed. It will just increase the complexity. These attribute are only used developers and I don't see any reason to implement setter to validate the inputs.

mvesin: Yes, you're right it's not needed for `self.status` as it's read/write.

Yet `status` `error` and `reply` are probably concerned, and maybe also `message` and `node`

fedbiomed/transport/node_agent.py
mvesin: Discussion: only test in `on_reply`, not `GrpcServer.send()` (because it's ok to send messages without request_id

fedbiomed/transport/node_agent.py
mvesin: Pushed the proposed modification to handle the case of duplicate replies for stopped requests.

mvesin (approved):

PR #919, opened 2023-11-28
Feature/875 grpc tls

srcansiz:
**PR description**

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:


fedbiomed/common/constants.py
```diff
@@ -39,7 +40,16 @@
 JOB_PREFIX = 'job_'
 """Prefix for job ID"""
 
+
+ETC_FOLDER_NAME = 'etc'
```
mvesin, line 44: - `ETC_FOLDER_NAME` is duplicate of `CONFIG_FOLDER_NAME` => removing it
- `CERTS_FOLDER_NAME` could be also used for the existing hardcoded use of `etc/certs` => modifying this

fedbiomed/transport/client.py
```diff
@@ -84,25 +152,23 @@ def __init__(self, node_id: str, researcher: ResearcherCredentials, update_id_ma
...
         self._id = None
+        self._researcher = researcher
+        self._channels = Channels(researcher)
 
-        self._port = researcher.port
-        self._host = researcher.host
-
-        feedback_channel = create_channel(port=researcher.port, host=researcher.host, certificate=None)
-        feedback_stub = ResearcherServiceStub(channel=feedback_channel)
+        self._task_stub = None
```
mvesin, line 158: `self._task_stub` is now unused. Same for 3 following vars => removing them

fedbiomed/common/cli.py
```diff
@@ -135,6 +137,47 @@ def initialize_create_configuration(self) -> None:
                  "If the configuration file exists, leave it unchanged"
         )
 
+        create.add_argument(
```
mvesin, line 140: New syntax for configuration creation is not backward compatible (`fedbiomed_run node configuration create` and `fedbiomed_run researcher configuration create` don't work anymore). So we need:

1. either to provide backward compatible syntax
2. or update all the references existing. Found some in `fedbiomed_run` + in documentation and tutorials + in notebooks + in VPN configuration scripts ...

fedbiomed/common/config.py
```diff
@@ -0,0 +1,210 @@
...
+
+        # Write configuration file
+        return self.write()
+
+    @abstractmethod
+    def add_parameters(self):
+        """"Component specific argument creation"""
+
+
+class NodeConfig(Config):
```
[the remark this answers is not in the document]
srcansiz: Hi @mvesin, 

I wanted to implemented it as independently in `common.cli` .  If I would add it `researhcer.config` `common.cli` would have to import `researcher` and `node` modules.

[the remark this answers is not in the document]
srcansiz: Hi @mvesin, 
Thanks, I will check your ideas and see how it will go.

fedbiomed/common/constants.py
```diff
@@ -53,8 +59,8 @@
 # major version, else the minor version. Micro versions are supported but their use is currently discouraged.
 
 __version__ = FBM_Component_Version('5.0.0')  # Fed-BioMed software version
-__researcher_config_version__ = FBM_Component_Version('2')  # researcher config file version
-__node_config_version__ = FBM_Component_Version('2')  # node config file version
+__researcher_config_version__ = FBM_Component_Version('3')  # researcher config file version
```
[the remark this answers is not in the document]
srcansiz: I think it is normal to break configuration version (or any version) in develop because it is a development branch. If you add version 2 and then 3 in the develop the master will jump from 1 to 3 directly.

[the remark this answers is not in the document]
mvesin: ... but on the other hand if we don't do that we'll have to send yet-another warning message when merging this branch because it will break the configurations of most of us who already are using the develop branch ... as it did for me and I spent quite a while trying to figure why I got an ugly stack trace (instead of a clean "version changed" message). 

I can revert to v2, but I'll let you communicate to other people about that when we merge the branch :-)

fedbiomed/common/cli.py
mvesin: Thank you Sergen, I see now all mentions have updated syntax

scripts/fedbiomed_run
```diff
@@ -271,11 +268,12 @@ case $1 in
...
 
-        # TODO : better test with python module loading attempt and try/except
-#        source ${basedir}/scripts/fedbiomed_environment node "$IP_ADDRESS" -v
+
+         # Activate node envirenment
+        source "$basedir"/scripts/fedbiomed_environment node "$IP_ADDRESS"
 
         # silently initialize environment before any command
-        if ! forward_command_to_python_cli node "configuration create" "$IP_ADDRESS" >/dev/null; then
+        if ! python -m fedbiomed.common.cli configuration create --component node --name $CONFIG_FILE -uc  >/dev/null; then
```
mvesin, line 276: A (minor ?) issue with new syntax is that this results in creating 2 distinct node configurations:

```
./scripts/fedbiomed_run config config-n2.ini configuration create -c node -n config-n3.ini
```

This is a bit misleading (annoyance) but not really a bug, we might choose to ignore.

But I don't see any obvious way to improve while we are in this mixed handling of parameters between fedbiomed_run and CLI parser

fedbiomed/node/environ.py
```diff
@@ -27,39 +27,34 @@
...
         """Constructs NodeEnviron object """
         super().__init__(root_dir=root_dir)
-        logger.setLevel("INFO")
-        self._values["COMPONENT_TYPE"] = ComponentType.NODE
-        # Setup environment variables
-        self.setup_environment()
 
-    def default_config_file(self) -> str:
-        """Sets config file path """
+        self.config = NodeConfig(root_dir)
```
mvesin, line 39: Shouldn't `self.config` be a private `self._config` ? 
Same question for `fedbiomed.researcher.environ`

fedbiomed/common/config.py
```diff
@@ -0,0 +1,210 @@
...
+
+    DEFAULT_CONFIG_FILE_NAME: str = 'config_node.ini'
```
mvesin, line 161: Wouldn't `DEFAULT_CONFIG_FILE_NAME` `COMPONENT_TYPE` `CONFIG_VERSION` gain to be private variables ?
They are only used within the class (and parent class).

Same in `ResearcherConfig`.

srcansiz: I consider those as constant attributes of the config classes.  I think making it private is very little detail  but if you say that it is not clear to understand or confusing I can update make it private.

fedbiomed/node/environ.py
srcansiz: Hi Marc,
`self.config` is the reference for configuration file which a static file. Environ is a dynamic object which may change based on env variable. I believe it is useful to make config non-private and accessible for other classes in case there is need to access original variable set in the config.

mvesin: Thank you Sergen,
By "access" do you mean read or read+modify :-) ?
- if read+modify this is fine to keep as it is
- if read only, then it would be nice to have private var + getter

fedbiomed/common/config.py
mvesin: Thank you Sergen,
Yes it would be nice to make them private to clarify they are constants.

srcansiz: Sorry you right it is good to make it private to clearly state for developers that is not meant to change.

fedbiomed/node/environ.py
srcansiz: It is done

fedbiomed/common/config.py
srcansiz: I pushed a new commit for this.

fedbiomed/common/config.py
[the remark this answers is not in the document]
srcansiz: Hi @mvesin, 
I tried to implement your suggestions. However, it does not behave as it is expected. Since researcher cli import researcher environ, and researcher environ instantiates `ResearcherConfig` it automatically create a confgiuration file if there is no existing.  Therefore, the cli has no control over the configuration creation. So now I remember why I wanted to implement these config classes in the common module and intialize argument parser in `common.cli`. 

I don't know what do you think but I would prefer controlling configuration creation independently from environ class, and I couldn't find a way to do it smoothly from the component specific cli module since they all import environ class which depends on configuration file.

scripts/fedbiomed_run
srcansiz: No actually, I forgot to remove `initialize_create_configuration` from `node.cli` and `researhcer.cli`. However, -which is also related to comment above- if configuration is created with component specific CLI it will create two distinct configuration.

fedbiomed/common/config.py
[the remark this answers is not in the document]
srcansiz: What I would suggest is to conditionally and dynamically import `ResearcherConfig` and `NodeConfig` in `common.cli``

Here is the `_create_component` method of `common.cli`. 
```python
 if args.component.lower() == "node":
            NodeConfig = importlib.import_modue('fedbiomed.node.config').NodeConfig
            config = NodeConfig()
        elif args.component.lower() == "researcher":
            ResearcherConfig = importlib.import_modue('fedbiomed.researcher.config').ResearcherConfig
            config = ResearcherConfig(root=args.root, name=args.name, auto_generate=False)
        else: 
            print(f"Undefined component type {args.component}")
            return
```

You may name it as dirty solution :) but i think it is quite important to not to mix environ and config classes. What I wanted to do in this implementation was to handle config parsing and verification before starting the application/node/researcher. Than it is environ class that is responsible of reading config file and overwriting some value based on user decision.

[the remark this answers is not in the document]
mvesin: Thank you Sergen, I missed that point.
Nevertheless doesn't the same issue exist with current implementation (automatic creation of configuration) ?

```
source ./scripts/fedbiomed_environment clean
source ./scripts/fedbiomed_environment researcher
python -m fedbiomed.node.cli --list
```

... and we have a node config file created (but maybe it is linked to your last comment here https://github.com/fedbiomed/fedbiomed/pull/919#discussion_r1415558000

scripts/fedbiomed_run
mvesin: Would you push that modif without the `initialize_create_configuration` then ?
This might open room for a nicer solution to the environ/config issue :-)

fedbiomed/common/config.py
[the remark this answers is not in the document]
srcansiz: Yes because of `environ` class it is automatically created as a short cut to lunch components quickly. It is not a problem but when launching node. It will be problem when running configuration create `fedbiomed_run node configuration create --name test.ini' because there CLI 
- first: imports `environ` that instantiate `NodeConfig`  that creates configuration file or reads existing one. 
- second: it will go in arg parser for `_create_component` and try to recreate configuration however it will complain because there is already one created. 
- or there will be problem of generating two distinct configuration. 

Therefore, I wanted to separate Config management and Environ management, along with the new positional argument such as `--force` to overwrite config file. This approach will be much more flexible if once bashc CLI and python CLI redesigned to be merged in single Python CLI.

If you think that it is not necessary I can revert back to the version in the develop.

scripts/fedbiomed_run
srcansiz: Yes I am going to push that version soon.

fedbiomed/transport/client.py
```diff
@@ -57,25 +60,89 @@ def create_channel(
...
+        try:
+            s.connect(address)
+        except socket.error:
+            return False
+        else:
+            s.close()
+            return True
+
+
+class Channels:
```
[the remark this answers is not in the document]
srcansiz: Hi @mvesin, 
I pushed a fix for that.

scripts/fedbiomed_run
srcansiz: I pushed the new version a moment ago.

fedbiomed/common/config.py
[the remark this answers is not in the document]
srcansiz: I pushed a new version that implements the suggestion in this comment https://github.com/fedbiomed/fedbiomed/pull/919#discussion_r1415579178

fedbiomed/transport/client.py
```diff
@@ -140,6 +220,31 @@ async def send(self, message: Message) -> None:
...
+        """Updates connection state and dispatch event to run listeners
+
+        This method also implements auto-trust for server certificate
+        """
+
+        while True:
+            if is_server_alive(self._researcher.host, self._researcher.port):
+                # Gets server certificate before creating the channel
+                # This implementation assumes that the provided IP and PORT trusted
+                self._researcher.certificate = \
```
mvesin, line 265: Unless I'm mistaking, the server certificate is never saved by the node in this version, each time the node connects to the researcher it picks the current certificate for that session.

This is perfectly OK for this first version, but if you confirm this, I'd like to add an explicit comment for future developer that we have a hardcoded policy here, which is OK for honest but curious but still enables malicious MITM attacks.

mvesin:
Hi Sergen, thank you for this nice work.
No other comments on my side, especially found nothing special on the SSL part (which is the core of this PR).
FYI checked running some notebooks, tested several scenario of create/override/force configuration + sniffed network traffic to ensure gRPC really uses TLS.

fedbiomed/transport/client.py
mvesin: I added a few lines to your existing comment

srcansiz:
> Hi Sergen, thank you for this nice work. No other comments on my side, especially found nothing special on the SSL part (which is the core of this PR). FYI checked running some notebooks, tested several scenario of create/override/force configuration + sniffed network traffic to ensure gRPC really uses TLS.

Thanks @mvesin I pushed my last changes. If you don't have any other comments it can be merged now.

PR #943, opened 2024-01-03
Feature/896 job refactor

### LINKED ISSUE #896: refactor `Experiment` to have a Base shared between Experiment and FedAnalytics + refactor `Job`
Initial [proposal slides](https://docs.google.com/presentation/d/1Mz_0ICu6bhvK9J730B4f_KuqgYk6zvCBelNItXSRmV4/edit?usp=sharing)

See also PR #943 for additional description on the design choices

Here is a checklist of the tasks that still need to be done
- [x] #1031
- [x] Review Job (full hierarchy) implementation
- [x] #1032
- [x] #1010
- [x] Fix #1003 
- [x] #1028
- [x] #1029
- [x] #1030
- [x] #944
- [x] #1022
- [x] #1011
- [x] #1048
- [x] #1049
- [x] fix cleaning of temporary dirs created at training on node side (no issue)
- [x] #1033
- [x] #1042
- [x] #1012
- [x] #1041
- [x] [MV] cross check code for functions inconsistencies
   *  clean code from unused code/functions
   * update/remove references to refactored methods that changed signature or were removed
- [x] [MV] PEP8 code style (fix style, remove unused imports, ...)
- [x] [YB] review and manual testing of notebooks, including SecAgg
  * as part of tutorials/notebooks review in #1042
  * [SC] review of security tutorialx
  * [MV] review of scikit-learn tutorials & notebooks
- [x] test building containers + run some notebooks (101, secagg)
- [ ] merge develop
- [ ] ~Review potential interactions with work on declearn secure aggregation aux variables~
- [x] #1023

## Collaboration workflow

For reviews, suggestion, or small changes, you may modify the feature branch directly. 
For bigger changes, please create a branch from [the feature branch](https://github.com/fedbiomed/fedbiomed/pull/943) and create a merge request.

@sharkovsky in charge of keeping in sync with develop

sharkovsky:
Close #896 

TODO before merging PR
- [x] Verify that all version numbers have been correctly updated
- [x] Hunt down, tag, and close related issues 

This PR introduces a significant refactoring of the `Experiment` and `Job` classes, mainly in terms of their class hierarchy (but not limited to that).

Changes:
- New hierarchy for `Experiment`
- New hierarchy for `Job`
- Single source of truth and consistency for several user-level concepts, especially for model parameters, training data, nodes, etc...
- New module `fedbiomed.researcher.federated_workflows` which includes `Experiment` and `Job`
- Using a training plan file to initialize an `Experiment` is no longer supported

In short:
- `Experiment` and `Workflow` classes are the interface for the researcher. They implement what the researcher wants "to do", typically in their `run` method
- `Job` classes are a representation of a task that is executed on the node side. Multiple job classes can be used by one workflow, according to its needs.  
- `Job` classes now have a much more functional API, and a short life-cycle (they are destroyed when the corresponding task is finished)

The redesign follows a stratified approach, with the main layers explained below:

| layer | description |
| --- | --- |
| Notebook | The top-most layer with the code written by the researcher in their notebook |
| Experiment | Manages life-cycle of any objects that are relevant to the user, exposes an API that is inspired by the business logic |
| Job | Contains implementation details for manipulating the objects from the layer above, translates the business-logic APi into the levels below |
| Requests | Exposes an application-level API for remote communication, targeted at Fed-BioMed developers |
| Grpc | Low-level communication implementation |

## New class hierarchy for `Experiment`

We now have `FederatedWorkflow > TrainingPlanWorkflow > Experiment`. 
The goal is to prepare for `FederatedAnalytics`, which will also inherit from `FederatedWorkflow` (but without a training plan, as per our latest discussion)

| Concept of workflow | 
| --- | 
| A `FederatedWorkflow` contains the API to allow the researcher to perform a broad set of tasks that are somehow related. For example, the `Experiment` contains all the methods for referencing the remote data, configuring the model, potentially handle security, launching the training, interpreting the results, etc.. As another example, in the future `FederatedAnalytics` will also be a workflow, and it will contain methods for referencing the remote data, executing analytics, and interpreting results. |

Design point: these classes have members that have a tied life-cycle. The chosen approach is to try to automatically maintain consistency between class members, transparently to the user. 

### `FederatedWorkflow`
A FederatedWorkflow is the main entry point for the researcher to orchestrate both local and remote operations.
The FederatedWorkflow is an abstract base class from which the actual classes used by the researcher must inherit.
It manages the life-cycle of:
    - the federated data (i.e. the `FederatedDataset`) class
    - the training arguments
    - secure aggregation
    - the node state agent
Additionally, it provides the basis for the breakpoint functionality, and manages some backend functionalities such as the temporary directory, the experiment ID, etc...

### `TrainingPlanWorkflow`

This class exists because, in the future, I assume we may wish to introduce new experiment types that will use a training plan, but have a different "execution workflow" than the current `Experiment'.

A `TrainingPlanWorkflow` is an abstract entry point to orchestrate an experiment which uses a training plan.
The `TrainingPlanWorkflow` is responsible for **managing the life-cycle of the training plan and model parameters**.

Design point: Try to avoid duplication and redundancy. Specifically, the `self._training_plan` object contains the source of truth for the model parameters (e.g. the last-aggregated values).

#### Changes in public interface 

It is no longer possible to define a training plan via a file. The only accepted way is to define a class. 
As a consequences, the argument `training_plan_class` can no longer be of type `str`.

The `__training_plan_class` and `__training_plan` attributes are made private because their life-cycle management is complicated and *MUST* be handled through the provided functions. In particular:
- it is not allowed to set the `training_plan` directly, only `training_plan_class` can be set
- when `training_plan_class` is set, the `training_plan` is automatically reset a new instance, and the model parameters are re-initialized

## New class hierarchy for `Job`

We now have `Job` > (`TrainingJob`, `TrainingPlanApprovalJob`, ...)
`Job` classes are a representation of a task that is executed on the node side. Multiple job classes can be used by one workflow, according to its needs.  

`Job` classes still have knowledge 

**Important novelty:** Job classes have a life-cycle equivalent to the task, i.e. are destroyed when the task is finished.

### `Job`

Job is now a bare-bones base class with minimal functionality:
- it instantiates `Requests`
- it manages the `keep_files_dir` temporary directory 

In particular, this means that a new temp directory will be created and destroyed for each job. If this is too much overhead, we can move this logic to the FederatedWorkflow class instead.

### `TrainingJob` 

The TrainingJob is a task for training an ML model on the nodes by executing a training plan.
It mainly implements the `start_nodes_training_round` method, which used to be part of `Job`. 

### `TrainingPlanApprovalJob` 

Another example of how future `Job` classes may look like. 

## New Federated Workflows module

The new `fedbiomed.researcher.federated_workflows` module has been introduced, mainly as a way to organise the files. 

**Important** this implies that all the experiment import statements of the form `from fedbiomed.researcher.experiment import Experiment` will need to be changed. 

**Backward-compatibility**: to provide a form of backward compatibility, the `fedbiomed/researcher/experiment.py` file has been preserved. When researchers try to import it it imports the correct `Experiment` class but prints a warning. 

## Drop support for a file to initialize the `training_plan` in `Experiment` 

This is a legacy feature, that AFAIK is not used by anyone, which introduces a large degree of complexity in managing the life-cycle of the training plan in `Experiment`. Therefore I have decided to just drop it. 

The only supported way to define a training plan in the `Experiment` is now to define the `training_plan_class` **as a type**.

When the `set_training_plan_class` method is called, the underlying training_plan will be reinitialized.


ybouilla:
We might decide to change `job_id` into a different naming

fedbiomed/researcher/federated_workflows/_experiment.py
```diff
@@ -798,7 +798,7 @@ def run_once(self, increase: bool = False, test_after: bool = False) -> int:
         optim_aux_var = self._collect_optim_aux_var()
 
         # update node states when used node list has changed from one round to another
-        self._update_nodes_states_agent(before_training=True)
+        self._update_nodes_states_agent(training_nodes, before_training=True)
```
ybouilla, line 801: we might want to use the nodes specified dataset (`self._data.data()`) because you may sample some nodes for this round, but you want to consider here all Nodes available for training (`training_nodes` is a subset of `self._data.data()`) . While this issue doesnot happen with the `default_strategy`, it may be happening with strategies that involve sampling Nodes, such as random sampling

fedbiomed/researcher/federated_workflows/_experiment.py
```diff
@@ -820,11 +820,13 @@ def run_once(self, increase: bool = False, test_after: bool = False) -> int:
...
         )
 
+        nodes_that_replied = list(self._training_replies[self._round_current].keys())
+
         # NOTA: wondering if we could name aggregator_args aggr_args (created only for testing)
         # leaving aggregator_args and aggr_args to avoid side effects
 
         # update node states with node answers + when used node list has changed during the round
-        self._update_nodes_states_agent(before_training=False)
+        self._update_nodes_states_agent(nodes_that_replied, before_training=False)
```
ybouilla, line 829: should be working!

sharkovsky:
### Summary of discussion with @tkloczko 

The discussion broadly touched on several common points with the [DtkDistributed](https://gitlab.inria.fr/dtk/dtk-distributed) library.

Namely, one of the goals of both our projects is to allow users to write "sequential" code that will have a parallel execution in the background in a transparent way.

#### Data-Process-View paradigm

For refactoring, it can be useful to think in terms of the DPV paradigm: Data, Process, and View.
When analysing a class, we should ask which one of these categories it falls into. 
Also, we can think of the program's execution as a pipeline of Data -> Process -> Data -> Process -> Data -> View

#### FederatedObject

It might make sense to consider a `FederatedObject`, a base class for anything that requires interaction with the federation. Such object contains at least a Communicator (the `Requests` class in our case). It could also contain policies (e.g. the communication backend, or security policies such as encryption, etc...). In DtkDistributed it also contains a mapper but I don't think it really applies in our case (to be confirmed).

All Federated Processes, Federated Data, etc.. classes would inherit from `FederatedObject`. 

#### FederatedApplication

Another concept that would nice to have is a programmatic (python) way to instantiate a full network. 
This would allow the researcher to write something like this in their notebook

```python
from fedbiomed import FederatedApplication

app = FederatedApplication(
    node_configs = ['node.ini', 'other-node.ini'],
    node_params={ .... to be defined .... },
    researcher_config='researcher_config.ini',
)
```

which would automatically start, and manage the life-cycle of, the nodes and the researcher. 

In my opinion this would be particularly suitable for "simulated federated learning", but not so much for the real-world deployment. Regardless, it would be a very nice feature to have that would greatly simplify the development as well as the barrier-to-entry for newcomers.

fedbiomed/researcher/federated_workflows/_federated_workflow.py
```diff
@@ -0,0 +1,818 @@
...
+
+            # If it is empty list
+            tags_to_set = tags
+
+        elif isinstance(tags, str):
+            tags_to_set = [tags]
+        else:
+            msg = f"{ErrorNumbers.FB410.value} `tags` must be a non-empty str, " \
+                "a non-empty list of str"
+            logger.cirtical(msg)
```
sharkovsky, line 419: Typo

PR #958, opened 2024-01-11
Updating download_and_split_ixi.py script

ybouilla:
**PR description**

Due to update of config file in recent version of Fed-BioMed,  `download_and_split_ixi.py `  script used in `mdecial-folder-dataset` notebooks examples does not work anymore.
This patch intents to correct this script by using the new class `NodeObject`, so it can work with both current and future versions of Fed-BioMed.

This fix implied modifying the `NodeConfig` class (adding a setter), which may not be mandatory (it was done in the previous script so i decided to keep it in the fix)

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:


mvesin:
Thank you @ybouilla
I pushed a few proposed fixes, can you please give your advice regarding them:
- fix pre-existing typo in the script (`ixi-train` not `bids-train` as tag)
- handling properly the case where the file does not exist: we need `NodeConfig(... auto_generate=False)` to avoid creating the file at this step. If not, the file is written when calling `NodeConfig()` and we always have a warning that the file already exists (though it did not exist before running the script)
- handling properly the case where the file exists: we need a `else:` when the file exists and we do not force overwriting
- need to copy the modifs to `docs/tutorials/medical/download_and_split_ixi.py` (currently duplicated due to duplication notebooks/tutorials)

It would also be nice if @srcansiz or @clebreto can give advice about adding a `config.set()` function as they are also modifying the configuration handling in parallel

mvesin (approved):

PR #963, opened 2024-01-16
Feature/protoc improvment

j-l-s:
**PR description**

TO_BE_FILLED_BY_PR_CREATOR

- added a conda env for macosx-arm (grpcio version update)
- robustyfied scripts/protoc.py

### DIFF
--- a/envs/development/conda/fedbiomed-researcher-macosx-arm.yaml
+++ b/envs/development/conda/fedbiomed-researcher-macosx-arm.yaml
@@ -1,0 +1,74 @@
+#
+# environment for fedbiomed-researcher on macosx
+#
+#
+#
+name: fedbiomed-researcher
+
+channels:
+  - conda-forge
+
+dependencies:
+  # minimal environment
+  - python >=3.10,<3.11
+  - pip >= 23.0
+  # TODO: consider migrating for classical notebooks to notebook7 or jupyterlab
+  # https://jupyter-notebook.readthedocs.io/en/latest/migrate_to_notebook7.html
+  - notebook ~=6.5.0
+  - jupyter ~=1.0.0
+  - ipython ~=8.13.2
+  # tests
+  - pytest ~=7.2.0
+  - pytest-cov ~=4.1.0
+  - tinydb ~=4.7.1
+  - tabulate >=0.9.0,<0.10.0
+  # tools
+  - colorama >=0.4.6,<0.5
+  # code
+  - psutil ~=5.9.0
+  - requests ~=2.29.0
+  - validators >=0.20.0,<0.21.0
+  - git ~=2.40.1
+  - packaging ~=23.1
+  # these two have to be aligned
+  - cryptography ~=40.0.0
+  - pyopenssl ~=23.1.1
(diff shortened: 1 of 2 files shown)


envs/development/conda/fedbiomed-researcher-macosx-arm.yaml
```diff
@@ -0,0 +1,74 @@
...
+      # for nbconvert
+      - jupyter-contrib-nbextensions >=0.7.0,<0.8.0
+      - pathvalidate ~=3.0.0
+      # FLamby
+      - git+https://github.com/owkin/FLamby@main
+      # declearn
+      - declearn[torch1] ~=2.3.2
+      - gmpy2 >=2.1,< 2.2
+      # grpc
+      - grpcio
```
mvesin, line 70: Thank you @j-l-s 

Which version did you have to use for MacOS w/ ARM ?

It would be nice to:
- keep major/minor version specification in the yaml file (to have a future-safe yaml file, when `grpcio` versions on conda-forge evolve)
- if possible: test this version on MacOS Intel + Linux to see if we can use the same `grpcio` major/minor for all platforms

scripts/protoc.py
```diff
@@ -1,11 +1,43 @@
...
+    os.chdir(proto_dir)
+
+    # removing old proto files
+    for f in glob.glob("*.py"):
+        print(f"- cleaning: {f}")
+        os.remove(f)
+    for f in glob.glob("*.pyi"):
+        print(f"- cleaning: {f}")
+        os.remove(f)
+    shutil.rmtree('__pycache__', ignore_errors=True)
```
srcansiz, line 36: Hi @j-l-s, 

It is nice to print outputs but it is not necessary to clean `__pycache__` (they are not committed) and old `py` files. Modules are overwritten by `compile_proto_files`.  So I would keep only `Refreshing: ..` part to give feedback to user`.

envs/development/conda/fedbiomed-researcher-macosx-arm.yaml
j-l-s: I released the versioning in the yaml file, and let conda/mamba deal with the dependancies. Here are the installed version:
```
$ conda list | grep grpc
grpcio                    1.60.0                   pypi_0    pypi
grpcio-tools              1.60.0                   pypi_0    pypi
```

mvesin (approved):

scripts/protoc.py
mvesin: hi @srcansiz @j-l-s can I reviewed the part about conda envs. When you complete the part about protoc, you can merge.

srcansiz: Hi @mvesin, 
Thank you. We've checked the issue that @j-l-s  is having on his machine, and the problem is not really about the `grpcio` or `grpc-tools` version. It works only when the version is not specified, and does not work if any version is specified.  So, I don't think that it is necessary to keep `...-arm.yaml` file for conda. However, it is nice to have upgraded version of grpc.

srcansiz (approved):
Hi @j-l-s abnd @mvesin, 

Changes are okay for me too.

Issue #992, opened 2024-02-13
[BUG]: cannot save breakpoints when using `declearn` optimizers


ybouilla:
Bug spotted when saving breakpoint  using `declearn` optimizers that use auxiliary variables. An error is triggered:

```
Fed-BioMed researcher stopped due to exception:
FB413: cannot save or load breakpoint for experiment - save failed with message Object of type TorchVector is not JSON serializable
```

```
Object of type TorchVector is not JSON serializable

2024-02-13 11:46:55,283 fedbiomed CRITICAL - FB413: cannot save or load breakpoint for experiment - save failed with message Object of type TorchVector is not JSON serializable
```

ybouilla:
Could reproduce error  on `master` branch

ybouilla:
**Update**: after investigation, it seems the auxiliary variables are saved in the `training_replies` object, and kept for several Rounds. However, when saving breakpoints, it is saved assuming all parameters are JSON serializable, but it is not the case when saving optimizers that use auxiliary variables.

PR #993, opened 2024-02-14
Fix for declearn optimizer not able save breakpoint

ybouilla:
**PR description**

Proposes a fix for the bug described in #992 

`TrainingReplies` are saved in breakpoints, keeping copies of what is sent back by Nodes. However, they may contain auxiliary_variables parameters received from `Nodes` that are non JSON serializable.
This fix proposes to save these auxiliary variables and non JSON serializable parameters into files, so we keep consistency with the notion of `TrainingReplies`.

-  There is one file for node auxiliary variable parameters at each `Round`: we may to optimize the way they are saved in order to reduce the number of files we have
 - the auxiliary variable parameters are converted into a file during the auxiliary variable data extraction process. This might not be suitable with the purpose of the method (which is bsically to extract data, not to transform data)
 - auxiliary variable data are converted when processing data on the `Researcher` side, decreasing the amount of data stored in the memory (a path is stored instead of a Tensor)

*Example*:  optim_aux_var are converted and saved at each `Round` in this format:
```
{'NODE_1234': {'scaffold': {'state': TorchVector with 8 coefs:
       ... }},
 'NODE_4321'': {'scaffold': {'state': TorchVector with 8 coefs:
       ...}}
}
``` 

We might decide at some point whether or not we keep or not these pieces of information in the `TrainingReplies`
There is no test for this bug, since a complete refactor is coming soon, but please advise and ask me to implement one if needed. Also, I was wondering if we needed a method to revert back optimizer auxiliary variables (not implemented yet)

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/fedbiomed/researcher/job.py
+++ b/fedbiomed/researcher/job.py
@@ -227,6 +227,7 @@ class Job:
         """
 
         self._training_replies[round_] = {}
+        print("REPLY", replies)
 
         # Loops over errors
         for node_id, error in errors.items():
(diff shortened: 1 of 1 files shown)


mvesin:
Able to reproduce with `develop` branch using https://fedbiomed.org/latest/tutorials/optimizers/01-fedopt-and-scaffold/ paragraph 4 (scaffold)
+ adding `exp.set_save_breakpoints(True)`

PR #1005, opened 2024-02-23
Hotfix/1003 fix after training params

### LINKED ISSUE #1003: Persistent buffers missing in breakpoints when using SecAgg
# Description

Some layers are not being saved correctly in the breakpoint when using SecAgg. 
This leads to a `KeyError` when trying to reload the aggregated parameters from the breakpoint. 

## To reproduce

Modify the `Net`definition in the general-tutorial-secagg notebook to include a batch normalization layer as follows:
```python
    
    class Net(nn.Module):
        def __init__(self, model_args):
            super().__init__()
            self.conv1 = nn.Conv2d(1, 32, 3, 1)
            self.conv2 = nn.Conv2d(32, 64, 3, 1)
            self.dropout1 = nn.Dropout(0.25)
            self.fc1 = nn.Linear(9216, 128)
            self.bn1 = nn.BatchNorm1d(128)
            self.dropout2 = nn.Dropout(0.5)
            self.fc2 = nn.Linear(128, 10)

        def forward(self, x):
            x = self.conv1(x)
            x = F.relu(x)
            x = self.conv2(x)
            x = F.relu(x)
            x = F.max_pool2d(x, 2)
            x = self.dropout1(x)
            x = torch.flatten(x, 1)
            x = self.fc1(x)
            x = F.relu(x)
            x = self.bn1(x)
            x = self.dropout2(x)
            x = self.fc2(x)


            output = F.log_softmax(x, dim=1)
            return output
```

Then save a breakpoint and try to reload the aggregated_params.

sharkovsky:
Fix #1003.

This PR fixes an issue that I encountered whereby some layers were missing from the breakpoints when using SecAgg. 

While working on this, I noticed several inconsistencies in the way that we handle model parameters: we have multiple "entry points" to obtain them (e.g. `after_training_params`, `get_weights`), as well as modifiers that the user can apply (e.g. `share_persistent_buffers`). 

I do not claim that I fixed all these inconsistencies in an elegant way, but my goal was to get the code into a working state, since I plan to review and improve all of this in PR #943 . 

The main changes that I tried to introduce are: 
1. the `share_persistent_buffers` training argument is now taken into account also in `Job._get_model_params`, even though the implementation is slightly misleading. 
2. the `flatten` method (used in SecAgg) now takes as inputs the parameters to be flattened
3. the `unflatten` method (used in SecAgg) now takes as inputs the dictionary of model parameters, to be used as template for reconstructing the parameter shapes and names from the flat vector
4. the `TorchTrainingPlan.after_training_params` method has been fixed to correctly re-use the partial results that are computed within the method itself

### DIFF
--- a/fedbiomed/common/models/_torch.py
+++ b/fedbiomed/common/models/_torch.py
@@ -4,7 +4,7 @@
 """Torch interfacing Model class."""
 
 import copy
-from typing import Dict, List
+from typing import Any, Dict, List
 
 import numpy as np
 import torch
@@ -70,29 +70,31 @@ class TorchModel(Model):
             Model weights, as a dictionary mapping parameters' names to their
                 torch tensor.
         """
+        if only_trainable:
+            parameters_iterator = self.model.named_parameters()
+        else:
+            parameters_iterator = self.model.state_dict().items()
         parameters = {
-            name: param.detach().clone()
-            for name, param in self.model.named_parameters()
-            if param.requires_grad or not only_trainable
-        }
+                name: param.detach().clone()
+                for name, param in parameters_iterator
+                if param.requires_grad or not only_trainable
+            }
         return parameters
 
-    def flatten(self) -> List[float]:
+    def flatten(self, params: Dict[str, Any]) -> List[float]:
         """Gets weights as flatten vector
 
         Returns:
             to_list: Convert np.ndarray to a list if it is True.
         """
 
-        params: List[float] = torch.nn.utils.parameters_to_vector(
-            self.model.parameters()
-        ).tolist()
-
+        params: List[float] = torch.nn.utils.parameters_to_vector(params.values()).tolist()
         return params
 
     def unflatten(
             self,
-            weights_vector: List[float]
+            weights_vector: List[float],
+            model_params: Dict[str, Any]
     ) -> Dict[str, torch.Tensor]:
         """Unflatten vectorized model weights using [`vector_to_parameters`][torch.nn.utils.vector_to_parameters]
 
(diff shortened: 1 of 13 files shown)


fedbiomed/researcher/experiment.py
```diff
@@ -1506,7 +1506,9 @@ def run_once(self, increase: bool = False, test_after: bool = False) -> int:
             )
             # FIXME: Access TorchModel through non-private getter once it is implemented
             aggregated_params: Dict[str, Union[torch.tensor, np.ndarray]] = (
-                self._job.training_plan._model.unflatten(flatten_params)
+                self._job.training_plan._model.unflatten(weights_vector=flatten_params,
+                                                         model_params=self.training_plan().after_training_params(
```
srcansiz, line 1510: Hi @sharkovsky, 
`after_training_params` is a method reserved to be used on the node side. In the beginning, it is developed to revert to keys of the model parameters after DP. However, it evolved in time :). I don't think that it is good idea to call on the researcher side since it executes some DP action that expects to be called on the node side.

srcansiz (commented):
Hi @sharkovsky 

Thank you for the PR and fixing the bug! I have only one comment for the PR please find it above.

fedbiomed/researcher/experiment.py
ybouilla: I agree, but actually there are some spots where we are using `training_plan().after_training_params`  in the Experiment

sharkovsky: I will look into this as soon as possible, I had a hunch that this might be a not ideal but I remember thinking that the alternatives were much more complicated... I'll come back to you when I have a clearer idea and can suggest a better way.

fedbiomed/common/models/_torch.py
```diff
@@ -70,50 +70,54 @@ def get_weights(
...
+        super().unflatten(weights_vector, model_params)
 
         # Copy model to make sure global model parameters won't be overwritten
         model = copy.deepcopy(self.model)
         vector = torch.as_tensor(weights_vector).type(torch.DoubleTensor)
 
         # Following operation updates model parameters of copied model object
         try:
-            torch.nn.utils.vector_to_parameters(vector, model.parameters())
+            torch.nn.utils.vector_to_parameters(vector, model_params.values())
```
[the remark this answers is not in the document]
sharkovsky: I think the GPU part is taken care of in the `training_routine` part.

PR #1007, opened 2024-02-28
Feature/ Enable SecAgg over DecLearn Optimizer AuxVar

### LINKED ISSUE #1236: Propagate GPU usage from `Fed-BioMed` to `Declearn`:
Currently, when enabling or disabling GPU in Fed-BioMed, this instruction does not propagate to `Declearn` facility for enabling/disabling GPU.
In addition, `declearn` behaviour with GPU seems inconstant: it is not clear if `Declearn` enables GPU usage automatically or manually. Hence, it seems it is possible to use `Declearn` with GPU while disabling it in `Fed-BioMed`.

pandrey-fr:
### Summary

Closes #1236.

This PR is about enabling the joint use of SecAgg and of a DecLearn-backed Optimizer, the current use-case being to enable using Scaffold with SecAgg.

In short:
- This PR updates the DecLearn dependency to use the (upcoming) 2.4 version.
- It introduces dedicated utils to handle declearn-issued optimizer auxiliary variables for SecAgg.
- It ensures that any optimizer AuxVar is encrypted and aggregate-decrypted jointly with model parameters.

### Background

This was possibly made impossible by three limitations:
- Fed-BioMed did not support preserving node-side information across rounds, so that private Scaffold states could not be properly preserved.
- The Scaffold implementation from DecLearn had the researcher detain node-wise states (and send back node-wise deltas), which was not compatible with SecAgg (where the node should only detain and require aggregated information).
- The DecLearn API had optimizer plug-ins be responsible for aggregating nodes' information, which again is not suitable for SecAgg.

These points have now been addressed:
- The first point was tackled some time ago with Fed-BioMed nodes becoming stateful.
- The other two are being addressed with the upcoming release of DecLearn 2.4.0, which introduces a new API for Optimizer auxiliary variables.
    - In the revised API, node-side optimizers emit `AuxVar` instances, which are to be aggregated by the researcher (via the addition operator, which is overloaded by AuxVar classes to implement actual aggregation rules) prior to being passed to its optimizer.
    -  In addition, `AuxVar` exposes a `prepare_for_secagg` method that enables sorting information in two categories: on the one hand, numerical data that is to be aggregated via summation and should be encrypted when using SecAgg; on the other hand, any other information that does not require encryption and should be aggregated with the custom rules from its source AuxVar type.

#### Changes under `fedbiomed.common.optimizers`

The new DecLearn `AuxVar` generic type and `ScaffoldAuxVar` concrete subtype are now re-exported from `fedbiomed.common.optimizers.declearn`.

The new `fedbiomed.common.optimizers._secagg` private submodule implements utils (re-exported publicly from `fedbiomed.common.optimizers` that specifically tackle SecAgg over `AuxVar` instances:
  - `flatten_auxvar_for_secagg` transforms a dict of auxiliary variables (as output by `Optimizer.get_aux`) into a flat list of scalar values that need encryption/sum-decryption, and cleartext specifications and data.
  - `EncryptedAuxVar` is a generic data structure designed to hold encrypted scalar values and cleartext specifications and data. It overloads the addition operator to accumulate lists of encrypted values and aggregate cleartext data, based on wrapped specifications.
  - `unflatten_auxvar_after_secagg` transforms a flat list of aggregate-decrypted scalar values and cleartext specifications and data back into a dict of auxiliary variables (as expected by `Optimizer.set_aux` / as would have resulted from the cleartext aggregation of nodes' dicts of auxiliary variables).

`Optimizer` signatures regarding auxiliary variables were updated to reflect the use of `AuxVar` instances instead of dict previously. Furthermore, complexities about node-specific information were dropped in DecLearn 2.4.

#### Changes under `fedbiomed.node.round.Round`

Backend routines regarding `AuxVar` processing and collection were marginally updated.

When SecAgg is active, if `AuxVar` are to be sent to the researcher, they are flattened, values that need encryption are temporarily concatenated with model parameters to be processed via a unique `SecaggCrypter.encrypt` call, and then they are separated again and wrapped into an `EncryptedAuxVar` instance, a dict dump of which is passed in the resulting `TrainReply`.

I refactored some code, introducing private methods only, for the mere sake of readability.

#### Changes under `fedbiomed.researcher.job.Job`

`Job.extract_received_optimizer_aux_var_from_round` was revised. Its signature has changed to account for the new `AuxVar` and `EncryptedAuxVar` types, but its intention remains the same:
  - Collect any auxiliary variables sent from nodes.
  - Save collected data into a serialized dump file.
  - Replace collected data with a path to this file.
  - Return a dict mapping nodes' names to the auxiliary variables they sent.

The `encrypted` flag of nodes' replies to infer whether received auxiliary variables should be parsed into `EncryptedAuxVar` instances, but verification for coherence across nodes is delayed to other parts of the code.

#### Changes under `fedbiomed.researcher.experiment.Experiment`

A number of private methods were revised, replaced or introduced to achieve new behaviors as part of the main `run_once` method. Changes in the process are the following:
  - Auxiliary variables are extracted from nodes' replies soon after model parameters are. The verification for incoherence with the type of optimizer used is still there.
  - Model parameters and auxiliary variables are aggregated at the same time (using dedicated refactored private methods that differ depending on whether SecAgg is active or not).
  - In cleartext mode, model parameters are aggregated as usual, while auxiliary variables (if any) are now aggregated directly by `Experiment` prior to being passed to its `Optimizer`.
  - In SecAgg mode, model parameters and auxiliary variables are temporarily concatenated so that their aggregation-decryption occurs via a unique `SecaggCrypter.aggregate` call. They are then separated again, and auxiliary variables are unflattened into a dict of auxiliary variables that is finally passes to `Optimzier.set_aux`.

Another (more marginal) change is that researcher-emitted auxiliary variables can no longer be addressed to specific nodes, enabling to strip boilerplate code (including some legacy code dating from before the change of network communication technology). The researcher will now simply collect (optional) auxiliary variables and send them to all nodes that participate in the training round.

#### Other side changes

Messages that transmit auxiliary variables (namely, `TrainRequest` and `TrainReply`) were updated as to the typing of this field.
- `TrainRequest` now specifies `aux_var` as an optional dict (technically, `Optional[Dict[str, AuxVar]]`).
- `TrainReply` now specifies `aux_var` as an optional dict (technically, `Optional[Dict[str, AuxVar]]` when `encrypted=False`, and `Optional[Dict[str, Any]]` otherwise - which is in fact a `EncryptedAuxVar` dict dump).

The Fed-BioMed `Serializer` was extended to cover additional types:
  - DecLearn `AuxVar`
  - DecLearn `VectorSpec`
  - Fed-BioMed `EncryptedAuxVar`
  - built-in `set`

#### Added or revised tests

- Unit and basic functional tests were implemented to cover the new utils, under the new `test_optimizer_secagg.py` file.
- Unit tests were added for new `Serilializer`-supported types.
- A test was added for `Round.run_model_training` when using SecAgg and auxiliary variables. It is not as light as one could hope, due to the complexity of the method being tested, but is works.
- A test was added for `Job.extract_received_optimizer_aux_var_from_round` when nodes send encrypted data.
- Existing unit tests for all modified classes were revised to account for API changes, and verified to pass.

#### Broken tests

A couple of tests for `Experiment` are currently broken, due to the Scaffold implementation having been completely revised.
IMHO these tests have little utility, and at any rate they are too complex (and too deeply broken) for me to revise them in their current state. I think they should either be dropped or rewritten from scratch.

#### Missing tests

At the moment, I have not written nor run any integration test that would verify that the overall process runs properly, and/or that a model trained with SecAgg + Scaffold converges similarly to a model with cleartext Scaffold.

#### Management of time indices for Joye-Libert

Since the Joye-Libert algorithm used for SecAgg is based on time indices, processing of model parameters and optimizer auxiliary variables were merged to ensure there was no re-use of time indices across them. This is somewhat impractical, because model parameters are set to be averaged, while optimizer auxiliary variables are specified to require summation.

The current workaround is to divide nodes' values by local weight multiplier prior to encryption, and to multiply decrypted aggregates by the global averaging divisor, but this introduces unnecessary operations.

Furthermore, I believe that the "do not re-use time indices" rule is completely broken anyway in the current implementation:
- (a) researcher-set "encryption factors" are encrypted using the same time index as the first few private parameters
- (b) since time indices are defined as `[round_index : round_index + n_parameters]`, all but one time indices overlap across rounds

To me, a sensible change would be to have `SecaggCrypter` maintain an internal counter for iterations, preventing time-indices re-use. This, or any way to properly set the starting time index, could result in separating encryption/decryption of model parameters and optimizer auxiliary variables, which could make for easier-to-read refactored code.


pandrey-fr:
Note: tests are bound to fail for now, because DecLearn 2.4.0 was not released yet.

You may test things locally by installing from our [pre-release branch](https://gitlab.inria.fr/magnet/declearn/declearn2/-/tree/prerelease-2.4), which I have been using for development. I am aiming to finalize this release by mid-March at the latest, but would like to have the opportunity to alter planned API changes based on your feedback as to the current PR.

pandrey-fr:
DecLearn 2.4.0 has now been [released](https://gitlab.inria.fr/magnet/declearn/declearn2/-/releases/v2.4.0); this PR should therefore be easier to review once time comes.

srcansiz:
> DecLearn 2.4.0 has now been [released](https://gitlab.inria.fr/magnet/declearn/declearn2/-/releases/v2.4.0); this PR should therefore be easier to review once time comes.

Thank you @pandrey-fr !

pandrey-fr:
As agreed with the team, since Fed-BioMed has evolved quite a lot with the refactoring of Experiment and Job, I updated this PR by starting afresh from the development branch and iteratively cherry-picking and amending commits from my initial submission.

To prevent information loss, the "legacy" implementation can be found on [this branch](https://github.com/fedbiomed/fedbiomed/tree/attic/secagg-over-declearn-aux-var-old), while I force-pushed the new code on this branch to avoid duplicating the PR.

@srcansiz @ybouilla you may now review the code and tests whenever you have the time. Note that I am yet to run some proper integration test (and I suspect that I might run into trouble due to encrypted integers packing... I had to use the new `num_expected_params` of `SecaggCrypter.aggregate` in my tests, but as far as I can tell, it is not actually used in `Experiment`).

srcansiz:
> As agreed with the team, since Fed-BioMed has evolved quite a lot with the refactoring of Experiment and Job, I updated this PR by starting afresh from the development branch and iteratively cherry-picking and amending commits from my initial submission.
> 
> To prevent information loss, the "legacy" implementation can be found on [this branch](https://github.com/fedbiomed/fedbiomed/tree/attic/secagg-over-declearn-aux-var-old), while I force-pushed the new code on this branch to avoid duplicating the PR.
> 
> @srcansiz @ybouilla you may now review the code and tests whenever you have the time. Note that I am yet to run some proper integration test (and I suspect that I might run into trouble due to encrypted integers packing... I had to use the new `num_expected_params` of `SecaggCrypter.aggregate` in my tests, but as far as I can tell, it is not actually used in `Experiment`).

Hi @pandrey-fr, 

Thank you very much for updating the PR. We are going to continue the review as soon as possible respecting the priorities of the next sprint.

pandrey-fr:
I tried to run a modified version of the MNIST notebook to test this PR's features.
After fixing an error in the code, I was able to run MNIST using a DecLearn-backed Scaffold without SecAgg.

I then tried to run the same thing with SecAgg, but it failed when verifying the encryption factors.
This error is in fact not linked to the current PR: it occurs similarly when disabling the use of a DecLearn optimizer, and even when switching to the develop branch. I have no clue where it comes from, but SecAgg is seemingly broken.

Note that I managed to patch unit tests to pass the CI, but also encountered issues with the SecAgg there: I might be mistaken about what the `clipping_range` parameter means, but I would expect that increasing it would prevent overflow errors (due to actual value clipping) while sacrificing a light amount of precision when un-quantizing values (due to the integer field mapping a wider amount of floats). However, it turns out that increasing the parameter strongly degrades the reconstruction (ending up with large values that have nothing to do with the expected one). I seem to have the same problem when trying to use SecAgg on MNIST (the encryption factor is reconstructed to a large value, that increases if I try to manually specify some large clipping range). I do not know where this comes from (and again, I may be mistaken about the parameter's usage), but to me it feels like there is something wrong (either an actual error, a dangerous instability, a wrongful default value or something unproperly documented) with the current SecAgg backend.

pandrey-fr:
Now that I managed to have SecAgg work over model parameters, I was able to debug SecAgg over both model parameters and optimizer auxiliary variables. Formally, the code now works - I was able to run the SecAgg tutorial with a modified TrainingPlan that uses a DecLearn Optimizer with Scaffold.

I have not verified that computations are correct (formally, they are; but I have not verified possible side effects of clipping, quantization, and scaling tricks) / achieve convergence. To do so I would require a heterogeneous split of MNIST anyway (Scaffold over homogeneous clients is bound to converge to having zero-valued correction states).

There are still unit and functional tests to write, but I leave that up for later, once the FBM dev team has time to come back to this PR (that probably will require a new iteration to take SecAgg API changes into account).

fedbiomed/common/optimizers/declearn.py
```diff
@@ -14,12 +14,14 @@
 )
 
 from declearn.optimizer.modules import (
+    AuxVar,
     OptiModule,
     AdaGradModule,
```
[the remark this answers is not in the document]
pandrey-fr: I added them here so that in tests we can import them from `fedbiomed` rather than from `declearn`. I have not strong opinion about it, but it felt coherent with the fact that other declearn-defined classes are imported from fedbiomed (that explicitly filters what is available).

fedbiomed/researcher/experiment.py
```diff
@@ -1554,39 +1551,224 @@ def run_once(self, increase: bool = False, test_after: bool = False) -> int:
...
+            FedbiomedExperimentError: If any node sent unproper-type data.
+        """
+        failed = []  # type: List[str]
+        auxvar = {}  # type: Dict[str, AuxVar]
+        for node_name, node_dict in nodes_auxvar.values():
+            if not isinstance(node_dict, dict):
+                failed.append(node_name)
+            elif not failed:
+                for mod_name, mod_auxv in node_dict.items():
+                    auxvar[mod_name] = auxvar.get(mod_name, 0) + mod_auxv
```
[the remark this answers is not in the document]
pandrey-fr: No, because this is using a trick from the `AuxVar.__add__` method (inherited from its parent `declearn.utils.Aggregate`) that makes it so that a `0` integer is treated as a valid value, and merely means ("return the AuxVar value, discarding the added 0").

Another way to write line 1659 would be:
```python
if mod_name in auxvar:
    aux_var[mod_name] = aux_var[mod_name] + mod_auxv
else:
    aux_var[mod_name] = mod_auxv
```

tests/test_generic_optimizer.py
```diff
@@ -702,225 +540,8 @@ def test_declearnoptimizer_09_declearn_optimizers_2_torchmodel(self):
                 loss.backward()
                 optim_w.step()
 
-    
-    def test_declearnoptimizer_10_declearn_scaffold_1_sklearnModel(self):
```
ybouilla, line 706: I am ok to remove such tests, but i think we need some integration tests. current end2end tests doesnot test accurately values for scaffold

envs/development/conda/fedbiomed-gui.yaml
```diff
@@ -51,5 +51,5 @@ dependencies:
       # FLamby
       - git+https://github.com/owkin/FLamby@main
       # declearn
-      - declearn[torch1] ~=2.3.2
+      - declearn[torch1] ~=2.5.0
```
ybouilla, line 54: now that declearn v.2.6 has been released, could we specify it? or does it break compatibility with Fed-BioMed?

fedbiomed/node/round.py
```diff
@@ -339,25 +346,17 @@ def run_model_training(
...
-                            researcher_id=self.researcher_id)
-
-                model_weights = self._secure_aggregation.scheme.encrypt(
-                    params=model_weights,
-                    current_round=self._round,
-                    weight=results['sample_size'],
+                model_weights, enc_factor, aux_var = self._encrypt_weights_and_auxvar(
+                    model_weights=model_weights,
+                    optim_aux_var=results["optim_aux_var"],
+                    secagg_arguments=secagg_arguments,
```
ybouilla, line 352: `secagg_arguments` passed but unused (see `_encrypt_weights_and_auxvar`)

fedbiomed/node/round.py
```diff
@@ -383,14 +382,88 @@ def run_model_training(
             # Only for validation
             return self._send_round_reply(success=True)
 
+    def _encrypt_weights_and_auxvar(
+        self,
+        model_weights: List[float],
+        optim_aux_var: Dict[str, AuxVar],
+        secagg_arguments: Dict[str, Any],
```
ybouilla, line 389: `secagg_arguments` passed but unused

tests/test_optimizer_secagg.py
```diff
@@ -0,0 +1,380 @@
...
+        assert isinstance(cleartext, list)
+        assert all(isinstance(x, (dict, type(None))) for x in cleartext)
+        assert len(cleartext) == len(aux_var)
+        # Assert that 'clear_cls' is a list of module-wise (name, type) tuples.
+        assert isinstance(clear_cls, list)
+        assert len(clear_cls) == len(aux_var)
+        assert clear_cls == [(key, type(val)) for key, val in aux_var.items()]
+
+    def test_flatten_auxvar_simple(self) -> None:
+        """Test flattening a dict with a single 'SimpleAuxVar' instance."""
```
ybouilla, line 118: these unit tests are great, buit they don't respect the fbm conventions for unit tests: expecially, method name should be numbered

tests/test_optimizer_secagg.py
```diff
@@ -0,0 +1,380 @@
...
+    @staticmethod
+    def perform_secure_aggregation(
+        aux_a: Dict[str, AuxVar],
+        aux_b: Dict[str, AuxVar],
+    ) -> Dict[str, AuxVar]:
+        """Perform Secure Aggregation of Optimizer auxiliary variables."""
+        # Set up a SecAgg crypter and private and public parameters.
+        biprime = int.from_bytes(  # 1024-bits biprime number
+            b'\xe2+!\x9a\xdc\xc3.\xcaY\x1b\xd6\xfdH\xfc1\xaeG6\xc0O\xa5\x9a'
+            b'\x8bi)i \xac=\x88\xb5\xfdo\xac\xadS\x80\xb3xL\xa6\xc7\xca]\x17'
```
ybouilla, line 292: here: it could be great to test with more than 2 nodes

fedbiomed/researcher/federated_workflows/_experiment.py
```diff
@@ -847,36 +865,159 @@ def run_once(self, increase: bool = False, test_after: bool = False) -> int:
...
+        # Gather parameters that need aggregate-decryption.
+        encrypted = model_params
+        encrypted_auxvar = None  # type: Optional[EncryptedAuxVar]
+        aggregated_auxvar = None  # type: Optional[Dict[str, AuxVar]]
+        n_aux_var = 0
+
+        if optim_auxvar:
+            # Ensure auxiliary variables have the same node-order as model
+            # parameters, type-check and pre-aggregate them.
+            encrypted_auxvar = self._preaggregate_encrypted_optim_auxvar(
```
ybouilla, line 944: I fail to understand the whole process here. So first auxvar are preaggregated and then  aggregated ? 
What is the difference here?
It is really confusing and not very straightforward in my opinion

fedbiomed/researcher/federated_workflows/_experiment.py
```diff
@@ -847,36 +865,159 @@ def run_once(self, increase: bool = False, test_after: bool = False) -> int:
...
+                "Received unproper-type optimizer auxiliary variables from "
+                "1+ nodes.\nNodes that sent invalid data are the following: "
+                f"{failed}."
+            )
+        return auxvar
+
+    def _aggregate_encrypted_model_params_and_optim_auxvar(
+        self,
+        model_params: Dict[str, List[int]],
+        optim_auxvar: Dict[str, EncryptedAuxVar],
```
ybouilla, line 911: This method is in charge of too much things. Moreover, it should be split across `federated_workflow`, for other uses (think of federated analytics)

ybouilla:
Remaining Work to be done:
- find a way to encrypt model weights and auxvar using the same encrypt method
- try with more than 2 nodes
- implementation wise: be consistent: secagg needs model weights to be averaged before aggregation  but un-encrypted training doesnot

fedbiomed/common/optimizers/_secagg.py
```diff
@@ -0,0 +1,388 @@
...
+    @classmethod
+    def concatenate_from_dict(cls, data: Dict[str, Self]) -> Self:
+        auxvar = list(data.values())
+        # this converts List[List[List[int]]] -> List[List[int]]
+        obj = sum(auxvar[1:], start=auxvar[0])
+        return cls(obj.encrypted, obj.enc_specs, obj.cleartext, obj.clear_cls)
+
+    def get_mapping_encrypted_aux_var(self) -> Dict[str, List[int]]:
+        nodes_id = list(self.cleartext[0]['clients'])
+        return {n: p for n,p in zip(nodes_id, self.encrypted)}
```
ybouilla, line 332: check if nodes_id maps the actual encrypted values

ybouilla: does work

srcansiz (commented):
Hi @ybouilla,

Thank you for the modification. The PR is good to go for me. I have executed a few e2e tests. However, I couldn't test the part for GPU parameter changes. So I will rely on your tests.  
I have a few minor comments.

fedbiomed/common/optimizers/_secagg.py
```diff
@@ -0,0 +1,388 @@
+# This file is originally part of Fed-BioMed
```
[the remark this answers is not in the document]
ybouilla: I dont really understand... Objects of this module are currently displayed  in the `optimizer` section (ie http://127.0.0.1:8000/latest/developer/api/common/optimizers/#fedbiomed.common.optimizers.EncryptedAuxVar)

fedbiomed/common/serializer.py
```diff
@@ -108,6 +111,8 @@ def _default(obj: Any) -> Any:
             }
         if isinstance(obj, tuple):
             return {"__type__": "tuple", "value": list(obj)}
+        if isinstance(obj, set):
+            return {"__type__": "set", "value": list(obj)}
```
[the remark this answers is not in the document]
srcansiz: Ok thank you @ybouilla

fedbiomed/common/optimizers/_secagg.py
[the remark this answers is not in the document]
srcansiz: Ahh yes sorry. I thought that  it has to be declared another module.

fedbiomed/node/round.py
```diff
@@ -383,14 +383,92 @@ def run_model_training(
...
+            logger.info(
+                "Encrypting model parameters and optimizer auxiliary variables."
+                "This process can take some time depending on model size.",
+                researcher_id=self.researcher_id,
+            )
+            # Flatten optimizer auxiliary variables and divide them by scaling weights.
+            cryptable, enc_specs, cleartext, clear_cls = (
+                flatten_auxvar_for_secagg(optim_aux_var)
+            )
+            #cryptable = [x / sample_size for x in cryptable] # ?? already done while encrypting
```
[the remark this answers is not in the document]
ybouilla: Done! see #1250

PR #1039, opened 2024-03-12
Revision of Job classes

ybouilla:
**PR description**

Revision of job classes. Main additions / changes

- addition of an object for timers
- changes in some attributes
- addition of docstrings

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:


fedbiomed/researcher/federated_workflows/jobs/_training_job.py
```diff
@@ -176,16 +182,23 @@ def execute(self) -> Tuple[Dict, Optional[Dict]]:
...
+                replies = federated_req.replies()
+
+        training_replies = self._get_training_results(replies=replies,
+                                                      errors=errors)
+
+        timing_results = self._get_timing_results(replies, timer)
+        # `training_replies` can be empty if there wasnot any replies
+        for node_id in replies:
+            if training_replies.get(node_id):
+                training_replies[node_id].update({'timing': timing_results[node_id]})
```
sharkovsky, line 201: Looking at the original implementation, the function `self._get_training_testing_results` was indented inside the `reqs.send` context. It feels like it should probably be indented similarly, but here it is outside. 
I'm not an expert on the implementation, @srcansiz do you have any thoughts? Is it ok to indent outside or should it be inside the `reqs.send`? 

```suggestion
            training_replies = self._get_training_results(replies=replies,
                                                                                     errors=errors)

            timing_results = self._get_timing_results(replies, timer)
            # `training_replies` can be empty if there wasnot any replies
            for node_id in replies:
                if training_replies.get(node_id):
                    training_replies[node_id].update({'timing': timing_results[node_id]})
```

fedbiomed/researcher/federated_workflows/jobs/_training_job.py
```diff
@@ -46,7 +46,8 @@ def __init__(self,
         Args:
             nodes: A dict of node_id containing the nodes used for training
             keep_files_dir: Directory for storing files created by the job that we want to keep beyond the execution
-                of the job. Defaults to None, files are not kept after the end of the job.
+                of the job. Defaults to None, files are not kept after the end of the job. Usually, files are kept in
```
sharkovsky, line 49: We are changing a bit the semantics for `keep_files_dir` by removing the default value. 

From the docstring (although I don't know if it was actually true in the code) it seems that `keep_files_dir=None` had the meaning of  "do not keep files at the end of Job". However, I'm a bit confused if this even makes sense or not. 

@mvesin are you the most up-to-date person on this? Do you remember if `keep_files_dir=None` is actually implemented and used somewhere? Is it ok if we remove this default value, and we force it to be a `str`?

sharkovsky (changes requested):

fedbiomed/researcher/federated_workflows/jobs/_training_job.py
ybouilla: For that I think we were adding extra iterations for nothing. but ok, let's wait for others inputs

fedbiomed/researcher/federated_workflows/jobs/_training_job.py
ybouilla: It is what we have discussed and decided  with mvesin: `keep_files_dir` should be used each time a `job` is created and then used. there is no place `keep_files_dir=None` is actually used: each time we are passing a path and are saving things within it,

fedbiomed/researcher/federated_workflows/jobs/_job.py
```diff
@@ -27,46 +25,70 @@ class Job(ABC):
...
 
     Attributes:
-        requests: read-only [`Requests`][fedbiomed.researcher.requests.Requests] object handling communication with remote nodes
+        requests: read-only [`Requests`][fedbiomed.researcher.requests.Requests] object handling communication with
+                 remote nodes
         nodes: node IDs participating in the task
     """
 
     def __init__(self,
-                 *,
```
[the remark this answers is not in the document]
mvesin: Thank you @sharkovsky 
Yes, I confirm this is intentional to have only named arguments to avoid confusion + avoid ordering constraints for named/positional arguments.
Discussing with @ybouilla : his modification is unwanted, this is a side effect of back merging in this branch

fedbiomed/researcher/federated_workflows/jobs/_training_job.py
mvesin: Thank you @sharkovsky 
I confirm this modification is intentional (and discussed with  @ybouilla )

The "do not keep the file at the end" is an obsolete option we did not dare to remove before this refactoring.

fedbiomed/researcher/federated_workflows/jobs/_training_job.py
srcansiz: Hi @sharkovsky, 
Yes you are right. Thank you for pointing this issue. This implementation calculates same `rtime_total` for each node where it should be calculated differently for each node. However, this is not introduced in this PR it was already there because i didn't think about it while I was working on `Requests` redesign. Honestly, I liked @ybouilla 's implementation (context manager for timer) I would say that we can keep it as it is for now and create a new issue to move this timer in `FederatedRequest` and `Request` classes.  Basically, we can have timer for each `Request` that will be accessible through `FedereatedRequest.timings()`.

mvesin: The solution proposed by @ybouilla and @srcansiz is also fine for me.
Thank you all for investigating (and re-explaining it to me this evening - sorry for the fuss).

Just one point to keep in mind for the separate issue: when implementing the moving of the timer in `FederatedRequest` / `Request` pay attention to this detail: the `reply.timing` only exists for `TrainReply` messages. Moreover, it is only checked that it is a `dict` when receiving the message; it is not checked that `reply.timing` has `rtime_training` or `ptime_training` keys. Therefore, as in current implementation we should not try to access these fields without checking their existence (currently we just add a `reply.timing['rtime_training']`

fedbiomed/researcher/federated_workflows/_experiment.py
```diff
@@ -835,7 +835,7 @@ def run_once(self, increase: bool = False, test_after: bool = False) -> int:
                           optim_aux_var=optim_aux_var
                           )
 
-        logger.info('Sampled nodes in round ' + str(self._round_current) + ' ' + str(job.nodes))
+        logger.info('Sampled nodes in round ' + str(self._round_current) + ' ' + str(training_nodes))
```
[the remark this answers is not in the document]
sharkovsky: Thanks for sharing this, after second thoughts I think the best option is:
- let's keep the getter, and use `job.nodes` instead of `training_nodes`
- let's remove the getter: `Job` is one shot and its members should not be modified from the outside after init

fedbiomed/researcher/federated_workflows/jobs/_training_job.py
sharkovsky: I'm a bit confused: 
- for the computation of timing, all is clear and I agree with the current solution + Sergen's proposal
- but the indentation of `training_replies = self._get_training_results(replies=replies, errors=errors)` has also been changed. @srcansiz could you confirm that this is ok?

fedbiomed/researcher/federated_workflows/_experiment.py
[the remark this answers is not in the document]
mvesin: @sharkovsky do you mean "let's keep the getter, let's remove the **setter**" ?

This option would be fine for me.

fedbiomed/researcher/federated_workflows/jobs/_training_job.py
mvesin: Let's wait for @srcansiz final word, but my understanding is that he explained (and we agreed) on friday that this was (previously) uselessly called inside `with self._reqs.send()` context: not needed as `errors` and `replies` can be called after removing the context, thus OK with both implementations but maybe cleaner with the new one.

Both are OK for me.

fedbiomed/researcher/federated_workflows/_experiment.py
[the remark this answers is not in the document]
sharkovsky: yeah that's what I meant :(

fedbiomed/researcher/federated_workflows/jobs/_training_job.py
sharkovsky: OK great! 

I didn't know whether it was needed or not and wanted to be sure to avoid mistakes. 

I aggre with @ybouilla and @mvesin that, since the original implementation was redundant, this one is better.

srcansiz: Hi @sharkovsky, 
It seems ok to me unless we want to `rtime_total` covers the time for extracting training replies which may be not necessary.

mvesin (approved):

fedbiomed/researcher/federated_workflows/_experiment.py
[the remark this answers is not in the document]
ybouilla: It should be fixed now. just let me know if it is better a property or a setter `job.nodes` or `job.nodes()`

sharkovsky (approved):

Issue #1029, opened 2024-03-11
review internal class consistency tools for `FederatedWorkflow`


sharkovsky:
- tags
- training data
- the fact that nodes was excluded from the consistency rules

also, validate the choice that `nodes` now has become `nodes_filter`, and is a fixed value set by the researcher. Therefore, it does not track anymore the nodes used in the experiment.

PR #1043, opened 2024-03-14
Review federated workflow

srcansiz:
**PR description**

Address #1029

- Shorten long lines
- Update var names used for typing according the `typing` recommendations.
- Raise correct exceptions FederatedTrainingPlan
- Corrects passing empty list to `set_tags` 
- Updates training_data at any call of `set_tags`

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/fedbiomed/researcher/federated_workflows/_federated_workflow.py
+++ b/fedbiomed/researcher/federated_workflows/_federated_workflow.py
@@ -128,8 +128,8 @@ class FederatedWorkflow(ABC):
     @exp_exceptions
     def __init__(
             self,
-            tags: Union[List[str], str, None] = None,
-            nodes: Union[List[str], None] = None,
+            tags: Optional[List[str] | str] = None,
+            nodes: Optional[List[str]] = None,
             training_data: Union[FederatedDataSet, dict, None] = None,
             training_args: Union[TrainingArgs, dict, None] = None,
             experimentation_folder: Union[str, None] = None,
@@ -186,21 +186,24 @@ class FederatedWorkflow(ABC):
         # set internal members from constructor arguments
         self.set_secagg(secagg)
 
+        # TODO: Manage tags within the FederatedDataset to avoid conflicts
+        if training_data is not None and tags is not None:
+            raise FedbiomedValueError(
+                f"{ErrorNumbers.FB410.value}: Can not set `training_data` and `tags` at the "
+                "same time. Please provide only `training_data`, or tags to search for "
+                "training data.")
+
         # Set tags if it tags is not None
         if tags:
             self.set_tags(tags)
+
+        if training_data:
+            self.set_training_data(training_data)
 
         self.set_nodes(nodes)
         self.set_save_breakpoints(save_breakpoints)
         self.set_training_args(training_args)
 
-        # set training data
-        if training_data is not None:
-            # if training data was provided, it takes precedence
-            self.set_training_data(training_data)
-        elif self.tags() is not None:
-            # if tags were provided, set training data from tags
-            self.set_training_data(training_data, True)
         self.set_experimentation_folder(experimentation_folder)
         self._node_state_agent = NodeStateAgent(list(self._fds.data().keys())
                                                 if self._fds and self._fds.data() else [])
@@ -369,21 +372,18 @@ class FederatedWorkflow(ABC):
     def set_tags(
         self,
         tags: Union[List[str], str],
-        reset_training_data: bool = True
     ) -> List[str]:
         """Sets tags and verification on argument type
 
-        If reset_training_data is True setting tags updates also training data
-        by executing [`set_training_data`][fedbiomed.researcher.federated_workflows.\
-        FederatedWorkflow.set_training_data] method.
+        Setting tags updates also training data by executing [`set_training_data`]\
+        [fedbiomed.researcher.federated_workflows.FederatedWorkflow.\
+        set_training_data] method.
 
 
         Args:
             tags: List of string with data tags or string with one data tag. Empty list
                 of tags ([]) means any dataset is accepted, it is different from None
                 (tags not set, cannot search for training_data yet).
-            reset_training_data: If True resets training data, and resetting training data
-                will publish search request in all connected nodes.
         Returns:
             List of tags that are set.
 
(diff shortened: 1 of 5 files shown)


fedbiomed/researcher/federated_workflows/_federated_workflow.py
```diff
@@ -331,95 +343,93 @@ def info(self, info=None) -> Dict[str, Any]:
...
-        - if training data is None, then the attributes are consistent
-        - if training data is not None and tags is not None, check that the tags from the training data all contain
-            the tags declared in the self._tags attribute
-
-        The state diagram below explains the flow of setters and consistency checks
-        ![state diagram](https://www.plantuml.com/plantuml/png/ZP91ZzCm48Nl_XLpHjf3usGFqIhQ0ul4XGEik22qCiviQz7OnPvK8CH_9rESAgdjqalhsFE-zsRinq3AqpZinRGWXAUV1_HcG4llhI7uBG2UlJBMsErRHUfbMb4B7vn5Fb7RiDpv8oBb4nAVFRlFQZzYIgbQE7Wy6ZS6E799XF41JVyP4Xka85a2oKoafR84l5ytTv_moy12htKB5BzV-cbZHjStezzvDx0aPJSNRFYc0lRWBBYHj1iGt2ju_35YeDctoVbUNFlTNNTnXwMAT09p4te33mzwvup6hWCXrZm6S4aBUeVwEsXdWmc4Ll_YpCJjAjl3Qn-4td1rSIej67kMKtGFv0uS06tVTP4GDrjOL9_JoZHjqbjCBMzABKfvcV5HcO1FtZlFyQFIXDFRMtHGdpjO2EPEwkiMMWgXvVeg6L-ULpMxHLtSpCvh-lMqWKbnMasQk9Cy7JOS0thuDzqLe4e0LHBucbucUfbxAbbEleRLNzvyNRdKYKkTS_b_kqq2Qgw_x3L9Y4Uq_JZi_m80)
-
-        This function has the following behaviour:
+    def set_tags(
+        self,
```
mvesin, line 370: @srcansiz thank you for the review ! Here are several points in `set_tags()` I think relevant to discuss and maybe tune

1. I'm not sure if this is a good idea to prevent to set to `None` because it breaks backward compatibility while not really needed. But if you think appropriate, we can keep it. In any case, if we prevent `None` we should update the error strings accordingly (not `... a str, a list of str, or None`).

2. what is the purpose of additional test `if not tags:` ? Most badly typed args will execute:
    ```
        else:
            raise FedbiomedTypeError(
                f"{ErrorNumbers.FB410.value} `tags` must be a str, a list of str, or None")
    ```
    ... except if we have empty list or empty string.
    So if we want to check that case then we probably should have a coherent error message eg "`tags` must be a non-empty str, or a non-empty list of str"

3. naming `reset_training_data` and corresponding docstrings: I would suggest to name it rather `update_training_data` or equivalent. The name "reset" is rather misleading as it suggests returning to some initial value.

srcansiz: Hi @mvesin, 

Thank you for your comments.

1 - It seems to me that setting tags to `None` is not a good idea or does not make any sense. It was implemented that way because experiment instantiation was allowing to set tags to `None`. And `set_training_data` (which depends on tags) raises error if the tags are `None`. So why do we allow setting tags to `None`. Anyway, you are right that it is not a good idea to break the API while it is not needed, so I can change it to raise deprecation warning when the tags are set to None.  

2 - As you said `if not tags` makes sure that the list or string is not empty. And the error message is 
```
f"{ErrorNumbers.FB410.value}: Invalid value for tags argument {tags}, tags " 
"should be list of str o str.")
```
and I agree that it is a good idea to add `non-empty list`, `non-empty str` to the message.

3 - I used `reset` term more like set again training data again using new tags. It is true that reset is conflicting with the doc string. It is okay for me to change it as `update_..`.

4 - I removed those because it seems unnecessary to me. Basically we are raising an error to catch it and printed with as nicely (thanks to exp_exception) but also calling logger.critical to show to same output. So that's why i wanted to remove it. If you want me to make it coherent I can revert it to call logger.critical again or remove all logger.critical calls and just raise an error.

fedbiomed/researcher/federated_workflows/_federated_workflow.py
```diff
@@ -502,37 +512,44 @@ def set_training_data(
...
         elif isinstance(training_data, dict):
             self._fds = FederatedDataSet(training_data)
         else:
-            msg = ErrorNumbers.FB410.value + f' `training_data` has incorrect type: {type(training_data)}'
+            msg = ErrorNumbers.FB410.value + \
+                f' `training_data` has incorrect type: {type(training_data)}'
             logger.critical(msg)
             raise FedbiomedTypeError(msg)
         # check and ensure consistency
         if not self._fds_tags_consistent():
```
mvesin, line 551: @srcansiz @sharkovsky I don't feel very comfortable with `_fds_tags_consistent()`
I would propose to remove this function and have a simple rule for coherence (if we want to maintain some tags coherence):

```
if not from_tags:
  self._tags = None
```

Arguments for this proposal:
1. the actual behaviour is very difficult to undersand for the user

2. We might have to fix some cases

For example if:
- if `tags == [ 'one' ]` and  `training_data().data() == []`
- then `_fds_tags_consistent` returns False 
- so if `from_tags == False` then we set `self._tags = None` while we might consider the tags to be coherent with the datasets (any set of tags can give an empty set of datasets ...)

3. more generally, I suspect in some cases we cannot have a coherence check (or at least it questions what "coherent" means)

For example:
- if we have `self._tags [ 'un']`
- then we `set_training_data(FederatedDataset({ 'd1': { tags = [ 'un', 'deux'] } }), from_tags = False)
- so `_fds_tags_consistent` returns True as all `self._tags` are contained in `d1` tags
- BUT if we query `set_training_data(from_tags=True) we may get something like `{ 'd1': { tags = [ 'un', 'deux'] }, 'd2': { tags = [ 'un', 'trois'] } }` so it might not be coherent with the current status of our nodes and datasets. So I'm not sure the kind of coherence provided is very interesting ...

What do you think ?

mvesin (approved):
Thank you @srcansiz, nice job. OK for me.
Would be nice if we can just keep cohence regarding `logger.critical()`

srcansiz:
> Thank you @srcansiz, nice job. OK for me. Would be nice if we can just keep cohence regarding `logger.critical()`

Thank you @mvesin, I revert exception raises back to version that calls `logger.critical`.

PR #1050, opened 2024-03-20
adds option to switch on and off init_optimizer execution

### LINKED ISSUE #1048: In the `Experiment` refactoring, `set_model_args` fails if used before `set_training_args`
Bug introduced while doing the refactoring of `Experiment` (see #896 )

if one executes all the methods to configure an `Experiment`
```
from fedbiomed.researcher.experiment import Experiment
exp = Experiment()
exp.set_training_plan_class(training_plan_class=MyTrainingPlan)
exp.set_model_args(model_args={}) 
>>> fails with following error
 File "/home/ybouilla/Documents/github/fedbiomed/fedbiomed/common/training_plans/_torchnn.py", line 257, in _configure_model_and_optimizer
    optimizer = self.init_optimizer(self._optimizer_args)
KeyError: 'lr'
```

but if one runs 
```
training_args = {
    'loader_args': { 'batch_size': 48, }, 
    'optimizer_args': {
        'lr': 1e-3, 
    },
    'epochs': 1, 
}
exp.set_training_args(training_args=training_args)
exp.set_model_args(model_args={}) 
```
`Experiment` can be run
**proposal**: after discussion with Francesco, we should add a default `lr` value (simplest solution we could find)

srcansiz:
**PR description**

Closes #1048

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/fedbiomed/common/training_plans/_torchnn.py
+++ b/fedbiomed/common/training_plans/_torchnn.py
@@ -233,8 +233,7 @@ class TorchTrainingPlan(BaseTrainingPlan, metaclass=ABCMeta):
         """Configures model and optimizer before training
 
         Args:
-            initialize_optimizer: If True configures optimizer of
-                training training plan
+            initialize_optimizer: If True configures optimizer.
         """
 
         # Message to format for unexpected argument definitions in special methods
--- a/fedbiomed/common/training_plans/_torchnn.py
+++ b/fedbiomed/common/training_plans/_torchnn.py
@@ -260,6 +260,13 @@ class TorchTrainingPlan(BaseTrainingPlan, metaclass=ABCMeta):
         self._model = TorchModel(model)
 
         # Get optimizer defined by researcher ------------------------
+        # FIXME: This is implemented to solve the issue while setting model
+        # arguments before setting training arguments on the researcher side (#1048).
+        # the execution of init_optimizer during post init requires to have
+        # optimizer_args set. However, this happens only on the researcher side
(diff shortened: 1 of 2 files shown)


ybouilla:
Hey Sergen, the PR is not related to issue https://github.com/fedbiomed/fedbiomed/issues/1048

srcansiz:
> Hi Sergen, we have discussed with Francesco, and I guess just changing TrainingArgs (as described in the issue #1048) should be enough). If the fix doesnot work then we could either:
> 
>     * add an error message
> 
>     * or do the changes you have done. I have mixed feeling wrt what you have done for that PR, i would rather go for splitting `configure_model` and `configure_optimizer` as 2 distinct methods

Hi @ybouilla, 

Thank you for the review. OK I understand.  I also tested adding default value to `TrainingArgs`, it works but does not solve the bug because the error is not because of `lr` only.  It is also existing for other possible parameters of the optimizer which means that we have to add default values for each possible optimizer argument.  If user creates a notebook with `lr` and `decay` it will fail again. Basically, adding default argument will not cover every case. If the current fix in this PR is not satisfying I would go for the option of adding error message, but it is going to be diffucult because: 

1 - We will have to implement training plan specific check because it is only for torch training plan not for sklearn.
2 - User may not provide `init_optimizer` in the training plan and prefer the to user default one instead. In that case should not check is optimizer arguments are set or not. 

So that's why in the issue #1048 I commented that solving this issue 100% may be more complicated than adding default `lr` value.

PR #1063, opened 2024-03-26
reorganize constructor arguments of federated workflow and job classes

srcansiz:
**PR description**

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/fedbiomed/researcher/federated_workflows/jobs/_training_job.py
+++ b/fedbiomed/researcher/federated_workflows/jobs/_training_job.py
@@ -22,30 +22,26 @@ class TrainingJob(Job):
     [TrainingPlan][fedbiomed.common.training_plans.BaseTrainingPlan].
     """
 
-    def __init__(self,
-                 *,
-                 nodes: Optional[List[str]] = None,
-                 keep_files_dir: str,
-                 experiment_id: str,
-                 round_: int,
-                 training_plan: BaseTrainingPlan,
-                 training_args: Union[dict, TrainingArgs],
-                 model_args: Optional[dict],
-                 data: FederatedDataSet,
-                 nodes_state_ids: Dict[str, str],
-                 aggregator_args: Dict[str, Dict[str, Any]],
-                 secagg_arguments: Union[Dict, None] = None,
-                 do_training: bool = True,
-                 optim_aux_var: Optional[Dict[str, Dict[str, Any]]] = None,
-                 ):
+    def __init__(
+        self,
+        *args,
+        experiment_id: str,
+        round_: int,
+        training_plan: BaseTrainingPlan,
+        training_args: Union[dict, TrainingArgs],
+        model_args: Optional[dict],
+        data: FederatedDataSet,
+        nodes_state_ids: Dict[str, str],
+        aggregator_args: Dict[str, Dict[str, Any]],
+        secagg_arguments: Union[Dict, None] = None,
+        do_training: bool = True,
+        optim_aux_var: Optional[Dict[str, Dict[str, Any]]] = None,
+        **kwargs
+    ):
 
         """ Constructor of the class
 
         Args:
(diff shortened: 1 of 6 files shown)


fedbiomed/researcher/federated_workflows/jobs/_training_job.py
```diff
@@ -22,30 +22,26 @@ class TrainingJob(Job):
...
-                 data: FederatedDataSet,
-                 nodes_state_ids: Dict[str, str],
-                 aggregator_args: Dict[str, Dict[str, Any]],
-                 secagg_arguments: Union[Dict, None] = None,
-                 do_training: bool = True,
-                 optim_aux_var: Optional[Dict[str, Dict[str, Any]]] = None,
-                 ):
+    def __init__(
+        self,
+        *args,
```
mvesin, line 27: @srcansiz is there any reason to re-introduce positional arguments for `TrainingPlanApproveJob` and `TrainingJob` ?

They were purposely all forced to be named arguments a few PR behind. Due to the complexity of the signature of these classes constructors, it was felt it is a good idea to force named arguments to avoid argument errors or at least make them explicit.

srcansiz: Hi @mvesin, 

It is still the same. The arguments in `TrainingJob` (or in any child class of Job) are keyword-only arguments, however, `nodes` and `keepfiles_dir` can be both positional or keyword argument.

mvesin: Thank you @srcansiz 

> It is still the same. The arguments in TrainingJob (or in any child class of Job) are keyword-only arguments, 

OK, sorry, you are right: arguments in `TrainingJob` and `TrainingPlanApproveJob` are keyword-only.

Yet I would prefer to remove the `*args` in these function as we forcibly don't want to receive any positional arguments.  This would be clearer this way. In current implementation, we transmit them to `Job` and they will fail in `Job.__init__()`

Would it be ok for you to remove the `*args` ?

> however, nodes and keepfiles_dir can be both positional or keyword argument.

No, they are stll keyword-only (and this is what we want !)

srcansiz: Hi @mvesin, 
Sorry there is a change that i forgot to push. Which removes `*` from `Job` class. By removing `*` I was able to execute. 

```
TrainingPlanApproveJob(nodes=['x'], keep_files_dir=True, training_plan=MyTrainingPlan, description='desc')
```
and 
```
TrainingPlanApproveJob(nodes=['x'], keep_files_dir=True, training_plan=MyTrainingPlan, description='desc')
```

If you add back `*` in `Job` arguments, you won't be able to execute: 

```
TrainingPlanApproveJob(nodes=['x'], keep_files_dir=True, training_plan=MyTrainingPlan, description='desc')
```
But I didn't understand why executing `nodes` and `keep_files_dir` as positional arguments adds complexity. I think it makes more sense since these two arguments are the arguments that should be existing in every child class of  `Job`, and it is mandatory to be defined by the user. However, other arguments may change from one child to another, and therefore they are keyword-only arguments. Also, executing `TrainingPlanApproveJob(nodes=['x'], keep_files_dir=True, MyTrainingPlan,  'desc')`   will explicitly raise an error that says `you have to provide keyword-only arguments `training_plan` and `description`. 

This is my point of view but I am okay to remove `*args`, and force user to define argument as keyword-only. However, I don't think that forcing user/developer to define only keyword-only will make any difference from functional and comprehensible point of the API.

srcansiz: I pushed a new commit that disables positional arguments for Job API

mvesin (approved):

PR #1071, opened 2024-03-28
Feature896/update function inconsistencies

mvesin:
**PR description**

Addresses one of the tasks listed in #896 
> cross check code for functions inconsistencies
> -    clean code from unused code/functions
> -    update/remove references to refactored methods that changed signature or were removed

Specifically we found the following points:

- re-introduced `TrainingPlanWorkflow.training_plan_file()` that was removed during refactoring, but may be useful. For example it is mentioned in training plan approval tutorials in `docs/tutorials`

- renamed `_reset_training_plan`

- converted in `TrainingPlanWorklow` variable `self.__training_plan` to `self._training_plan` (single _ ). Same for `self._training_plan_class`

- removed unused `FederatedDataSet.fiter_nodes_from_ids`

- misc: explicitely added default parameters to some calls of `BaseTrainingPlan.get_model_parameters()` => `get_model_parameters(only_trainable=False, exclude_buffers=True)` + same for `Model.get_weights()`.
The reason is: as this code existed before changing these methods' signature for handling the case of frozen layers and buffers, we want to make it clear for developpers that these calls were re-assessed after this change

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/fedbiomed/common/optimizers/generic_optimizers.py
+++ b/fedbiomed/common/optimizers/generic_optimizers.py
@@ -162,7 +162,10 @@ class DeclearnOptimizer(BaseOptimizer):
         # Therefore, it is necessary to disable the sklearn internal optimizer beforehand
         # otherwise, computation will be incorrect
         grad = declearn.model.api.Vector.build(self._model.get_gradients())
-        weights = declearn.model.api.Vector.build(self._model.get_weights())
+        weights = declearn.model.api.Vector.build(self._model.get_weights(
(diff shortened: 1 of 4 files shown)


sharkovsky (approved):
All changes look good to me. 
The only discussion we may have (in the future) is whether when a user submits a wrong argument the exception raised should be `FedbiomedExperimentError` or `FedbiomedTypeError`/`FedbiomedValueError`

PR #1079, opened 2024-04-03
completes federated workflow unittests

srcansiz:
**PR description**

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/fedbiomed/researcher/federated_workflows/_federated_workflow.py
+++ b/fedbiomed/researcher/federated_workflows/_federated_workflow.py
@@ -110,17 +110,10 @@ class FederatedWorkflow(ABC):
     Additionally, it provides the basis for the breakpoint functionality, and manages some backend functionalities such
     as the temporary directory, the experiment ID, etc...
 
-    The attributes `training_data`, `tags` and `nodes` are co-dependent. Attempting to modify one of those may result
+    The attributes `training_data` and `tags` are co-dependent. Attempting to modify one of those may result
     in side effects modifying the other, according to the following rules:
-    - setting any of those to None leaves the others untouched, potentially leaving the class in an inconsistent state
-    - modifying tags or nodes when training data is None will simply set the intended value
-    - modifying tags if training data is not None will reset the training data based on the
-        current nodes and the new tags, and afterwards the nodes will be updated according to the new training data
-    - modifying nodes if training data is not None will reset the training data based on the
-        current tags and the new nodes, and then the nodes will be updated again according to the new training data
-    - modifying the training data from tags will update the nodes, while modifying the training data from an object
-        resets the tags to None and updates the nodes
-
+    - modifying tags if training data is not None will reset the training data based on the new tags
+    - modifying the training data using a FederatedDataset object or a dict will set tags to None
     """
     @exp_exceptions
     def __init__(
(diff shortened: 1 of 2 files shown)


fedbiomed/researcher/federated_workflows/_federated_workflow.py
```diff
@@ -110,10 +110,17 @@ class FederatedWorkflow(ABC):
     Additionally, it provides the basis for the breakpoint functionality, and manages some backend functionalities such
     as the temporary directory, the experiment ID, etc...
 
-    The attributes `training_data` and `tags` are co-dependent. Attempting to modify one of those may result
+    The attributes `training_data`, `tags` and `nodes` are co-dependent. Attempting to modify one of those may result
```
mvesin, line 113: Thank you @srcansiz for this work !

Only one point for me, regarding this new comment about the `FederatedWorkflow`. Could we rediscuss it ?

1. it seems to me it's not correct to say that `setting any of those to None leaves the others untouched` : `training_data` and `tags` are maintained coherent (setting one of them keeps the other coherent).

   So I would leave the previous comment unchanged regarding `training_data` and `tags`, and treat the case of `self._nodes_filter` separately

2. you spotted a tricky point regarding the `self._nodes_filter`, thank you very much !

    @sharkovsky had warned us about some behaviour change and potential interaction, so it would be nice if he can participate to the discussion (the current changes may be intentional and I may just have missed their real consequences ...)

   From the behaviour of `set_nodes()` `all_federation_nodes()` and `filtered_federation_nodes()` and comment in `FederatedWorkflow.__init__()` arguments I would understand we intend the `self._nodes_filter` to be a fully "dynamic" notion: we can set them at any time, they need not be coherent with the current `self._fds` == they may include nodes that don't currently provide data in the `self._fds`. The nodes filter is applied each time we do a request.
   - regarding that, it's OK for me to `set_nodes()` without checking for tags or training data. I don't see it as "incoherent". We may just explain that point
   - ... but when doing a `Requests.search()` in `set_training_data` https://github.com/fedbiomed/fedbiomed/blob/7b7c7b3310f52765de34498da5af335acee3bf69/fedbiomed/researcher/federated_workflows/_federated_workflow.py#L564 we may consider that we should search on **all** nodes, not on `self._nodes_filter` to keep this notion of dynamicity (if we later change `self._nodes` we may see more datasets). So this may be considered as a bug (*)
   - we may also consider it not a bug (keep as now). Then we need to explain it, but maybe not saying it is "incoherent". It is rather that the `self._nodes_filter` is applied for any request, from searching for datasets to training, so changing it during the experiment may result in complex and unwanted behaviour.
 
(*) if we change it we may also consider whether we want to keep `TrainingPlanApproveJob(nodes=self.training_data().node_ids() ....)`  or change it to `TrainingPlanApproveJob(nodes=self.filtered_federation_nodes() ....)`. Same for `TrainingPlanCheckJob`

mvesin: Thank you @sharkovsky for your feedback ! After discussion:
- the discussed comment change on behaviour of `tags` `training_data` and `nodes` is an unwanted revert to old version => remove this change
- regarding the proposed changes: we won't do any change for now. The proposed changes don't fully solve the coherence for `self._nodes_filter`. For example `TrainingPlanApproveJob(nodes=self.training_data().node_ids() ....)` means requesting approval for unused nodes, and `TrainingPlanApproveJob(nodes=self.filtered_federation_nodes() ....)` means we need to resend an approval request if we change the node filter during the experiment (does it work ? not tested yet ... - is it the best solution ? need to decide) => keep track of the discussion, treat as a separate issue in the future.

PR #1088, opened 2024-04-11
Add brain segmentation test

sharkovsky:
Edit @srcansiz : Please do not merge before #1098 
Add another end2ned test based on `MedicalFolderDataset` and brain segmentation task.

### DIFF
--- a/tests/end2end/e2e_ixi_brain_segmentation.py
+++ b/tests/end2end/e2e_ixi_brain_segmentation.py
@@ -1,0 +1,191 @@
+import pandas as pd
+import time
+import tqdm
+import os
+import pytest
+import requests
+import shutil
+from sklearn.model_selection import train_test_split
+from zipfile import ZipFile
+
+
+from helpers import (
+    add_dataset_to_node,
+    start_nodes,
+    kill_subprocesses,
+    clear_researcher_data,
+    clear_experiment_data,
+    get_data_folder,
+    create_researcher,
+    create_multiple_nodes
+)
+
+from experiments.training_plans.ixi_brain_segmentation import UNetTrainingPlan
+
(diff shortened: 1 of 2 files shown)


ybouilla:
Hi Sharkovsky, thanks for this PR: I think it is a good additional test, but needs some improvement (we may want to use `download_and_split_ixi.py` before  doing any testing (instead of re-doing the process in the testing script)

sharkovsky:
Thanks for your comment, 

I chose this implementation because:
- the functions inside `download_and_split_ixi` still needed some tuning (e.g. they put the data inside `notebooks/data` folder, while we want to put it inside `data`)
- the `download_and_split_ixi` cannot be imported as a module, because it is inside the `notebooks` folder which doesn't have an `__init__`. 

Although I agree that it would be nice to avoid duplication, the tradeoff is that it would require a lot of restructuring to make the `download_and_split_ixi` available as a generic module. 
Should I do it anyway?

sharkovsky:
Otherwise I'm also happy to delay this PR until we design a better test suite for the brain segmentation task. It was a good exercise for me anyway during the review too test how easy it was to implement new tests.

ybouilla:
> Otherwise I'm also happy to delay this PR until we design a better test suite for the brain segmentation task. It was a good exercise for me anyway during the review too test how easy it was to implement new tests.

Yes let 's see the input of the others of the team. For other notebooks, we may need to have a facility that run extra scripts

srcansiz:
Hi @sharkovsky,

Thank you adding this test. I think it is a good test to add. 

@ybouilla I agree that duplicating this code is not good but since this test is oriented to e2e to test the functionality rather than downloading a data file and parsing it, I would be okay keep such helper functions separated from notebooks.

Issue #1056, opened 2024-03-21
secagg Joye Libert failing with 4 nodes


mvesin:
Notebook `general-tutorial-secagg.ipynb` fails most often (not every time, yet), from `develop` branch, with 4 nodes (no failure with 3 nodes).
Same issue exists also with `v5.1.0` for example

```
...
-----------------------------------------------------------------

2024-03-21 17:46:28,129 fedbiomed INFO - INFO
					 NODE NODE_8888909b-f6d4-4379-8c8a-557b4655dd7d
					 MESSAGE: Encryption is completed!
-----------------------------------------------------------------

2024-03-21 17:46:28,407 fedbiomed INFO - Nodes that successfully reply in round 0 ['NODE_a40ce146-2ed4-46f3-8ede-07d09799884e', 'NODE_8888909b-f6d4-4379-8c8a-557b4655dd7d', 'NODE_d85762b4-e287-4595-acc5-ce58d20cd799', 'NODE_41bf9abd-ec58-4d4b-a5b0-209d09563daa']

2024-03-21 17:46:28,441 fedbiomed INFO - Validating secure aggregation results...

2024-03-21 17:46:28,445 fedbiomed INFO - Aggregating 4 parameters from 4 nodes.

2024-03-21 17:46:28,445 fedbiomed DEBUG - Aggregation is completed in 0.0 seconds.


--------------------
Fed-BioMed researcher stopped due to exception:
FB417: secure aggregation error: Aggregation is failed due to incorrect decryption.
--------------------
```

mvesin:
Summary: first round of investigation.

1. `develop` branch (commit ob92971)
    - KO (failing with reported error message) most of the time
    - 8 runs of 2 rounds: 8/8 failed (including 7/8 fail at 1st round)
2. failing most of the time also with `v5.1.0` and `v5.0.0`
3. `v4.5.0` branch
    - mix KO and OK
    - 8 runs of 2 rounds: 2/8 failed, 6/8 OK for the 2 rounds
4. `develop` branch plus manually undo the patches from PR #920 in `v5.0.0` for (1) fixing bug for 3+ nodes in secagg (2) introducing weighted fedaverage in secagg
    -  mix KO and OK
    - 8 runs of 2 rounds: 4/8 failed, 4/8 OK for the 2 rounds

5. `develop` branch with 5 then 6 nodes
    - KO failing most of the time


[...]

PR #1089, opened 2024-04-12
Fix secagg failures with 4+ nodes

mvesin:
**PR description**

- address issue #1056 in `divide()` by using `uint64` to avoid overflow when summing several `uint32`. Problem was introduced by weighted average.

- misc robustify for
  * `divide()` by raising an error if data would be lost by converting from `int` to `uint64` (which should not happen with < 2**32 nodes with current weighting function, but will avoid similar bug in the future)
  * `quantize()` by making all type conversions of weight explicit, so it is clear the maximum supported `target_range` is 2**64
  * `reverse_quantize()` by also making explicit the weight type conversions, plus raising an error if encoded weights values are outside of bounds, which would result in information loss (or means we have faulty input)

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/fedbiomed/common/secagg/_jls.py
+++ b/fedbiomed/common/secagg/_jls.py
@@ -124,8 +124,8 @@ def divide(xs: List[int], k: int) -> List[int]:
     # CAVEAT: `dtype` must to allow bigger values than each nodes's weighted value
     # This implementation allows at most 2**32 nodes (as weighted value uses uint32)
     max_val = np.iinfo(np.uint64).max
-    if any([v > max_val for v in xs]):
-        raise FedbiomedSecaggCrypterError(f"{ErrorNumbers.FB624.value}: Cannot divide, values exceed maximum number")
+    if any([v > max_val or v < 0 for v in xs]):
+        raise FedbiomedSecaggCrypterError(f"{ErrorNumbers.FB624.value}: Cannot divide, values outside of bounds")
 
     xs = np.array(xs, dtype=np.uint64)
     return (xs / k).tolist()
@@ -152,7 +152,7 @@ def reverse_quantize(
 
     # CAVEAT: there should not be any weight received that does not fit in `uint64`
     max_val = np.iinfo(np.uint64).max
-    if any([v > max_val for v in weights]):
+    if any([v > max_val or v < 0 for v in weights]):
         raise FedbiomedSecaggCrypterError(
             f"{ErrorNumbers.FB624.value}: Cannot reverse quantize, received values exceed maximum number"
         )
--- a/fedbiomed/common/secagg/_jls.py
+++ b/fedbiomed/common/secagg/_jls.py
@@ -79,15 +79,17 @@ def quantize(
 
     _check_clipping_range(weights, clipping_range)
 
+    # CAVEAT: ensure to be converting from `float` to `uint64`` (no intermediate `int64`)
+    # Process ensures an to compute an `int`` in the range [0, target_range -1]
+    # This enables to use at most 2**64 as target_range (max value of `uint` - 1)
     f = np.vectorize(
         lambda x: min(
             target_range - 1,
-            (sorted((-clipping_range, x, clipping_range))[1] + clipping_range) *
-            target_range /
-            (2 * clipping_range),
-        )
+            (sorted((-clipping_range, x, clipping_range))[1] + clipping_range) * target_range / (2 * clipping_range),
+        ),
+        otypes=[np.uint64]
     )
-    quantized_list = f(weights).astype(int)
+    quantized_list = f(weights)
 
     return quantized_list.tolist()
 
@@ -123,7 +125,6 @@ def divide(xs: List[int], k: int) -> List[int]:
     # This implementation allows at most 2**32 nodes (as weighted value uses uint32)
     max_val = np.iinfo(np.uint64).max
     if any([v > max_val for v in xs]):
-        # TODO RAISE ERROR
         raise FedbiomedSecaggCrypterError(f"{ErrorNumbers.FB624.value}: Cannot divide, values exceed maximum number")
 
     xs = np.array(xs, dtype=np.uint64)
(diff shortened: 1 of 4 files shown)


fedbiomed/common/secagg/_jls.py
```diff
@@ -117,7 +120,14 @@ def divide(xs: List[int], k: int) -> List[int]:
...
     """
-    xs = np.array(xs, dtype=np.uint32)
+
+    # CAVEAT: `dtype` must to allow bigger values than each nodes's weighted value
+    # This implementation allows at most 2**32 nodes (as weighted value uses uint32)
+    max_val = np.iinfo(np.uint64).max
+    if any([v > max_val or v < 0 for v in xs]):
+        raise FedbiomedSecaggCrypterError(f"{ErrorNumbers.FB624.value}: Cannot divide, values outside of bounds")
+
+    xs = np.array(xs, dtype=np.uint64)
```
srcansiz, line 130: Hi @mvesin,

First of all thank you for fixing this bug. 

I understand that there you want to verify if any value is not larger than max value of `int64`. However, following code will raise an error anyway. 

```
np.array([2**64], dtype=np.int64)
```
Therefore, I propose to use exception rather than looping over values which is not performance efficient.

srcansiz: ~~Or we can directly use builtin Python types rather than numpy which will avoid getting overflow errors.~~ Nevermind, it is not a good idea because we can get division errors.

srcansiz (commented):
Hi @mvesin, 

Thank you for fixing this bug! 

I have only one comment to discuss. Once it is resolved I think you can merge it.

fedbiomed/common/secagg/_jls.py
srcansiz: And I didn't understand why any value of `xs` can't be less than 0.

mvesin: Thank you very much @srcansiz for your review !

> And I didn't understand why any value of xs can't be less than 0.
> OK I see, quantized values are always 0 or positive. So this modification makes this function very specific 
> positive integers or quantized values. I would recommend to leave it more generic.

I see your point ...
The idea was to robustify the computation as (1) `xs` elements are typed as `int` (and should be a positive `int`, but this property is only met if the very complex secure aggregation code has no bugs (2) if `xs` element is negative, this will not be detected when converting to `uint64`. This will result in the same type of silent bug that we are fixing.
Maybe then we should move the test <0 to `SecaggCrypter._apply_average` (which is calling `divide` and is specific to this case) ?

> I understand that there you want to verify if any value is not larger than max value of int64. However, following code will raise an error anyway.

The 2 reasons for adding this test were:
1. overhead is little as most of the computation time is in `aggregate()` not in the `divide()`
2. we can give a more explicit message to the user. I hoped to help identifying and fixing future issues in this piece of code :-) If we remove this test we have something like:

```
File /home/mvesin/GIT/fedbiomed/fedbiomed/fedbiomed/common/secagg/_jls.py:130, in divide(xs, k)
...
--> 130 xs = np.array(xs, dtype=np.uint64)
    131 return (xs / k).tolist()

OverflowError: Python int too large to convert to C long
```

... but if you don't agree we can remove this test.

fedbiomed/common/secagg/_jls.py
```diff
@@ -117,7 +120,14 @@ def divide(xs: List[int], k: int) -> List[int]:
     Returns:
         List of divided integers
     """
```
sharkovsky, line 122: @mvesin there is also a `multiply` function which converts to `uint32`. 
Do you think we should also change it to `uint64`?

mvesin: Good question, thank you @sharkovsky !

I wondered whether to robustify also this function, but did not do it because we have "some" guarantees on the way the values are built in `SecaggCrypter.encrypt()`

But at second read I fully agree with you it would be more coherent with the other proposed changes to include it !

Rather than changing the type to `uint64` I would propose to add a check to the received value (that it is in the [0, 2**32-1] range) : in this case we don't need to handle bigger values, we just need to check the received value correspond to what is expected.

fedbiomed/common/secagg/_jls.py
srcansiz: Hi @mvesin, 

Thank you for your answer. I agree that moving these checks to `SecaggCrypter._apply_average` make sense. My point was to keep lower layer functions more generic and deal such invalid types in the upper layers. Because those functions, such as divide, multiply, can be useful for other type of algorithms which may accept negative values. 

For example,  `_apply_average` is a private member of `SecaggCrypter` which calls  `divide` function. There, we can control if is there any negative values or values larger than  max `int64`. This can be done by creating  `np.array, dtype=int64` or looping over the list then it can call `divide` function because it is sure that type and value is correct.

The same argument may be valid for `reverse_quantized`, however, `reverse_quantized` is more specific that it should expect positive values all the time. So it is up to you. 

I am not saying that it is better. It is just to take your opinion about it.
 
> overhead is little as most of the computation time is in aggregate() not in the divide()

Regarding the performance, you are right that it does not take much time compare to aggregation or encryption. My point is just looping over a list and converting list to numpy array takes more or less same amount of time. In this implementation both of them are existing. However, 1M parameters won't make any difference but when you go up to 40M (e.g UNET). It takes around 1.5 seconds, plus 1.5 for converting it to `np.array` 3 seconds in total, same operation for `reverse_quantized` 6 seconds in total. However, I know that there many such loops in secure aggregation implementation so changing this one doesn't make much difference :)

mvesin: > My point was to keep lower layer functions more generic and deal such invalid types in the upper layers. 

OK, so I will move the test for `multiply()` (to be added as suggested by @sharkovsky ) and `divide()` to `SecaggCrypter`.

For `reverse_quantize()` I would keep it as it is, for the reason you mention (specific function for this processing).

> Regarding the performance

I would like to keep these checks, for robustness sake.

We can gain performance by removing the conversion to numpy: as the treatment is simple (divide/multiply) this is more costy to convert twice from `int => numpy => int` than do raw python computation

```
a = list(range(100*10**6))

before = datetime.now() ; b = divide(a,7) ; print(datetime.now() - before)
0:00:06.745853

def divide(xs: List[int], k: int) -> List[int]:
    return ([e / k for e in xs])

before = datetime.now() ; b = divide(a,7) ; print(datetime.now() - before)
0:00:04.025618
```

mvesin: If we plan to re-use the `quantize` / `reverse_quantize` and `multiply` / `divide` it might be nice to re-implement fully in `numpy` (so that we only convert / unconvert once in the process, and then do full numpy computation). But this is beyond the scope of this fix, maybe for (next ?) secagg sprint.

srcansiz (approved):

PR #1103, opened 2024-05-15
adds endurance test

srcansiz:
**PR description**

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/tests/end2end/conftest.py
+++ b/tests/end2end/conftest.py
@@ -22,14 +22,15 @@ from fedbiomed.common.constants import ComponentType, ComponentType
 from fedbiomed.common.utils import CONFIG_DIR
 
 
-pytest.server_port = 50052
-
+_PORT = 50052
 
 @pytest.fixture(scope='module')
 def port():
     """Increases and return port for researcher server"""
-    pytest.server_port += 1
-    return str(pytest.server_port)
+    global _PORT
+
+    _PORT += 1
+    return str(_PORT)
 
 
 @pytest.fixture(scope='module', autouse=True)
--- a/tests/end2end/conftest.py
+++ b/tests/end2end/conftest.py
@@ -21,11 +21,15 @@ from helpers import  (
 from fedbiomed.common.constants import ComponentType, ComponentType
 from fedbiomed.common.utils import CONFIG_DIR
 
-_PORT = 50052
 
-@pytest.fixture(scope='session')
+pytest.server_port = 50052
+
+
+@pytest.fixture(scope='module')
 def port():
-    return str(_PORT + 1)
+    """Increases and return port for researcher server"""
+    pytest.server_port += 1
+    return str(pytest.server_port)
 
 
 @pytest.fixture(scope='module', autouse=True)
--- a/tests/end2end/endurance_e2e_big_neural_network.py
+++ b/tests/end2end/endurance_e2e_big_neural_network.py
@@ -1,0 +1,86 @@
+"""
+Test module to test a big model (60MB) using dry run training
+"""
+import pytest
+
+from helpers import (
+    add_dataset_to_node,
+    start_nodes,
+    kill_subprocesses,
+    clear_experiment_data,
+    clear_researcher_data,
+    get_data_folder,
+    create_researcher,
+    create_multiple_nodes
+)
+
+from experiments.training_plans.mnist_pytorch_training_plan import BigModelMyTrainingPlan
+
+from fedbiomed.common.metrics import MetricTypes
+from fedbiomed.researcher.federated_workflows import Experiment
+from fedbiomed.researcher.aggregators.fedavg import FedAverage
+from fedbiomed.researcher.aggregators.scaffold import Scaffold
+
+
+# Set up nodes and start
+@pytest.fixture(scope="module", autouse=True)
+def setup(port, post_session, request):
+    """Setup fixture for the module"""
+    data_folder = get_data_folder('MNIST-e2e-test')
+    with create_multiple_nodes(port, 3) as nodes:
+        dataset = {
+            "name": "MNIST",
+            "description": "MNIST DATASET",
+            "tags": "#MNIST,#dataset",
+            "data_type": "default",
+            "path": data_folder
+        }
+
+        for node in nodes:
+            add_dataset_to_node(node, dataset)
+
+        researcher = create_researcher(port=port)
+        node_processes, thread = start_nodes(list(nodes))
+
+        yield
+
+        kill_subprocesses(node_processes)
+        thread.join()
+        print("Clearing component data")
+        clear_researcher_data(researcher)
+
+#############################################
+### Start writing tests
+### Nodes will stay up till end of the tests
+#############################################
+
+
+
+def test_01_mnist_pytorch_big_model_training_dry_run():
+    """Tests running training mnist with basic configuration"""
+    model_args = {}
+    tags = ['#MNIST', '#dataset']
+    rounds = 100
+    training_args = {
+        'loader_args': { 'batch_size': 48, },
+        'optimizer_args': {
+            "lr" : 1e-3
+        },
+        'num_updates': 100,
+        'dry_run': True,
+
+    }
+
+    exp = Experiment(
+        tags=tags,
+        model_args=model_args,
+        training_plan_class=BigModelMyTrainingPlan,
+        training_args=training_args,
+        round_limit=rounds,
+        aggregator=FedAverage(),
+        node_selection_strategy=None,)
+
+    exp.run()
+
+    clear_experiment_data(exp)
+
--- a/tests/end2end/endurance_e2e_big_neural_network.py
+++ b/tests/end2end/endurance_e2e_big_neural_network.py
@@ -60,7 +60,7 @@ def test_01_mnist_pytorch_big_model_training_dry_run():
     """Tests running training mnist with basic configuration"""
     model_args = {}
     tags = ['#MNIST', '#dataset']
-    rounds = 100
+    rounds = 30
     training_args = {
         'loader_args': { 'batch_size': 48, },
         'optimizer_args': {
--- a/tests/end2end/endurance_e2e_big_neural_network.py
+++ b/tests/end2end/endurance_e2e_big_neural_network.py
@@ -78,7 +78,8 @@ def test_01_mnist_pytorch_big_model_training_dry_run():
         training_args=training_args,
         round_limit=rounds,
         aggregator=FedAverage(),
-        node_selection_strategy=None,)
+        node_selection_strategy=None,
+        retain_full_history=False)
 
     exp.run()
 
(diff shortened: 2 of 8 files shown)


tests/end2end/helpers/_datasets.py
```diff
@@ -0,0 +1,55 @@
+import os
```
ybouilla, line 1: I would rather name the file by `_datasets_generator.py`

tests/end2end/experiments/training_plans/mnist_pytorch_training_plan.py
```diff
@@ -60,3 +60,60 @@ def training_step(self, data, target):
...
+            x = self.dropout1(x)
+            x = torch.flatten(x, 1)
+            x = self.fc1(x)
+            x = F.relu(x)
+            x = self.dropout2(x)
+            x = self.fc2(x)
+
+            output = F.log_softmax(x, dim=1)
+            return output
+
```
ybouilla, line 107: In my opinion such model is not that big (as big as what we have in tutorials). Only number of neurons in the linear layer has increased. You may add more layers, or use existing models such as `DenseNet`, `Unet`, ..., (some are provided in `pytorch`) as long as the CI slaves can handle it of course.
Also it is usually  good practice to use multiple of `2` as  number of neurons

tests/end2end/conftest.py
```diff
@@ -21,11 +21,15 @@
 from fedbiomed.common.constants import ComponentType, ComponentType
 from fedbiomed.common.utils import CONFIG_DIR
 
-_PORT = 50052
 
-@pytest.fixture(scope='session')
+pytest.server_port = 50052
+
```
ybouilla, line 26: Unless I am missing something, I don't see why you are creating a new attribute for `pytest`.

I don't think it is good practice to do things that way

srcansiz: Hi @ybouilla,

It is to be able to share variable between modules in a test session.

tests/end2end/endurance_e2e_big_neural_network.py
```diff
@@ -0,0 +1,86 @@
...
+        tags=tags,
+        model_args=model_args,
+        training_plan_class=BigModelMyTrainingPlan,
+        training_args=training_args,
+        round_limit=rounds,
+        aggregator=FedAverage(),
+        node_selection_strategy=None,)
+
+    exp.run()
+
```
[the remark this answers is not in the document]
srcansiz: Hi @ybouilla,

Thank you very much,  I thought that I used `retain_full_history=False` it is there in the commit b6c1aa171d5a991614178e97423ccc1e0d8ba9ee but some how it is overwritten. Now, I see you added retain full history as False.

tests/end2end/helpers/_datasets.py
srcansiz: Thank you I will do that.

srcansiz: ... sorry I forgot why I used that name in the first place. It is a generator for now, but it can be the module where we can add all dataset download operations.

tests/end2end/experiments/training_plans/mnist_pytorch_training_plan.py
srcansiz: Hi @ybouilla, 

Thank you for the suggestion but the purpose of this example is not creating a real neural network to do training. The purpose is to test mostly gRPC layer to see how it behaves while exchanging big model files.  If you see in the example it applies dry run. 

However, I think the model size is quite big around 600MB.

tests/end2end/conftest.py
ybouilla: Ok I see. But is it something commonly done in Pytest?

For what you want to do, i was thinking about the following alternative:
```
def port():
    global _PORT
    if _PORT not in globals():
        _PORT = 50052
    else:
        _PORT += 1
    return str(_PORT)
```
For me this solution is *slightly* better for it doesnot touch `pytest` module

srcansiz: Thank you @ybouilla, I will implement as you suggested. What do you think if I rewrite your suggestion as below? 

```
_PORT = 50051
def port():
    global _PORT
    _PORT += 1
    return str(_PORT)
```

tests/end2end/experiments/training_plans/mnist_pytorch_training_plan.py
ybouilla: Yes this part I understood! But I think we should use bigger models holding a higher number of neurons.

For example, with the model `vit_h_14` from `torchvision.model`, you can get a model of size 2GB.

I think we should use these kind of models size for stress tests, provided CI slaves can handle it

tests/end2end/conftest.py
ybouilla: Thanks it is even better!

tests/end2end/experiments/training_plans/mnist_pytorch_training_plan.py
srcansiz: Ok I see your point but could you please tell what do you want to test by providing `vit_h_14` in order to increase number of neurons?

ybouilla: I thought we wanted to test that gRPC can handle exchange of heavy files over several rounds, is that what you were testing?

tests/end2end/endurance_e2e_big_neural_network.py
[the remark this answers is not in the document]
ybouilla: For that i don't really know...

Giving a look to the logs, I can see the following:
```
2024-05-29T07:48:08.1974830Z fedbiomed/common/data/_data_loading_plan.py:101
2024-05-29T07:48:08.1975140Z   invalid escape sequence '\w'
2024-05-29T07:48:08.1975300Z 
2024-05-29T07:48:08.1975410Z fedbiomed/researcher/aggregators/scaffold.py:27
2024-05-29T07:48:08.1975720Z   invalid escape sequence '\('
2024-05-29T07:48:08.1975880Z 
2024-05-29T07:48:08.1975970Z fedbiomed/researcher/filetools.py:176
2024-05-29T07:48:08.1976230Z fedbiomed/researcher/filetools.py:178
2024-05-29T07:48:08.1976510Z   invalid escape sequence '\.'
```
Could it be the reason why tests are failing?

tests/end2end/experiments/training_plans/mnist_pytorch_training_plan.py
srcansiz: Yes exactly, and the model has more than `12800` neurons in the layer FC1 and FC2 so total 25600 neurons. Once it is exported through `torch.save` it is around 500MB, and I think it is enough to test gRPC, because gRPC streams the large message by dividing into 4MB chuncks.

tests/end2end/endurance_e2e_big_neural_network.py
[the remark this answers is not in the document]
srcansiz: What logs are you talking about because tests are passing in the CI now. Please make sure that you don't have any conflict in your files. I think these logs are errors just warning regarding usage of invalid escape string it is there since a long time.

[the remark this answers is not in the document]
ybouilla: nevermind...

tests/end2end/experiments/training_plans/mnist_pytorch_training_plan.py
srcansiz: Hi @ybouilla, 

All endurance tests are passing now. If it is okay for you I am going to squash the commits, push and merge.

ybouilla: > Yes exactly, and the model has more than 12800 neurons in the layer FC1 and FC2 so total 25600 neurons. Once it is exported through torch.save it is around 500MB, and I think it is enough to test gRPC, because gRPC streams the large message by dividing into 4MB chuncks.

Sorry, I am quite confused...

- You want to do a stress test on a high number of Rounds, correct? Ok then for that I think we can just increase the number of `Rounds` to something close to `100` (if i remember correctly, there was a failure on Round 70~80 . And we should rename the test called `e2e_big_neural_network` by something more explicit `e2e_high_number_of_rounds_torch`
- Otherwise, there is not too much difference between the model size of `e2e_mnist_pytorch_training,py` and this model. (they are both around `500MB)`. So you are basically doing the same test several time

ybouilla: Besides `12800` is not a valid number of neurons, for there is no valid reason not to use a power of `2`

ybouilla: > Please see e2e_increased_number_of_nodes.py and read module doc string


You are doing the test only with  scikit-learn, this is why i was wondering if you wanted to do it on pytorch models, hence the confusion. Maybe another test on pytorch may prove useful


> Are you sure? because I tested classical MNIST one is around 5MB


Sorry, you are right, it depends on how you save the model but you can reach 5MB with classical MNIST model. So ok for this test (even if we could try with bigger models)

> e2e_big_neural_network.py -> to test big model exchange through gRPC. This test DOES NOT test convergence or cares about whether number of neurons are valid or not. It tests gRPC large message exchange.

I never talked about caring for convergence. When I meant by number of neurons are valid, it does mean that the model will run in a more optimized way. As a rule of thumb, it is recommended  to use a power of `2` for neurons (for hardware specific and memory reasons). At the very moment you have some training operation done (even if `dry_run=True`), i think it can be good to optimize model training on hardware. But this is merely a recommendation, so do as you want, the ore important part is that tests are passing

ybouilla: > 4.1 Linear quantization
For reasons that will become clear shortly, we used 8-bit quantization to convert activations into
unsigned char, intermediate layer weights into signed char, with the exception of biases
that are encoded as 32-bit int. One exception is the input layer, which remains floating-point, to
better accommodate the potentially larger dynamic ranges of inputs that are not probabilities. In
our particular use-case, the input layer is much smaller (440x2000) than any subsequent layer and
doesn’t weigh strongly on the overall speed. One of the important benefits of quantizing the majority
of the network down to 8 bits is that the total memory footprint of the network consequently shrinks
by between 3× and 4×.

In https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/37631.pdf

srcansiz: I used sklearn because it is smaller than neural networks in terms of model size. We can do more round such as 200. If I use MNIST (5MB) 5 x 200 = 1GB there are 10 nodes 10GB disk space in the slave, where we have 40GB free space so it didn't want to risk it.  The goal of this test is to run more rounds using 10 nodes to test all 10 nodes are orchestrated well through researcher.

srcansiz: If you want to add more endurance test please feel free to add.

srcansiz: > In https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/37631.pdf

Thank you for this useful document and explaining me again that number of neuron should be power of 2. However, I don't think that it will make a difference for dry run.

srcansiz: I pushed a new commit to use valid number of neurons.

Issue #1111, opened 2024-05-28
Create a class to implement key aggrement of diffie-hellman


srcansiz:
The goal of this issue is: 

- [x] to decide which crypto module is better to use `pycryptodome` or `cryptogprahy`
- [x] update conda env `envs/development/conda/fedbiomed-<node|researcher>.yaml`
- [x] implement a class that wraps operations regarding the key agreement 
- [x] implementation assumes that class will take `public_key`, `pravite_key` and other party's `public_key` as an input. 
- [x] as an output it will return the session key (agreed key) of two parties. 
- [x] add unit test to make sure implementation is working.

Please see the skeleton of the class and feel free to modify,

This class can be added in `fedbiomed/common/secagg/_dh.py`

```python

def kdf(x):
    # Implement a Key Derivation Function
    # respecting the security level that we want to have
    pass


class DHKeyAcccord:

    def __init__(
        self,
        private_key: str,
        public_key: str,
        v_public_key: str
    ):
        pass
        # Create attributes
        # generate ephemeral key


    def _import_key(self):
        # Import given key
        # For example:
        #    from Crypto.PublicKey import ECC
        #    ECC.import_key()
        # This is just an example but the idea is to convert static string
        # key to a key instance thay may change depending on the crypto
        # library used
        pass

    def agree(self):

        # There it executes key aggregation t generate pair-wise key
        # session
        pass

        # Return key as in format that LOM supports

```

Please add  pros and cons of using  `pycryptodome` or `cryptogprahy` before deciding which module is going to be used.

srcansiz:
After discussion with @rtaiello and @mvesin, the class `DHKeyAggrement`  will keep the private and public key state for a session. The `.agree(x, y)` method of this class will take the `x: public key of a party` a argument. It will return byte-code agreed key for a single party. 

@rtaiello the `seed` for KDF can be an argument for `DHKeyAggrement` that will be instantiated in `__init__` method and later used in `agree` method.

PR #1126, opened 2024-06-05
Feature/1111 DH KeyExchange

rtaiello:
**DH Key Exchange Class**

Addresses Issue #1111

### Summary of Changes:
- Conducted a comparison between `cryptography` and `pycryptodome` libraries, opting for `pycryptodome` based on the details outlined [here](https://notes.inria.fr/s/tntJVxovZ).
- Updated the environments for both the node and researcher to include `pycryptodome`, and removed `cryptography` because never used in Fed-BioMed.
- Introduced the `DHKeysGeneration` class, which generates Diffie-Hellman (DH) secret and public keys for each node. This class includes methods to export keys in PEM -> bytes format.
- Introduced the `DHKeyAgreement` class, instantiated by each node with its DH secret key. This class provides the `agree` method to perform key agreement with another node's public key. The key derivation function is based on `salt_session` (corresponding to the secagg ID in Fedbiomed) and incorporates node IDs to ensure uniqueness.
- Tests in `tests/test_dh.py`

### DIFF
--- a/envs/development/conda/fedbiomed-node-macosx.yaml
+++ b/envs/development/conda/fedbiomed-node-macosx.yaml
@@ -27,7 +27,7 @@ dependencies:
   - git ~=2.40.1
   - packaging ~=23.1
   # these two have to be aligned
-  - cryptography ~=40.0.0
+  - pycryptodome ~=3.2.0
   - pyopenssl ~=23.1.1
   #
   - joblib >=1.2.0,<1.3.0
--- a/envs/development/conda/fedbiomed-node-macosx.yaml
+++ b/envs/development/conda/fedbiomed-node-macosx.yaml
@@ -27,7 +27,7 @@ dependencies:
   - git ~=2.40.1
   - packaging ~=23.1
   # these two have to be aligned
-  - pycryptodome ~=3.2.0
+  - cryptography ~=40.0.0
   - pyopenssl ~=23.1.1
   #
   - joblib >=1.2.0,<1.3.0
--- a/envs/development/conda/fedbiomed-node.yaml
+++ b/envs/development/conda/fedbiomed-node.yaml
@@ -27,7 +27,7 @@ dependencies:
   - git ~=2.40.1
   - packaging ~=23.1
   # these two have to be aligned
-  - cryptography ~=40.0.0
+  - pycryptodome ~=3.2.0
   - pyopenssl ~=23.1.1
   #
   - joblib >=1.2.0,<1.3.0
--- a/envs/development/conda/fedbiomed-node.yaml
+++ b/envs/development/conda/fedbiomed-node.yaml
@@ -27,7 +27,7 @@ dependencies:
   - git ~=2.40.1
   - packaging ~=23.1
   # these two have to be aligned
-  - pycryptodome ~=3.2.0
+  - cryptography ~=40.0.0
   - pyopenssl ~=23.1.1
   #
   - joblib >=1.2.0,<1.3.0
--- a/fedbiomed/common/secagg/_dh.py
+++ b/fedbiomed/common/secagg/_dh.py
@@ -1,46 +1,105 @@
-from Crypto.PublicKey import ECC
-from Crypto.Hash import TupleHash128
-from Crypto.Protocol.DH import key_agreement
+from cryptography.hazmat.primitives.asymmetric import ec
+from cryptography.hazmat.primitives import serialization, hashes
+from cryptography.hazmat.primitives.kdf.concatkdf import ConcatKDFHash
+from cryptography.hazmat.backends import default_backend
+
+from typing import Callable
+
 from fedbiomed.common.exceptions import FedbiomedSecaggCrypterError
 from fedbiomed.common.constants import ErrorNumbers
 
 
-class DHKeysGeneration:
+class DHKey:
     """
-    Key Generation with ephemeral-ephemeral ECDH, a session is defined for each experiment.
+    Key handling for ephemeral-ephemeral ECDH, a session is defined for each experiment.
 
-    This class handles the generation of ECC keys using the P-256 curve. It provides methods
+    This class handles the generation and importing of ECC keys using the P-256 curve. It provides methods
     to export the private and public keys in PEM format as bytes.
 
     Attributes:
-        private_key (ECC.EccKey): The user's private ECC key.
-        public_key (ECC.EccKey): The user's public ECC key.
+        private_key: The user's private ECC key.
+        public_key: The user's public ECC key.
     """
 
-    def __init__(self):
+    def __init__(
+        self,
+        private_key_pem: bytes | None = None,
+        public_key_pem: bytes | None = None
+    ) -> None:
         """
-        Initializes the DHKeysGeneration instance by generating a private key and deriving the public key.
-        """
-        self._private_key: ECC.EccKey = ECC.generate(curve="P-256")
-        self._public_key: ECC.EccKey = self._private_key.public_key()
+        Initializes the DHKey instance by generating a new key pair or importing the provided keys.
 
-    def export_private_key(self) -> bytes:
+        Args:
+            private_key_pem: Optional. The private key in PEM format.
+            public_key_pem: Optional. The public key in PEM format.
+        """
+        if private_key_pem:
+            self.private_key = self._import_key(
+                serialization.load_pem_private_key,
+                data=private_key_pem,
+                password=None,
+                backend=default_backend()
+            )
+        elif not public_key_pem:
+            self.private_key = ec.generate_private_key(
+                ec.SECP256R1(), default_backend()
+            )
+        else:
+            # Means that only public key is loaded
+            self.private_key = None
+
+        if public_key_pem:
+            self.public_key = self._import_key(serialization.load_pem_public_key,
+                data=public_key_pem, backend=default_backend()
+            )
+        else:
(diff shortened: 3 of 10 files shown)


srcansiz:
Hi @rtaiello, 

Thank you for the PR and also writing the unit tests! great job! In general the implementation looks good to me. However, I have two comments regarding the coding style.  I will have some other suggestions later on, regarding some implementation but I will share those by adding comments to changed files. 

- Could you please check the length of line to be compatible to flake there are lines longer than 100, including comments.

```python
#for example this line
 node_ids = self._node_u_id + node_v_id if self._node_u_id < node_v_id else node_v_id + self._node_u_id
 
 #can be
 node_ids = self._node_u_id +  node_v_id if self._node_u_id  \ 
      < node_v_id else node_v_id + self._node_u_id
``` 

- For doc strings you don't have to specify the types between parenthesis. 
```
Args:
    key: The shared secret key.
    node_v_id: The ID of the other node (node_v).
```

Thank you!

fedbiomed/common/secagg/_dh.py
```diff
@@ -0,0 +1,122 @@
+from Crypto.PublicKey import ECC
+from Crypto.Hash import TupleHash128
+from Crypto.Protocol.DH import key_agreement
+from fedbiomed.common.exceptions import FedbiomedSecaggCrypterError
+from fedbiomed.common.constants import ErrorNumbers
+
+
+class DHKeysGeneration:
```
srcansiz, line 8: Hi @rtaiello, 

This class can be responsible also importing keys. For example, 

- it can take public and private key as input. Then, it imports the keys using method `_import_key_pem`. 
- if there is no public or private key provided as arguments (`None`) it generates a new one. 

Basically this class will be responsible for wrapping keys that are going to be used for DH.  If you apply this suggestion you I think you can rename this class as `DHKey`.

fedbiomed/common/secagg/_dh.py
```diff
@@ -0,0 +1,122 @@
...
+        Performs the key agreement and derives the shared secret.
+
+        Args:
+            node_v_id (str): The ID of the other node (node_v).
+            public_key_pem (bytes): The public key of the other node in PEM format as bytes.
+
+        Returns:
+            bytes: The derived shared secret.
+        """
+        public_key = self._import_key_pem(public_key_pem, key_type="PUBLIC")
```
srcansiz, line 116: If you apply the change suggested in previous comment, there, you can assume that the type of `public_key_pem` will be `DHKey` and you can directly use it.

srcansiz (commented):
Hi @rtaiello,

Thanks again for your implementation. There is only one suggestion that I made regarding the implementation. Please let me know if need more information about that, or you think that it is not relevant for now.

envs/development/conda/fedbiomed-node.yaml
```diff
@@ -27,7 +27,7 @@ dependencies:
   - git ~=2.40.1
   - packaging ~=23.1
   # these two have to be aligned
-  - cryptography ~=40.0.0
```
mvesin, line 30: Thank you @rtaiello !

We should probably also update the files `envs/vpn/conda/fedbiomed-node.yaml` and  `envs/vpn/conda/fedbiomed-researcher.yaml` to keep coherent configuration for the docker/VPN packaging, shouldn't we ?

Thanks !

envs/development/conda/fedbiomed-node-macosx.yaml
```diff
@@ -27,7 +27,7 @@ dependencies:
   - git ~=2.40.1
   - packaging ~=23.1
   # these two have to be aligned
-  - cryptography ~=40.0.0
```
mvesin, line 30: Hi @rtaiello @srcansiz 

Regarding the removal of `cryptography`:

Did we assess the compatibility of `pycryptodome` with `pyopenssl` ?

Touching to conda environments, especially when removing or changing version of packages requires much attention to avoid regression bugs :-/

 The version of `pyopenssl` we currently used 23.1.1 mentions compatibility
 with `cryptography <= 40.0.x`. What happens with `pycryptodome` ?

 The only use I found of `pyopenssl` is in `certificate_manager.py` am I missing something ?

We should probably at least:

1. test that we can re-generate the conda environments (remove them completely, then re-create from scratch) 
2. test that certificate manager works in this new setting


BTW I would avoid changing the version of `pyopenssl` (not done in the current PR, but if you would consider it ...). In this case we would need to redo quite intensive testing after removing/re-create from scratch all the conda envs in the CI. Also we would need pretty extensive manual testing with containers (no CI for that part yet).

fedbiomed/common/secagg/_dh.py
rtaiello: Hey @rtaiello I just pushed a new commit, following you suggestions, many thx.

fedbiomed/common/secagg/_dh.py
rtaiello: Since now is implemented via `cryptography` I removed this function

envs/development/conda/fedbiomed-node.yaml
rtaiello: Fixed it!

envs/development/conda/fedbiomed-node-macosx.yaml
rtaiello: Hey @mvesin , as we discussed I just implemented the DH key and KA, with only `cryptography` so no new dependencies, we keep the same as before. many thx!

mvesin: Thank you @rtaiello !
Is there any specific reason why you change the versions of `cryptography` and `pyopenssl` to older versions than the one that are currently used ?

fedbiomed/common/secagg/_dh.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        self._node_u_id = node_u_id
+        try:
+            self._private_key = serialization.load_pem_private_key(
+                node_u_private_key, password=None, backend=default_backend()
+            )
+        except ValueError:
+            raise FedbiomedSecaggCrypterError(
+                f"{ErrorNumbers.FB628.value}: Invalid private key format"
+            )
```
srcansiz, line 111: Hi @rtaiello,

Thank you for updating `DHKey` class. I think you can use DHKey here with respect to your changes in DHKey. Because DHKey already implement key import.  

```suggestion
        self._private_key = DHKey(private_key_pem=node_u_private_key)
```

fedbiomed/common/secagg/_dh.py
```diff
@@ -0,0 +1,159 @@
...
+            The derived shared secret.
+        """
+        try:
+            public_key = serialization.load_pem_public_key(
+                public_key_pem, backend=default_backend()
+            )
+        except ValueError:
+            raise FedbiomedSecaggCrypterError(
+                f"{ErrorNumbers.FB629.value}: Invalid public key format"
+            )
```
srcansiz, line 156: Same comment here

```suggestion
       public_key = DHKey(public_key=public_key_pem)
```

fedbiomed/common/secagg/_dh.py
```diff
@@ -0,0 +1,159 @@
...
+
+        if public_key_pem:
+            try:
+                self._public_key = serialization.load_pem_public_key(
+                    public_key_pem, backend=default_backend()
+                )
+            except ValueError:
+                raise FedbiomedSecaggCrypterError(
+                    f"{ErrorNumbers.FB629.value}: Invalid public key format"
+                )
```
srcansiz, line 53: Public key and private key attributes can be public  so `DHKeyAggreement can access`

 ```suggestion
        if private_key_pem:
            try:
                self.private_key = serialization.load_pem_private_key(
                    private_key_pem, password=None, backend=default_backend()
                )
            except ValueError:
                raise FedbiomedSecaggCrypterError(
                    f"{ErrorNumbers.FB628.value}: Invalid private key format"
                )
        else:
            self.private_key = ec.generate_private_key(
                ec.SECP256R1(), default_backend()
            )

        self.public_key = self._private_key.public_key()

        if public_key_pem:
            try:
                self.public_key = serialization.load_pem_public_key(
                    public_key_pem, backend=default_backend()
                )
            except ValueError:
                raise FedbiomedSecaggCrypterError(
                    f"{ErrorNumbers.FB629.value}: Invalid public key format"
                )
```

File: fedbiomed/common/secagg/_dh.py [hunk 8]
```diff
@@ -0,0 +1,159 @@
...
+        """
+        try:
+            public_key = serialization.load_pem_public_key(
+                public_key_pem, backend=default_backend()
+            )
+        except ValueError:
+            raise FedbiomedSecaggCrypterError(
+                f"{ErrorNumbers.FB629.value}: Invalid public key format"
+            )
+        shared_secret = self._private_key.exchange(ec.ECDH(), public_key)
```
srcansiz, line 157:
```suggestion
        shared_secret = self._dh_key.private_key.exchange(ec.ECDH(), dh_v_key.public_key)
```

envs/development/conda/fedbiomed-node-macosx.yaml
rtaiello: hey @mvesin my mistake, I'll update them.

srcansiz:
Hi @rtaiello, 


Thank you for the great work! and  applying the suggestions. I've just squashed some of the commits  as extra to have a clear history.

PR #1134, opened 2024-06-18
Extend skeleton for LOM secure aggregation training

mvesin:
**PR description**

**TO BE MERGED AFTER #1129 #1133**

**Developer Certificate Of Origin (DCO)**

By opening this merge request, you agree to the
[Developer Certificate of Origin (DCO)](https://github.com/fedbiomed/fedbiomed/blob/master/CONTRIBUTING.md#fed-biomed-developer-certificate-of-origin-dco)

This DCO essentially means that:

- you offer the changes under the same license agreement as the project, and
- you have the right to do that,
- you did not steal somebody else’s work.

**License**

Project code files should begin with these comment lines to help trace their origin:
```
# This file is originally part of Fed-BioMed
# SPDX-License-Identifier: Apache-2.0
```

Code files can be reused from another project with a compatible non-contaminating license.
They shall retain the original license and copyright mentions.
The `CREDIT.md` file and `credit/` directory shall be completed and updated accordingly.

**Guidelines for PR review**

General:

* give a glance to [DoD](https://fedbiomed.org/latest/developer/definition-of-done/)
* check [coding rules and coding style](https://fedbiomed.org/latest/developer/usage_and_tools/#coding-style)
* check docstrings (eg run `tests/docstrings/check_docstrings`)

Specific to some cases:

* update all conda envs consistently (`development` and `vpn`, Linux and MacOS)
* if modified researcher (eg new attributes in classes) check if breakpoint needs update (`breakpoint`/`load_breakpoint` in `Experiment()`, `save_state_breakpoint`/`load_state_breakpoint` in aggregators, strategies, secagg, etc.)
* if modified a component with versioning (config files, breakpoint, messaging protocol) then update the version following the rules in `common/utils/_versions.py`

Issue #1112, opened 2024-05-28
Add new class inherits from `BaseSecaggSetup`  for LOM implementation for node


srcansiz:
This issue addresses creating a new `SecaggSetup` class on the node side to start secure aggregation context setup for DH key exchange. 
- Takes secure aggregation setup request
- Validates the request (number of parties etc)
- Dipatches DH key exchange protocol using DHController
- Returns the reply back to researcher.
 
Please see following UML diagram. 

![secagg_setup_uml](https://github.com/fedbiomed/fedbiomed/assets/28705816/9cd7f76f-94ba-4c6c-85be-39eb451790f2)

To complete de issue: 

- [x] `SecaggDHSetup` class implementation
- ~~[ ] Unit tests~~ => moved to #942 (cannot write unit tests before inserting real payload)

PR #1135, opened 2024-06-19
Refactor node side `SecaggBaseSetup` hierarchy and add `SecaggDhSetup`

mvesin:
**PR description**

**TO BE MERGED AFTER #1129 #1133 #1134**

- address #1112 
- fix/enhance researcher side `SecaggDhContext._secagg_round_specific()` to properly
  handle cases with error (problem introduced in first version of this code in  #1133)
- fix node side `SecaggBaseSetup.setup()` to properly handle renegotiation
      of a cryptographic contest with the same `secagg_id` as a previous partially failed
      context (issue already existing with JoyeLibert secagg)
- fix/enhance node side for checking parties at the begining of each training round
  in `Round._check_configure_secagg()`. Implies transmitting round's parties list in
  `secagg_arguments` of `TrainRequest`

Reviewer can test running `101_TEMPORARY.ipynb` notebook with:
- 2 nodes, 3 nodes
- LOM secagg, JL secagg, no secagg
- sequence of: 1 round with a type of secagg, next round with another secagg
- error cases in DH context establishment (change the maximum delay to 12-15 in protocol_handler.py )

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:


mvesin:
Comment from @srcansiz initially in #1129 and moved to this PR:

I think this part needs more simple implementation to facilitate the usage. As developer I would expect more like this usage
 
```
try:
  replies = send_node(list_of_nodes, message, wait=True, timeout=timeout)
except FedbiomedError: 
  return ....
  
return ...success...
``` 

 I thought about how it can be implemented, and it seems that only way is to take `send_node` into `GrpcClient`. This can be done by adding a new method or aggregating a new class into `GrpcController` to manage node-to-node send messages as you did in `send_overlay_message`.
 
 If it is not possible to make this change or it makes implementation complicated I would suggest to take `wait` operation into `send_overlay_message`, and rename `send_overlay_message` as `send_nodes`.  

 ---

Sorry maybe I was not explicit in the example usage that shared. (1) My point was `send_overlay_message` could be implemented in a way that we don't need to pass `grpc_client` and `pending_request` each time we need to send a request. (2) As developer the naming of the method is not explicit (`send_overlay_message`), I prefer a naming that explicitly indicates that this methods sends messages to nodes. (3) calling `wait` is not difficult but it can be simpler if it is called automatically from  the `send_overlay_message`. (4) as a developer, I do not prefer catch unexpected exceptions each time I call `send_overlay_message`, instead I prefer `send_overlay_message` handles unexpected exceptions and I can catch only `FedbiomedError` to return a status.

fedbiomed/common/utils/_secagg_utils.py
```diff
@@ -1,6 +1,38 @@
...
+    # but not with [ 'un', 'deux', 'quatre' ]
+    return (
+        # Commented tests can be assumed from calling functions
+        #
+        # isinstance(context, dict) and
+        # 'parties' in context and
+        # isinstance(context['parties'], list) and
+        # len(context['parties']) >= 1 and
+        # isinstance(parties, list) and
+        parties[0] == context['parties'][0] and
```
srcansiz, line 32: Hi @mvesin,

I didn't understand why the first party need to be same?

fedbiomed/node/secagg.py
```diff
@@ -316,48 +309,128 @@ def __init__(
...
+        all_received, messages = self._pending_requests.wait(listener_id, TIMEOUT_NODE_TO_NODE_REQUEST)
+
+        logger.debug(f"Completed Diffie-Hellmann setup with success={all_received} "
+                     f"node_id='{environ['NODE_ID']}' secagg_id='{self._secagg_id}")
+        if not all_received:
+            nodes_no_answer = set(other_nodes) - set([m['node_id'] for m in messages])
+            msg = f"{ErrorNumbers.FB318.value}: Some nodes did not answer during Diffie Hellman secagg " \
+                  f"context setup: {nodes_no_answer}"
+            logger.error(msg)
+            raise FedbiomedSecaggError(msg)
```
srcansiz, line 411: Hi @mvesin,
`logger.error` can be called where this exception is caught in the end. For example, before sending reply in `setup` function, because there are also other exception may be raised and caught in the same place.

srcansiz (commented):
Hi @mvesin,

Thank you for the PR! I have just two comments regarding this PR, after they are resolved it can be merged.

fedbiomed/common/utils/_secagg_utils.py
mvesin: Hi @srcansiz 

> checked for consistency, but not security important

If I understand it correctly, this is not really needed in the sense that there is no cryptographic material on the researcher (`parties[0]`) , so technically a `secagg_id` could be used by another researcher.

But on the other hand, a `secagg_id` is attached to an `experiment_id`, and an experiment is itself attached to a `researcher_id`. So it should not be used with another researcher, when we implement multiple researchers in the future.
So I would say this is good practice to implement that already, though I agree it does not change anything for now, as we use single `researcher_id`

fedbiomed/node/secagg.py
mvesin: Hi @srcansiz
Thank you for the review.

I agree this is not needed. We can completely discard this `logger` entry.

The logging you propose in `SecaggBaseSetup.setup()` already exists (with `debug` instead of `error`), so my `logger` entry in `_setup_specific` is really duplicate.
An `error` entry is also launched later in the `self._create_secagg_reply()` called from `setup()`

```
2024-06-25 08:56:48,349 fedbiomed DEBUG - Completed Diffie-Hellmann setup with success=False node_id='NODE_e2853014-c202-4a62-83b5-2662a4e8495d' secagg_id='secagg_178c1ca5-cc4a-4d06-8fe2-f2ad719f032c
2024-06-25 08:56:48,349 fedbiomed ERROR - FB318: Secure aggregation setup error: Some nodes did not answer during Diffie Hellman secagg context setup: {'NODE_9ff33d22-a893-4649-9535-ebf465a5221f'}
2024-06-25 08:56:48,350 fedbiomed DEBUG - FB318: Secure aggregation setup error: Some nodes did not answer during Diffie Hellman secagg context setup: {'NODE_9ff33d22-a893-4649-9535-ebf465a5221f'}
2024-06-25 08:56:48,350 fedbiomed ERROR - Can not setup secure aggregation context on node for secagg_178c1ca5-cc4a-4d06-8fe2-f2ad719f032c.
```

fedbiomed/common/utils/_secagg_utils.py
srcansiz: Ok I see you point. However, researcher is not part of this secagg context. Therefore, I would not include researcher in the list of parties. In terms of consistency, I think the consistency between experiment and researcher should be done before the code reaches this point. Node should verify if the experiment belongs to that researcher or not before processing anything (maybe just after receiving a request related to a experiment). However, we do not have this mechanism yet so I would leave it to later rather than implementing now.

mvesin: OK, then I can remove the test for `parties[0] == context['parties'][0]`.

Nevertheless, I would like to keep passing *all* parties as arguments, to have common prototype with the other `matching_parties_*()` functions. So we can easier refactor this code later.
Servkey and biprime need all the parties to properly check compatibility of parties with the secagg context element.

srcansiz:
Hi @mvesin,

Thank you. The PR is okay for me. You can merge it when you think it is ready after your last tests.

mvesin:
> (1) My point was send_overlay_message could be implemented in a way that we don't need to pass grpc_client and pending_request each time we need to send a request. (2) As developer the naming of the method is not explicit (send_overlay_message), I prefer a naming that explicitly indicates that this methods sends messages to nodes. (3) calling wait is not difficult but it can be simpler if it is called automatically from the send_overlay_message. (4) as a developer, I do not prefer catch unexpected exceptions each time I call send_overlay_message, instead I prefer send_overlay_message handles unexpected exceptions and I can catch only FedbiomedError to return a status.

- (2) => OK, I do it
- (4) => does not apply anymore in current version (we don't have to call within `try/except`)
- (1) and (3) => as discussed we want to address them, but in a specific branch

---

For (1) an alternative to adding method/class to `GrpcController` would be to transform `send_overlay_message` to a class, that is a layer above `GrpcController` eg

```
class OverlayController:
  def __init__(self, grpc_client: GrpcController, pending_requests: PendingRequests):
    ...
  def def send_overlay_message(
        researcher_id: str,
        nodes: List[str],
        messages: List[InnerMessage]) -> Optional[int]:
```

So we would call:
```
self._overlay_controller.send(researcher, nodes, messages)
```

```

**Comment by mvesin** [comment]
Found a (huge) regression bug introduced late with this branch: does not work when not using secure aggregation. Fix it in same commit as modifications for review comments.

When no secagg node fails training with:
```
2024-06-25 15:54:30,078 fedbiomed ERROR - FB314: Node round error: Bad parties list in train request: None
```

Issue #1114, opened 2024-05-28
Add new manager called SecaggDhManager where LOM context is managed.


srcansiz:
`SecaggManager` classes are the classes that manages secure aggregation context database read and write operation. There are currently `SecaggServkeyManager` and `SecaggBiprimeManager` manager. This issue requires to create another manager called `SecaggDHLOMManager` which is responsible for: 

- preparing the secagg context respects to the format for LOM implementation. This part is highly related to please see issue #1113 
- implements CRUD operations, create, read, update, delete

PS: This is can be implemented with the issue #1113

Design: Please respect and use `BaseSecaggManager`

PR #1138, opened 2024-06-24
 redesign database for secagg contexts

### LINKED ISSUE #1113: Redesign database for secagg to add LOM secagg context
Secure aggregation contexts are save in the database respecting certain formats respecting the type of context: 

```
//Biprime
{
 'secagg_id': xxxx,
 'secagg_version': xxx,
 'parties': <list of participating parties>,
 'type': <enum BiprimeType.DYNAMIC.value, BiprimeType.DEFAULT.value>,
 'context': Dict[str, int]
}

// Servkey 
{
 'secagg_id': xxxx,
 'secagg_version': xxx,
 'parties': <list of participating parties>,
 'context': Dict[str, int],
 'experiment_id': xxxx,
}

```
In Tinydb there are different tables for Servkey and Biprime. 

This issue focuses on how database should be redesign after adding LOM context. LOM context is different that other context since it consist of key-pairs between a node and other nodes. For example, 

This is an example secagg context that is going to be saved in `node-1` database in a FL experiment. There are `node-2` and `node-3` in the experiment:

`<key-pair-node-1-node-2>, key-pair-node-1-node-3`

During the encryption phase, `-` and `+`  signs are assigned to these individual keys  to sum or subtract from an integer depending on the id of the node that is paired. Therefore, it is important to know the identity of the node that the key is paired. Respecting this information please provide: 

- [ ] How `context` property of db entry should be design: Dict, List, Set etc. 
- [ ] a.1- Should we use single table to keep all types of secagg context or keep using different tables for different context types. 
- [ ] a.2 - depending on  a.1 please propose a final database entry design + create an additional issue if needed (*)

(*) if `a.1` decides to merge the existing secagg tables, then an additional issue needs to be created for updating `SecaggServkeyManager` and `SecaggBiprimeManager`: 
- extend with another field `type` (servkey, biprime, lom) to be able to easily associate a secagg entry to its type
- rewrite the methods from `SecaggServkeyManager` and `SecaggBiprimeManager` (eg for "remove" a `SecaggServkeyManager.remove()` should never be able to remove a biprime key, even if we provide the correct secagg_id)

ybouilla:
**PR description**

Closes #1113 and #1114

Adds:
- saving/loading both DH and JL contexts in/from the database
- additional entries in the database
- minor refactoring
- updates on `test_secagg_manager_common.py`

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:


fedbiomed/researcher/secagg/_secagg_context.py
```diff
@@ -306,6 +306,7 @@ def setup(self) -> bool:
             'parties': self._parties,
             'command': 'secagg',
         })
+        print("PARTIES", self._parties, self._create_payload)
```
mvesin, line 309: Probably a debug output that can now be suppressed ?

fedbiomed/common/secagg_manager.py
```diff
@@ -113,30 +122,60 @@ def _add_generic(self, secagg_id: str, parties: List[str], specific: dict):
...
+                      f' an entry already exists for secagg_id={secagg_id}'
+            logger.error(errmess)
+            raise FedbiomedSecaggError(errmess)
+
+    def _raise_error_incompatible_retrieved_entry(self,
+                                                  entry: Union[None, Dict],
+                                                  experiment_id: str,
+                                                  secagg_id: str,
+                                                  component: SecaggElementTypes):
+        if entry is not None and entry['experiment_id'] != experiment_id:
```
mvesin, line 154: Thank you @ybouilla for this PR !

Here shouldn't we also test  that the `entry['secagg_elem']` matches the `component.value` ?

fedbiomed/common/secagg_manager.py
```diff
@@ -113,30 +122,60 @@ def _add_generic(self, secagg_id: str, parties: List[str], specific: dict):
...
-            logger.error(errmess)
-            raise FedbiomedSecaggError(errmess)
-
+        self._check_existing_entry_in_db(secagg_id)
         specific.update({
             'secagg_version': str(__secagg_element_version__),
             'secagg_id': secagg_id,
-            'parties': parties
+            'parties': parties,
+            'secagg_scheme': secagg_scheme.value,
```
mvesin, line 130: Hi @ybouilla 

I don't understand why the `secagg_scheme` is needed ? (moreover we save it but don't really use it)
Isn't the `secagg_elem` enough ?

ybouilla: Actually it was in provision for future developments. I thought it could be useful to treat some context regarding its scheme. But if you think it is meaningless we can remove it.

fedbiomed/common/secagg_manager.py
ybouilla: > Here shouldn't we also test that the entry['secagg_elem'] matches the component.value ?

Yes this sounds like a good idea!

> Plus we should also call _raise_error_incompatible_retrieved_entry() from SecaggBiprimeManager.get() around line 330 ...

For that I am not sure. In the previous version there is no error raised, so I decided to leave it as it is. Is it something we should change?

fedbiomed/common/secagg_manager.py
```diff
@@ -113,30 +122,60 @@ def _add_generic(self, secagg_id: str, parties: List[str], specific: dict):
...
+                                                  secagg_id: str,
+                                                  component: SecaggElementTypes):
+        if entry is not None and entry['experiment_id'] != experiment_id:
+            errmess = f'{ErrorNumbers.FB623.value}: error getting {component.value} element: ' \
+                      f'an entry exists for secagg_id={secagg_id} but does not belong to ' \
+                      f'current experiment experiment_id={experiment_id}'
+            logger.error(errmess)
+            raise FedbiomedSecaggError(errmess)
+
+    def _raise_error_mismatch_secagg_id_and_experiment_id(self,
```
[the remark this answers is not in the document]
ybouilla: I am going to add this test!

ybouilla (commented):
Thanks for your review Marc! Here are some answers and explanation.

I just realized that I havenot pushed latest updates on the github server

fedbiomed/common/secagg_manager.py
mvesin: >>  Plus we should also call _raise_error_incompatible_retrieved_entry() from SecaggBiprimeManager.get() around line 330 ...
> For that I am not sure. In the previous version there is no error raised, so I decided to leave it as it is. Is it something we should change?

Thank you @ybouilla for your answer.

Sorry, I really think this is not coherent.
Of course in the previous version we don't test that the entry['secagg_elem'] matches the component.value in `SecaggBiprimeManager.get()`. We also don't test it in `SecaggServkeyManager.get()`.
This is perfectly fine in both cases as in previous version the database table is **dedicated** to a **single** `secagg_element` type. So we cannot retrieve another type of element.

Now the database table is shared between all types of context element (servkey, biprime, DH) so when we retrieve an element we always need to ensure that the element is of the proper type ...

If it is not clear to you maybe we can discuss it this afternoon ?

fedbiomed/common/secagg_manager.py
mvesin: I think it is not needed.

But it's not harmful either ...

fedbiomed/common/secagg_manager.py
ybouilla: Oh ok Sorry I misunderstood: you are speaking about the check against `secagg_element`, correct?
I thought you were speaking about `experiment_id` (which is not checked in `SecaggBiprimeManager`). Actually ` _raise_error_incompatible_retrieved_entry` is first and foremost checking if `experiment_id`s match, but is unused in `SecaggBiprimeManager`. So I would create another method that checks correctness of `secagg_element`. 
Let me know if there is something I am missing or misunderstanding

mvesin: > you are speaking about the check against secagg_element, correct?

Yes, exactly.

> Actually  _raise_error_incompatible_retrieved_entry is first and foremost checking if experiment_ids match, but is unused in SecaggBiprimeManager. So I would create another method that checks correctness of secagg_element.

Yes, I agree with that.
We still don't want to check `experiment_id` for `SecaggBiprimeManager`

fedbiomed/common/secagg_manager.py
mvesin: ... but if we keep it, then we should check it also every time we check the `secagg_elem` ... so it's probablly better to remove them to avoid useless code checks.

fedbiomed/common/secagg_manager.py
```diff
@@ -113,30 +122,82 @@ def _add_generic(self, secagg_id: str, parties: List[str], specific: dict):
...
+            FedbiomedSecaggError: error raised if above condition(s) is/are matched.
+        """
+        errmess: str = None
+        if entry is not None:
+            if experiment_id is not None and entry['experiment_id'] != experiment_id:
+                errmess = f'{ErrorNumbers.FB623.value}: error {database_operation_name} {component.name} element: ' \
+                          f'an entry exists for secagg_id={secagg_id} but does not belong to ' \
+                          f'current experiment experiment_id={experiment_id}'
+
+            if entry is not None and entry['secagg_elem'] != component.value:
```
mvesin, line 187: we already know that `entry is not None` => remove redundant test

fedbiomed/common/secagg_manager.py
mvesin: > Plus we should also call _raise_error_incompatible_retrieved_entry() from SecaggBiprimeManager.get() around line 330 ...

Adding a test in `SecaggBiprimeManager.get()`

mvesin:
Hi @ybouillar 

While testing the branch, I discovered an issue introduced because with now use same table for all secagg component types, updating the table from distinct `Table` objects (each `SecaggManager` instantiates a `Table` object). 

Failure scenario:
1. use secagg LOM
2. then use secagg Joye Libert with same experiment
3. re-use secagg LOM
4. etc.

At each secagg type change we get errors:

```
2024-07-01 13:02:02,171 fedbiomed DEBUG - FB623: Secure aggregation database error: failed adding an entry in table "<DBTable name='SecaggManager', total=6, storage=<tinydb.storages.JSONStorage object at 0x7f6efb91a470>>" for secagg element=SERVER_KEY secagg_id=secagg_d849f3ec-2626-49a8-8e6d-25be8ee8688c with error: Document with ID 5 already exists
2024-07-01 13:02:02,171 fedbiomed ERROR - Can not setup secure aggregation context on node for secagg_d849f3ec-2626-49a8-8e6d-25be8ee8688c.
```

Possible solution (pushed it to the branch): add a singleton around `Table` to ensure all `SecaggManager` use the same object

mvesin (approved):

Issue #1013, opened 2024-02-29
Bug: `Metrics` class fails to recognize the number of samples the training dataset contains when it is of size 1


ybouilla:
**Bug description:**
When using the testing facility, `fedbiomed.common.metrics.Metrics` fails to recognize the number of samples of `y_pred` and `y_true` have, in the method `_configure_y_true_pred`. This raise the error  `Predictions and true values should have equal number of samples`, because the check `len(y_pred) == len(y_true)` fails, when using MNIST dataset. Indeed, `y_pred` is of size `(1, 10)`  - probability of each of the 10 digits - whereas `y_true` of size `(1,1)` so the check gives `len(y_pred)=10` and `len(y_true)=1`. Check fails albeit `y_true` and `y_pred` have correct sizes.

**Suggestions**
we can pass the number of samples as argument, or handle specifically `y_pred` and `y_true` when the testing dataset is equal to `1` or improve the batch_size inference

srcansiz:
After checking again implementation of `_configure_y_true_pred`, I've realized that  squeezing `y_true` causes the issue. Then, I checked to see if squeezing is necessary, and finally it is not really needed. This operation is already done by  metric function provided by `scikit-learn`. For example,  `y_true = [[1]]` and `y_pred = [2]` works fine while calculating RMSE using scikit-learn. 

I tested, MNIST pytorch, MNIST sklearn, classification and regression examples with sklearn. All tests are done using `batch_size=1` and also `batch_size = >1`. Finally, it works fine after removing squeeze operation.

PR #1155, opened 2024-07-24
New feature: testing batch size

### LINKED ISSUE #893: Node out of memory because of testing batch size
Our current implementation for both [torch](https://github.com/fedbiomed/fedbiomed/blob/6c568b892e8222e04d15c23e39a331ab028cc2f9/fedbiomed/common/data/_torch_data_manager.py#L128) and [sklearn](https://github.com/fedbiomed/fedbiomed/blob/6c568b892e8222e04d15c23e39a331ab028cc2f9/fedbiomed/common/data/_sklearn_data_manager.py#L356) sets the batch size of the testing loader to be equal to the full testing dataset. 

This leads to an out-of-memory problem when the data types are too large (e.g. in the FeTS dataset, where each image has shape [128,128,128] and there are 4 modalities) and/or when the testing dataset is too large. 

Two possible solutions would be:
1) always set testing batch size equal to 1
2) allow user to set testing batch size through `loader_arguments`

This issue is related to #765, but is different from it because the problem described here (i.e. the issue with batch size) would still be relevant even if we fixed #765.

ybouilla:
**PR description**
This PR adds the testing_batch_size feature to the testing facility, in order to fix the issue depicted https://github.com/fedbiomed/fedbiomed/issues/893 , ie partially fixing issues where testing dataset triggers a Node `MemoryError`, by computing testing metrics in batches instead of the whole dataset.
This feature is disabled by setting `test_batch_size=None` or `test_batch_size=0` entry, to be set in the `TrainingArgs`.
Closes #893.

**Warning**
There is an issue (not due to this new feature) that has been spotted while developing: if `test_batch_size=1`. then an error is triggered complaining that prediction and targets don't have the same size (see issue https://github.com/fedbiomed/fedbiomed/issues/1013 ). This issue is still not solved with this new feature.

**Warning 2**

My initial working branch contained commits that should not have belong in this branch (merged branch `feature/lom-secure-agregation` by accident). Since I was afraid it could conflict when merging into `develop`, I decided to remove commits related to branch `feature/lom-secure-aggregation`. The other branch can be found at `feature/893-testing-batch-size-2`

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/docs/user-guide/researcher/model-testing-during-federated-training.md
+++ b/docs/user-guide/researcher/model-testing-during-federated-training.md
@@ -2,7 +2,7 @@
 title: Model Validation on the Nodes in Federated Training
 description: The validation part in FL plays an important role while evaluating model performance that is trained on different nodes with
              different datasets. Fed-BioMed provides a validation routine on the datasets that are randomly split at each round of training.
-keywords: validation in FL, Fed-BioMed model validation, validating model performance
+keywords: validation in FL, Fed-BioMed model validation, validating model performance, test_batch_size
 ---
 
 # Model Validation During Federated Training on the Nodes
@@ -74,6 +74,8 @@ Here is the list of validation arguments that can be configured.
     defined in the training plan (see section: Define Custom Validation Step) default metric will be
     `ACCURACY`.
  - `test_metric_args`: A dictionary that contains the arguments that will be used for the metric function.
+ - `test_batch_size`: A value used to compute metrics using batches instead of
+ loading the full Testing dataset (specified by `test_ratio`). This option should be considered if data are very heavy, and can lead to [`MemoryError`](https://docs.python.org/3/library/exceptions.html#MemoryError) errors. By specifying it, it performs several computations over a batch of data. By using this option, you should select wisely the batch size and the metric, for some metrics can be meaningless if computed over several batches of data. `test_batch_size` should be greater or equal than `2` (enabled) or equal to `0` or `None` (disabled).
 
 !!! info
     Validation functions for each default metric executes functions from scikit-learn framework.
@@ -83,7 +85,7 @@ Here is the list of validation arguments that can be configured.
 
 
 To activate validation on the node side, the arguments `test_ratio` and at least one of `test_on_local_updates` or
-`test_on_global_updates` should be set. Since the default values of `test_on_local_updates` and `test_on_global_updates`
+`test_on_global_updates` should be set to `True`. Since the default values of `test_on_local_updates` and `test_on_global_updates`
 are `False`, setting `test_ratio` will only split dataset as validation and train sets but won't perform validation.
 
 
@@ -107,6 +109,7 @@ training_args = {
     'test_metric': MetricTypes.F1_SCORE,
     'test_on_global_updates': True,
     'test_on_local_updates': True,
+    'test_batch_size': 0,
     'test_metric_args': {'average': 'macro'}
 }
 
@@ -130,6 +133,7 @@ training_args = {
     },
     'epochs': 2,
     'batch_maxnum': 100,
+    'test_batch_size': 0
 }
 exp = Experiment(training_args=training_args)
 
@@ -149,8 +153,25 @@ exp.set_set_test_metric(MetricTypes.ACCURACY)
 exp.run(rounds=2, increase=True)
 ```
 
+## Using Validation facility with heavy datasets
+
+In some specific cases, you may have some very huge datain your datasets, such as 3D images, or you want a huge validation set, that you can load on `Node`s only through batches, in order to avoid `Nodes` to crash due to lack of Memory. `Fed-BioMed` provides an option to compute validation metrics through batches, in the `TrainingArgument`: `test_batch_size`.
+
+By setting `test_batch_size`, selected metrics will be computed over batches instead of computing metric using the whole dataset (which is the default behaviour). For now, `Fed-BioMed` doesnot provide a way to reconcile all these metrics under one metric (such as computing the mean of Accuracy for instance). If you want to do so, please consider defining your own `Validation step` by write your own `testing_step` method in the `TrainingPlan` (see below for further datails).
 
 
+```python
+training_args = {
+    'optimizer_args': {
+        'lr': 1e-3,
+    },
+    'epochs': 2,
+    'test_ratio': .4,
+    'test_batch_size': 64  # validation metrics will be done by considering batches of size 64 each
+}
+exp = Experiment(training_args=training_args)
+
+```
 ## Define Custom Validation Step
 
 Fed-BioMed training plans allow defining custom validation steps for model evaluation on the node side. The name of the
--- a/docs/user-guide/researcher/model-testing-during-federated-training.md
+++ b/docs/user-guide/researcher/model-testing-during-federated-training.md
@@ -75,7 +75,7 @@ Here is the list of validation arguments that can be configured.
     `ACCURACY`.
  - `test_metric_args`: A dictionary that contains the arguments that will be used for the metric function.
  - `test_batch_size`: A value used to compute metrics using batches instead of
- loading the full Testing dataset (specified by `test_ratio`). This option should be considered if data are very heavy, and can lead to [`MemoryError`](https://docs.python.org/3/library/exceptions.html#MemoryError) errors. By specifying it, it performs several computations over a batch of data. By using this option, you should select wisely the batch size and the metric, for some metrics can be meaningless if computed over several batches of data. `test_batch_size` should be greater or equal than `2` (enabled) or equal to `0` or `None` (disabled).
+ loading the full testing dataset (specified by `test_ratio`). Setting `test_batch_size` can avoid having [`MemoryError`](https://docs.python.org/3/library/exceptions.html#MemoryError) errors due to large and /or heavy datasets. You should select wisely the batch size and the metric, for some metrics can be meaningless if computed over several batches of data. `test_batch_size` should be greater or equal than `2` (enabled) or equal to `0` or `None` (disabled).
 
 !!! info
     Validation functions for each default metric executes functions from scikit-learn framework.
@@ -157,8 +157,7 @@ exp.run(rounds=2, increase=True)
 
 In some specific cases, you may have some very huge datain your datasets, such as 3D images, or you want a huge validation set, that you can load on `Node`s only through batches, in order to avoid `Nodes` to crash due to lack of Memory. `Fed-BioMed` provides an option to compute validation metrics through batches, in the `TrainingArgument`: `test_batch_size`.
 
-By setting `test_batch_size`, selected metrics will be computed over batches instead of computing metric using the whole dataset (which is the default behaviour). For now, `Fed-BioMed` doesnot provide a way to reconcile all these metrics under one metric (such as computing the mean of Accuracy for instance). If you want to do so, please consider defining your own `Validation step` by write your own `testing_step` method in the `TrainingPlan` (see below for further datails).
-
+Validation will be computed over batches instead of computing metric using the whole dataset (which is the default behaviour) if `test_batch_size` is set. For now, `Fed-BioMed` does not provide a way to reconcile all these metrics under one metric (such as computing the mean of Accuracy for instance). If you want to do so, please consider defining your own `Validation step` by writing your own `testing_step` method in the `TrainingPlan` (see below for further details).
 
 ```python
 training_args = {
--- a/docs/user-guide/researcher/tensorboard.md
+++ b/docs/user-guide/researcher/tensorboard.md
@@ -33,6 +33,9 @@ exp = Experiment(tags=tags,
                 )
 ```
 
+!!! info "Validation facility"
+    Tensorboard displays results got from validation metric, in the testing facility. For each `Round`, values will be exported into `Tensorboard`. If `test_batch_size` is set to a specific value, each computed value will be repported (depending on the size of `test_batch_size` set).
+
 ## Launching Tensorboard
 ### 1. From Terminal
 
--- a/docs/user-guide/researcher/tensorboard.md
+++ b/docs/user-guide/researcher/tensorboard.md
@@ -34,7 +34,7 @@ exp = Experiment(tags=tags,
 ```
 
 !!! info "Validation facility"
-    Tensorboard displays results got from validation metric, in the testing facility. For each `Round`, values will be exported into `Tensorboard`. If `test_batch_size` is set to a specific value, each computed value will be repported (depending on the size of `test_batch_size` set).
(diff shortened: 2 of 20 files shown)


docs/user-guide/researcher/model-testing-during-federated-training.md
```diff
@@ -74,6 +74,8 @@ Here is the list of validation arguments that can be configured.
     defined in the training plan (see section: Define Custom Validation Step) default metric will be
     `ACCURACY`.
  - `test_metric_args`: A dictionary that contains the arguments that will be used for the metric function.
+ - `test_batch_size`: A value used to compute metrics using batches instead of
+ loading the full Testing dataset (specified by `test_ratio`). This option should be considered if data are very heavy, and can lead to [`MemoryError`](https://docs.python.org/3/library/exceptions.html#MemoryError) errors. By specifying it, it performs several computations over a batch of data. By using this option, you should select wisely the batch size and the metric, for some metrics can be meaningless if computed over several batches of data. `test_batch_size` should be greater or equal than `2` (enabled) or equal to `0` or `None` (disabled).
```
srcansiz, line 78:
```suggestion
 loading the full testing dataset (specified by `test_ratio`). Setting batch size can avoid having [`MemoryError`](https://docs.python.org/3/library/exceptions.html#MemoryError) errors due to large and heavy datasets.  You should select wisely the batch size and the metric, for some metrics can be meaningless if computed over several batches of data. `test_batch_size` should be greater or equal than `2` (enabled) or equal to `0` or `None` (disabled).
```

docs/user-guide/researcher/model-testing-during-federated-training.md
```diff
@@ -149,8 +153,25 @@ exp.set_set_test_metric(MetricTypes.ACCURACY)
 exp.run(rounds=2, increase=True)
 ```
 
+## Using Validation facility with heavy datasets
 
+In some specific cases, you may have some very huge datain your datasets, such as 3D images, or you want a huge validation set, that you can load on `Node`s only through batches, in order to avoid `Nodes` to crash due to lack of Memory. `Fed-BioMed` provides an option to compute validation metrics through batches, in the `TrainingArgument`: `test_batch_size`.
```
srcansiz, line 158:
```suggestion
Depending on data type, using whole dataset can use too much space in memory (e.g 3D images) or some datasets can have high number of observations that consumes too much memory. If whole dataset is used during the testing nodes can crash due to lack of memory or computation takes longer. To avoid such issues `Fed-BioMed` provides an option to compute validation metrics through batches, in the `TrainingArgument`: `test_batch_size`.
```

docs/user-guide/researcher/model-testing-during-federated-training.md
```diff
@@ -149,8 +153,25 @@ exp.set_set_test_metric(MetricTypes.ACCURACY)
...
 
+By setting `test_batch_size`, selected metrics will be computed over batches instead of computing metric using the whole dataset (which is the default behaviour). For now, `Fed-BioMed` doesnot provide a way to reconcile all these metrics under one metric (such as computing the mean of Accuracy for instance). If you want to do so, please consider defining your own `Validation step` by write your own `testing_step` method in the `TrainingPlan` (see below for further datails).
```
srcansiz, line 160:
```suggestion
Validation will be computed over batches instead of computing metric using the whole dataset (which is the default behaviour) if `test_batch_size` is set. For now, `Fed-BioMed` does not provide a way to reconcile all these metrics under one metric (such as computing the mean of Accuracy for instance). If you want to do so, please consider defining your own `Validation step` by writing your own `testing_step` method in the `TrainingPlan` (see below for further details).
```

docs/user-guide/researcher/tensorboard.md
```diff
@@ -33,6 +33,9 @@ exp = Experiment(tags=tags,
                 )
 ```
 
+!!! info "Validation facility"
+    Tensorboard displays results got from validation metric, in the testing facility. For each `Round`, values will be exported into `Tensorboard`. If `test_batch_size` is set to a specific value, each computed value will be repported (depending on the size of `test_batch_size` set).
```
srcansiz, line 37:
```suggestion
    Tensorboard displays the results of the validation/testing. During each `Round`, values will be exported into `Tensorboard` in real time. If `test_batch_size` is set to a specific value, each computed value will be reported (depending on the size of `test_batch_size` set).
```

fedbiomed/common/data/_data_manager.py
```diff
@@ -59,7 +60,7 @@ def extend_loader_args(self, extension: Dict):
         """
         self._loader_arguments.update(
             {key: value for key, value in extension.items() if key not in self._loader_arguments}
-        )
+        )
```
srcansiz, line 63:
```suggestion
        ) 
```

fedbiomed/common/training_plans/_base_training_plan.py
```diff
@@ -549,13 +554,17 @@ def evaluate(data, target):
...
                 if isinstance(target, torch.Tensor):
                     target = target.numpy()
+
                 return metric_controller.evaluate(
                     target, output, metric=metric, **metric_args
                 )
             metric_name = metric.name
         # Iterate over the validation dataset and run the defined routine.
         num_samples_observed_till_now: int = 0
+
```
srcansiz, line 564:
```suggestion
```

fedbiomed/common/training_plans/_base_training_plan.py
```diff
@@ -568,25 +577,30 @@ def evaluate(data, target):
...
-                    test=True,
-                    test_on_local_updates=(not before_train),
-                    test_on_global_updates=before_train,
-                    total_samples=n_samples,
-                    batch_samples=num_samples_observed_till_now,
-                    num_batches=n_batches
+            # Reporting
+
+            if idx % self.training_args()['log_interval'] == 0 or idx == 1 or idx == n_batches:
+                # FIXME: should we use MiniBatchTrainingIterationAccountant here ?
```
srcansiz, line 583: Please create an issue as candidate to inspect this proposal in detail.

fedbiomed/node/round.py
```diff
@@ -770,4 +770,5 @@ def _split_train_and_test_data(self, test_ratio: float = 0):
         # self.training_data will be equal to None
 
         # Split dataset as train and test
-        return data_manager.split(test_ratio=test_ratio)
+        test_batch_size = self.testing_arguments.get('test_batch_size')
+        return data_manager.split(test_ratio=test_ratio, test_batch_size=test_batch_size)
```
srcansiz, line 774:
```suggestion
        return data_manager.split(
            test_ratio=test_ratio,
            test_batch_size=self.testing_arguments.get('test_batch_size'))
```

srcansiz (commented):
Hi @ybouilla,
Thank you for the PR! Please find my comments and some suggestions. I haven't tested yet the test batch size functionality. I am going to test and add some other suggestion if it is necessary.

fedbiomed/common/metrics.py
```diff
@@ -369,7 +369,7 @@ def _configure_y_true_pred_(y_true: Union[np.ndarray, list],
         if y_true.ndim == 0:
             y_true = y_true.reshape((1,))
 
-        if len(y_pred) != len(y_true):
+        if y_pred.shape[0] != y_true.shape[0]:
```
srcansiz, line 372: Hi @ybouilla, 

I don't but what about removing `y_pred = np.squeeze(y_pred)` and `y_true = np.squeeze(y_true)`  to avoid getting error if batch size is equal to 1. Do you think it will create some regression bugs? Because I only tested using MNIST and it seems that it is working.

srcansiz: I tested a regression model using scikit-learn model,  and it failed due to non-squeezed predictions list.

fedbiomed/common/data/_sklearn_data_manager.py
```diff
@@ -349,7 +352,12 @@ def split(self, test_ratio: float) -> Tuple[NPDataLoader, NPDataLoader]:
             self._subset_test = (x_test, y_test)
             self._subset_train = (x_train, y_train)
 
-        test_batch_size = max(1, len(self._subset_test[0]))
+        if not test_batch_size:
+            # case where test_batch_size=0 or test_batch_size=None, ie where user doesnot want
+            # to use test_batch_size
+            test_batch_size = len(self._subset_test)
```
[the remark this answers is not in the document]
ybouilla: That could be a nice idea, but I thought it would have gone beyond the scope of this PR. Basically your idea is to set a threshold to something like `limit`
```python
limit = 1
for elem in dataset.shape:
     limit *= elem
```
But usually dataset shape is `[dataset size, 1st dim, 2nd dim, ...]`
for instance for images: `[1000, 28, 28]`, limit would be `limit = 784`, is that what you were thinking` ? Take the MNIST dataset,  you could use the whole dataset instead of setting a limit. 

In my opinion `test_batch_size` should be used in very specific and rare cases, therefore i don't think it could be that useful to have a limit.

fedbiomed/common/metrics.py
```diff
@@ -422,10 +422,10 @@ def _configure_y_true_pred_(y_true: Union[np.ndarray, list],
             # If y_pred and y_true is 2D array
             # Example: y_true: [ [0,1],[1,0]] | y_pred : [[-0.2, 0.3], [0.5, 1,2 ]]
             elif output_shape_y_pred > 0 and output_shape_y_true > 0:
-
+                print("INTERMEDIATE", y_pred.shape, y_true.shape, output_shape_y_pred, output_shape_y_true)
```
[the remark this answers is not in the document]
ybouilla: yes it is, i am going to remove it

fedbiomed/common/training_plans/_base_training_plan.py
ybouilla: Oh sorry, actually it was more for the review: what do you think about this condition statement? I slightly prefer what I did since it is "simple and stupid" (KISS), but one can use  `MiniBatchTrainingIterationAccountant` for a more advanced but complex solution

docs/user-guide/researcher/tensorboard.md
ybouilla: I would update your suggestion that way: 

Tensorboard displays the results of the validation/testing steps. During each `Round`, values will be exported into `Tensorboard` in real time. If `test_batch_size` is set to a specific value, each computed value will be reported (depending on the size of `test_batch_size` set).

fedbiomed/common/training_plans/_base_training_plan.py
srcansiz: I am not sure if it is compatible or not. It seems that `MiniBatchTrainingIterationAccountant` is designed for training and it accesses training plan attributes related to training such as `batch_size`, `num_updates` etc. It could be nice to make this class more generic in order to use it in testing or anywhere else but for now I think your implementation is fine. However, you can create an issue to propose to use `MiniBatchTrainingIterationAccountant`.

fedbiomed/common/data/_sklearn_data_manager.py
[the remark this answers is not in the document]
srcansiz: Hi @ybouilla, 

I agree that batch size can be used in very specific cases. What I mean is that using test batch size is decided by researcher and it is not safe since it will block the memory on machine that node is running if researcher does not set it. This may create other bugs. For example, training fails due to memory issue but at the same time there may be another request coming to node (eg dataset search) it does not work because the memory is full. 

Therefore, node can set a limit for batch size to avoid such issues. This issues is also existing for training, and I agree that it is beyond the scope of this PR. It could be nice to create an issue to investigate it more.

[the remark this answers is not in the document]
ybouilla: Ok what i wanted to point out is that even if your idea is nice,  it is not something easy to do. Understood, let's open an issue for that!

fedbiomed/common/training_plans/_base_training_plan.py
ybouilla: If it is ok for you,  i will remove this comment. I don't think we will have the opportunity to improve these lines of code 
in the future

srcansiz: Ok that's fine then.

fedbiomed/common/metrics.py
srcansiz: After checking some implementation in batch iterator. I saw that `NPDataLoader` adds another dimension of target variable if it is 1D.
```python
        # If the researcher gave a 1-dimensional dataset, we expand it to 2 dimensions
        if dataset.ndim == 1:
            dataset = dataset[:, np.newaxis]

        # If the researcher gave a 1-dimensional target, we expand it to 2 dimensions
        if target.ndim == 1:
            target = target[:, np.newaxis]
```
Issue can be solved by disabling `newaxsis` for `target` variable. Instead, it can be implemented in `BaseSkLearnModel.train` function. 


```python 
    def train(
        self,
        inputs: np.ndarray,
        targets: np.ndarray,
        stdout: Optional[List[List[str]]] = None,
        **kwargs,
    ) -> None:
        """Run a training step, and record associated gradients.

        Args:
            inputs: inputs data.
            targets: targets, to be fit with inputs data.
            stdout: list of console outputs that have been collected
                during training, that contains losses values.
                Used to plot model losses. Defaults to None.

        Raises:
            FedbiomedModelError: if training has not been initialized.
        """
        if inputs.ndim == 1:
            dataset = inputs[:, np.newaxis]

        # If the researcher gave a 1-dimensional target, we expand it to 2 dimensions
        if targets.ndim == 1:
            targets = targets[:, np.newaxis]

        batch_size = inputs.shape[0]
        w_init = self.get_weights()
        w_updt = {key: np.zeros_like(val) for key, val in w_init.items()}
        # Iterate over the batch; accumulate sample-wise gradients (and loss).
        for idx in range(batch_size):
            # Compute updated weights based on the sample. Capture loss prints.
            with capture_stdout() as console:
                self.model.partial_fit(inputs[idx : idx + 1], targets[idx])
            if stdout is not None:
                stdout.append(console)
            # Accumulate updated weights (weights + sum of gradients).
            # Reset the model's weights and iteration counter.
            for key in self.param_list:
                w_updt[key] += getattr(self.model, key)
                setattr(self.model, key, w_init[key].copy())
            self.model.n_iter_ -= 1
        # Compute the batch-averaged, learning-rate-scaled gradients.
        # Note: w_init: {w_t}, w_updt: {w_t - eta_t * sum_{s=1}^B(grad_s)}
        #       hence eta_t * avg(grad_s) = w_init - (w_updt / B)

        self._gradients = {
            key: w_init[key] - (w_updt[key] / batch_size)
            for key in self.param_list
        }
```

ybouilla: Thanks for this finding: let's resolve conversation and update the related issue #1013

srcansiz (approved):
Hi @ybouilla,

Thanks again for the PR and the great work. I am running e2e test before merging. I am going to merge it once there are passed successfully.

Issue #804, opened 2023-06-21
Insecure validation sec agg


rtaiello:
After a discussion with @srcansiz, we realized that the validation of the aggregation through the encryption of random values, also known to the aggregator can be an information leakage.
There might be several approaches, depending on the assumptions that we can make, here the list of a couple of possibilities:
- https://openreview.net/pdf?id=WA7I-Fm4tmP
- https://eprint.iacr.org/2015/562

mvesin:
Implementing a change to make secagg validation:
- enabled by default in `development` mode
- disabled by default in `vpn` container mode

- [ ] implement in library as described below
- [ ] test in VPN/container mode
- [ ] adapt unit tests
- [ ] adapt documentation

Implemented solution:
1. when generating a node/researcher config file, create a param `security > secagg_insecure_validation`, boolean that defaults to `True`
2. when launching node/researcher or creating config file, can overwrite this value with `SECAGG_INSECURE_VALIDATION` environment variable. If not overwriting: use value from file.
3. in VPN/container mode documentation (later: scripts) execute commands that *disable* validation in config file
4. at run time
    * if validation is disabled on researcher, then it sends a `None` value for `secagg_random`. If validation is enabled then it sends a float (unchanged)
    * if validation is disabled on node and received `secagg_random` is not `None` then node sends an error to the researcher and does not train
    * if node receives a `secagg_random` which is `None`, then its reply has `encryption_factors` to `None`
    * if researcher receives a node reply with `encryption_factor` to `None` while it sent a float `secagg_random`, then it fails

PR #1166, opened 2024-08-20
Make secagg insecure validation optional

mvesin:
**PR description**

- implement #804 for making secagg insecure validation optional
- misc: fix for secagg end2end test & `DefaultStrategy()` ; adapt VPN documentation for LOM secagg ; update `Environ` docstrings

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/fedbiomed/common/environ.py
+++ b/fedbiomed/common/environ.py
@@ -7,40 +7,26 @@ All environment/configuration variables are provided by the
 
 **Environ** is a singleton class, meaning that only an instance of Environ is available.
 
-Descriptions of global/environment variables
 
-Researcher Global Variables:
+Description of the common Global Variables:
 
-- RESEARCHER_ID           : id of the researcher
-- ID                      : equals to researcher id
-- TENSORBOARD_RESULTS_DIR : path for writing tensorboard log files
-- EXPERIMENTS_DIR         : folder for saving experiments
-- MESSAGES_QUEUE_DIR      : Path for writing queue files
-
-Nodes Global Variables:
-
-- NODE_ID                           : id of the node
-- ID                                : equals to node id
-- MESSAGES_QUEUE_DIR                : Path for queues
-- DB_PATH                           : TinyDB database path where datasets/training_plans/loading plans are saved
-- DEFAULT_TRAINING_PLANS_DIR        : Path of directory for storing default training plans
-- TRAINING_PLANS_DIR                 : Path of directory for storing registered training plans
-- TRAINING_PLAN_APPROVAL            : True if the node enables training plan approval
-- ALLOW_DEFAULT_TRAINING_PLANS      : True if the node enables default training plans for training plan approval
-
-Common Global Variables:
-
-- COMPONENT_TYPE          : Node or Researcher
-- CONFIG_DIR              : Configuration file path
-- VAR_DIR                 : Var directory of Fed-BioMed
-- CACHE_DIR               : Cache directory of Fed-BioMed
-- TMP_DIR                 : Temporary directory
-- MPSPDZ_IP               : MP-SPDZ endpoint IP of component
-- DEFAULT_BIPRIMES_DIR    : Path of directory for storing default secure aggregation biprimes
-- ALLOW_DEFAULT_BIPRIMES  : True if the component enables the default secure aggregation biprimes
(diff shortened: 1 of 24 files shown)


fedbiomed/common/environ.py
```diff
@@ -183,3 +168,10 @@ def set_environment(self):
             "MPSPDZ_CERTIFICATE_PEM",
             os.path.join(self._values["CONFIG_DIR"], public_key)
         )
+
+        # make `secagg_insecure_validation` optional in config file (and avoid changing config file version)
+        # by using a `fallback` value
+        secagg_insecure_validation = self.config.get('security', 'secagg_insecure_validation', fallback='true')
```
srcansiz, line 174: Please find the suggestion for reducing the size of the line and a shorter comment.  

```suggestion
        # Optional secagg_insecure_validation optional in config file
        secagg_insecure_validation = self.config.get(
            'security', 'secagg_insecure_validation', fallback='true')
```

fedbiomed/node/round.py
```diff
@@ -350,11 +358,14 @@ def run_model_training(
...
-                    weight=results['sample_size'],
-                )
+                if self._secure_aggregation.scheme.secagg_random is not None:
+                    results["encryption_factor"] = self._secure_aggregation.scheme.encrypt(
+                        params=[self._secure_aggregation.scheme.secagg_random],
+                        current_round=self._round,
+                        weight=results['sample_size'],
+                    )
+                else:
+                    results["encryption_factor"] = None
```
srcansiz, line 368: Encryption factor can be `None` by default `results["encryption_factor"]`

```suggestion
                results["encryption_factor"] = None
                if self._secure_aggregation.scheme.secagg_random is not None:
                    results["encryption_factor"] = self._secure_aggregation.scheme.encrypt(
                        params=[self._secure_aggregation.scheme.secagg_random],
                        current_round=self._round,
                        weight=results['sample_size'])                 
```

fedbiomed/researcher/secagg/_secure_aggregation.py
```diff
@@ -306,8 +306,11 @@ def _configure_round(
             experiment_id: The id of the experiment
         """
 
-        # For each round it generates new secagg random float
-        self._secagg_random = round(random.uniform(0, 1), 3)
+        if environ['SECAGG_INSECURE_VALIDATION'] is True:
+            # For each round it generates new secagg random float
+            self._secagg_random = round(random.uniform(0, 1), 3)
+        else:
+            self._secagg_random = None
```
srcansiz, line 313:
```suggestion
        self._secagg_random = None
        if environ['SECAGG_INSECURE_VALIDATION'] is True:
            # For each round it generates new secagg random float
            self._secagg_random = round(random.uniform(0, 1), 3)            
```

fedbiomed/researcher/secagg/_secure_aggregation.py
```diff
@@ -331,18 +334,18 @@ def aggregate(
...
-        num_expected_params: int | None = None
+        encryption_factors: Dict[str, Union[List[int], None]],
+        num_expected_params: int | None = None,
     ) -> bool:
         """Validate given inputs"""
 
-
-        if encryption_factors is None:
+        # at this point we know `isinstance(encryption_factors, dict) and len(encryption_factors) == num_nodes`
+        if any([v is None for v in encryption_factors.values()]):
```
srcansiz, line 343: if think `all(encryption_factors.values())` does the same think. 

```suggestion
        if not all(encryption_factors.values()):
```

srcansiz (commented):
Hi @mvesin,

Thank you for the PR. Generally, the PR is good to go. Please find my suggestions about a few little details.  

I would also suggest to add a basic e2e test to test the error case.

fedbiomed/researcher/secagg/_secure_aggregation.py
mvesin: This is not exactly the same test. For example with `encryption_factors.values = [False, [1]]` tests don't return the same value. Testing specifically for `None` value intends to catch only the case of setup mismatch between researcher and node, while doing a `not all()` test will also catch some malicious attack cases, misleading the user to believe this is a setup mismatch (while malicious case are caught in `SecaggCrypter` with explicit message indicating the type of the encryption factors is not correct).

Sio I would prefer to keep the current more explicit test

PR #1196, opened 2024-09-09
Feature/1193 create additive secret sharing context researcher

### LINKED ISSUE #1193: Create researcher `SecaggKeyContext ` for additive secret sharing in `fedbiomed/researcher/secagg/secagg_context`
Create researcher `SecaggKeyContext ` for additive secret sharing in `fedbiomed/researcher/secagg/secagg_context`

- The implementation will be very similar to `SecaggDHContext`.
- If it is needed a base for that is goingto be inherited by`SecaggDHContnext` and `SecaggKeyContext` can be created.

ybouilla:
**PR description**

ScaggKeyContext on Researcher side

Closes #1193 

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/fedbiomed/researcher/secagg/_secagg_context.py
+++ b/fedbiomed/researcher/secagg/_secagg_context.py
@@ -11,7 +11,7 @@ from abc import ABC, abstractmethod
 import time
 import random
 
-from fedbiomed.common.secagg._additive_ss import Share, Shares
+from fedbiomed.common.secagg import AdditiveShare, AdditiveShares
 from fedbiomed.researcher.environ import environ
 from fedbiomed.researcher.requests import Requests, StopOnDisconnect, StopOnError, \
     StopOnTimeout
@@ -637,9 +637,9 @@ class SecaggKeyContext(SecaggContext):
             # Additive Secret Sharing: sum up shares from nodes
             for rep in replies.values():
                 share = rep.share
-                context['share'].append(Share(share))
+                context['share'].append(AdditiveShare(share))
 
-        context['share'] = Shares(context['share']).reconstruct()
+        context['share'] = AdditiveShares(context['share']).reconstruct()
 
         if all(status.values()):
             # only create context if previous steps succeeded
--- a/fedbiomed/researcher/secagg/_secagg_context.py
+++ b/fedbiomed/researcher/secagg/_secagg_context.py
@@ -11,6 +11,7 @@ from abc import ABC, abstractmethod
 import time
 import random
 
+from fedbiomed.common.secagg._additive_ss import Share, Shares
 from fedbiomed.researcher.environ import environ
 from fedbiomed.researcher.requests import Requests, StopOnDisconnect, StopOnError, \
     StopOnTimeout
@@ -92,6 +93,12 @@ class SecaggContext(ABC):
 
         # to be set in subclasses
         self._secagg_manager: Optional[BaseSecaggManager] = None
+
+    def _raise_if_missing_parties(self, parties: List[str], required_nb_parties: int):
+        if len(parties) < required_nb_parties:
+            raise FedbiomedSecaggError(
+                f'{ErrorNumbers.FB415.value}: {self._element.value}, bad parameter `parties` : {parties} : need '
+                f'at least {required_nb_parties} nodes for secure aggregation')
 
     @staticmethod
     def _check_secagg_id_type(value) -> bool:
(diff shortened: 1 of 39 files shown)


fedbiomed/researcher/secagg/_secagg_context.py
```diff
@@ -539,6 +543,111 @@ def _create_payload_specific(self) -> Tuple[Union[dict, None], bool]:
...
+        return context, True
+
+    def setup(self) -> bool:
+        msg = ResearcherMessages.format_outgoing_message({
+            'researcher_id': self._researcher_id,
+            'secagg_id': self._secagg_id,
+            'element': self._element.value,
+            'experiment_id': self._experiment_id,
+            'parties': self._parties,
+            'command': 'secagg-additive-ss-setup-request',
```
srcansiz, line 595: Hi @ybouilla,

I recommend you to use directly `AdditiveSSSetupRequest`, and since there is a defualt value for `command` you won't need to specify it.

fedbiomed/researcher/secagg/_secagg_context.py
```diff
@@ -539,6 +543,111 @@ def _create_payload_specific(self) -> Tuple[Union[dict, None], bool]:
...
+        status = {node_id: False for node_id in self._parties}
+
+        # one controller per secagg object to prevent any file conflict
+
+        context = {'share': []}
+        with self._requests.send(msg, self._parties, policies) as fed_request:
+            replies = fed_request.replies()
+            errors = fed_request.errors()
+
+            status = {rep.node_id: rep.success for rep in replies.values()}
```
srcansiz, line 634: I know that it is in the scope of this PR but I recommend to make `success` field optional, and ignore checking this value. `fed_request.errors()` collects `ErroMessage`s so on the node side we can implement it in a way if there there is an error during setup node just return `ErrorMessage`.

fedbiomed/researcher/secagg/_secagg_context.py
```diff
@@ -560,14 +669,9 @@ def __init__(self, parties: List[str], experiment_id: str, secagg_id: Union[str,
...
-
-        if len(parties) < 2:
-            raise FedbiomedSecaggError(
-                f'{ErrorNumbers.FB415.value}: LOM, bad parameter `parties` : {parties} : need '
-                'at least 2 nodes for secure aggregation')
-
-
         self._element = SecaggElementTypes.DIFFIE_HELLMAN
+        self._raise_if_missing_parties(parties, required_nb_parties=2)
+
```
srcansiz, line 674: Hi @ybouilla, 

I think a nice way of implementing it could be having a class attribute `_min_num_parties`, and check it in base `SecaggContext.__init__`, as it is done in SecaggSetup classes on the node side.

ybouilla: I agree with you, but in the previous implementation, there were no `_mion_num_parties`. I can implement it though!

fedbiomed/researcher/secagg/_secagg_context.py
ybouilla: Ok I can fix this

fedbiomed/researcher/secagg/_secagg_context.py
ybouilla: As you prefer, but the base class `setup` method doesnot provide `SecaggRequest`, but  uses the `ResearcherMessages.format_outgoing_message` instead. I can change both (`SecaggContext` and `SecaggKeyContext`)

fedbiomed/researcher/secagg/_secagg_context.py
ybouilla: Actually, i was wondering if we can have this edge case: if all the nodes doesnot reply before timeout? Or if for some reason the message is badly formatted and errors are not registered inside the fed_request? DO you think this could ever happen?

srcansiz: If the errors are badly formated fed request will fail due to time out. It is already checked in the code `fed_request.has_stopped_any()`

ybouilla: Any if an error happened during the `Round` run? 
eg if for some reason breakpoint is not saved?
 `self._send_round_reply(success=False, message="Can't save new node state.")` 
For what I understood the node only sends `TrainReply`

srcansiz: >And if a node to node connection fails? will we get a ErrorMessage?

Yes

PR #1206, opened 2024-09-16
Implements Message serialize and deserialize methods

srcansiz:
- Adds methods to serialize and deserialize messages.
- Removes 'command' property and uses Message (request, reply) class names
- Removes format_outgoing and format_incoming methods uses serialize, deserialize
- Updates unittests and all related implementation within the library.

**PR description**

TO_BE_FILLED_BY_PR_CREATOR

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:


fedbiomed/common/message.py
```diff
@@ -83,10 +85,24 @@ def __post_init__(self):
...
+                ErrorNumbers.FB601.value
+                + ": bad input value for message: "
+                + self.__str__()[0:1000]
+                + "..."
+            )
             logger.critical(_msg)
             raise FedbiomedMessageError(_msg)
 
+    @property
+    def __name__(self) -> str:
```
mvesin, line 98: Thank you @srcansiz for this nice refactor !

I would propose to use `name()` rather than `__name__()` as it is meant to be a public method.

```suggestion
    def name(self) -> str:
```

fedbiomed/common/message.py
```diff
@@ -116,29 +132,61 @@ def __validate(self, fields: Dict[str, Any]) -> bool:
...
             value = getattr(self, field_name)
             if not isinstance(value, field_def.type):
-                logger.critical(f"{field_name}: '{value}' instead of '{field_def.type}'")
+                logger.critical(
+                    f"{field_name}: '{value}' instead of '{field_def.type}'"
+                )
                 ret = False
         return ret
 
+    def serialize(self) -> Dict:
```
mvesin, line 141: Sorry for discussing details, but the names `(de)serialize` look to me misleading. We are not doing serialization there, but conversion to/from `dict` to `Message`

Maybe rename as `to_dict` and `to_message` ? or other ....

fedbiomed/common/message.py
```diff
@@ -83,10 +85,24 @@ def __post_init__(self):
         """
 
         if not self.__validate(self.__dataclass_fields__.items()):
-            _msg = ErrorNumbers.FB601.value + ": bad input value for message: " + self.__str__()[0:1000] + "..."
+            _msg = (
```
mvesin, line 88: We miss some validation when changing fields in a message

OK when creating a message:

```
mes = PingReply(
                    request_id=4,
                    researcher_id='my researcher',
                    node_id='my node',
 )
...
FedbiomedMessageError: FB601: message error: bad input value for message: PingReply(protocol_version='4', request_id=4, researcher_id='my researcher', node_id='my node')...
```

But not checked when changing a message

```
mes = PingReply(
                    request_id='123456 string',
                    researcher_id='my researcher',
                    node_id='my node',
 )
# no error
mes.request_id = 124
# returns int
type(mes.request_id)
```

fedbiomed/common/message.py
```diff
@@ -402,27 +461,13 @@ class KeyReply(InnerRequestReply, RequiresProtocolVersion):
...
-# @catch_dataclass_exception
-# @dataclass
-# class DummyInner(InnerMessage, RequiresProtocolVersion):
-#     """Dummy example of one-wway (not request-reply) inner message
-#     """
-#     dummy: str     # Temporary dummy payload
-#     command: str
-
-
-# --- Node <=> Researcher messages ----------------------------------------------
```
mvesin, line 425: Misc: even if we suppress the example of one-way n2n message we need to keep the separation of node <=> researcher messages

```
# --- Node <=> Researcher messages ----------------------------------------------
```

fedbiomed/common/message.py
```diff
@@ -14,10 +14,13 @@
...
 import fedbiomed.transport.protocols.researcher_pb2 as r_pb2
 
-from fedbiomed.common.constants import ErrorNumbers, __messaging_protocol_version__, \
-    SecureAggregationSchemes
-from fedbiomed.common.utils import raise_for_version_compatibility
-from fedbiomed.common.exceptions import FedbiomedMessageError
+from fedbiomed.common.constants import (
+    ErrorNumbers,
+    __messaging_protocol_version__,
+    SecureAggregationSchemes,
```
mvesin, line 20:
```suggestion
- SecureAggregationSchemes,
```

misc: unused import

fedbiomed/common/message.py
mvesin: We can also set the "hidden" attributes without check

```
n [60]: mes
Out[60]: PingReply(protocol_version='4', request_id='my id', researcher_id='my researcher', node_id='my node')

In [61]: mes.protocol_version = 1

In [62]: mes
Out[62]: PingReply(protocol_version=1, request_id='my id', researcher_id='my researcher', node_id='my node')
```

fedbiomed/common/message.py
```diff
@@ -116,29 +132,61 @@ def __validate(self, fields: Dict[str, Any]) -> bool:
...
+
+        message = {**obj}
+
+        if "__type_message__" not in message:
+            raise FedbiomedValueError(
+                "Message does not include '__type_message__', can not serialize"
+            )
+
+        type_ = message["__type_message__"]
+        cls_ = import_object(type_["module"], type_["class"])
```
mvesin, line 164: This part is security sensitive !
We cannot trust anything that is received from the network unless it was previously checked by the receiver component. 

Major security issue ? we cannot trust the  `type_["module"]` => it should always be `fedbiomed.common.message` 
This could also be enforced in the places (eg `fedbiomed.transport`) were it is actually security sensitive, but I would rather always check it in `Message.deserialize()` to avoid unwanted security hole in the future.

Minor point: we should check that the keys exist in `type_` before accessing them. Don't see any direct exploit, but as the code is security sensitive let's be cautious ...

```
type_ = message["__type_message__"]
if ('module', 'class') in type_:
   ...
```

fedbiomed/common/message.py
srcansiz: Ok. However, this is just a comment. I think that  if you want to separate these different messages the correct way to do it 
1 - create sperate modules
2 - make them inherit from class that is named explicitly, for example -> `overlay -> NodeToNode`

fedbiomed/common/message.py
srcansiz: This is not something that I introduced, it was already the case. So, if you think the it is important I propose to create a separate issue for that.

fedbiomed/common/message.py
srcansiz: Hi @mvesin,

`__name__` is known convention to get class name, and using it outside of the class is not against PEP8 or flake8. Therefore, I wanted to use `__name__`, and it is a property rather than a method. So it is used as `message.__name__`. I can not do the same `message.name` because `name` can be a property of a dataclass.

fedbiomed/common/message.py
mvesin: Sure you did not introduce it. I wondered if we wanted to remove while refactoring `Message` :-)
OK i'll create a separate issue to track this point

fedbiomed/common/message.py
mvesin: I leave you the last word to decide on that point ...

mvesin (approved):

Issue #1202, opened 2024-09-12
Use symmetric encryption for node to node communications


mvesin:
Replace asymmetric encryption of node to node messages with symmetric encryption.
Use the `DHKeyAgreement` class, with a `salt` generated by sending node and sent cleartext in `OverlayMessage`

PR #1218, opened 2024-09-30
Node to node symmetric encryption

mvesin:
**PR description**

Complete #1202

**Caveat** you probably need to `source ./scripts/fedbiomed_environment clean` to remove you existing n2n channel keys in the DB (as the format changed from RSA to EC). I did not increment the channel DB version number, as current version (v1) has just been merged to develop, and not yet included in a release.

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:


srcansiz:
Hi @mvesin,

Thanks for the PR.

I've been reading DH documentation from `cryptoprahie` and they strongly recommend generating `private_key` for each `.exchange` execution (DHE). https://cryptography.io/en/3.4.8/hazmat/primitives/asymmetric/dh.html to provide forward secrecy. However, it is provided using `nonce` in `ChaCha20` algorithm, or `salt` in key derivation function, but not validated if same salt or nonce is used, it is normal since it requires more detailed implementation. 

Then, I thought about using DHE which will ensure that the session secret will never be the same even there is no salt or nonce is set. Plus, it will avoid creating  entries in the database which will provide simpler implementation. 

Sorry, I might miss lead you using same key for multiple sessions, and I know that it may be too late to change it but what do you think about this improvement in the feature that will avoid implementing another layer for making sure that same nonce and salt not used more than a time ?

mvesin:
Thank you @srcansiz for the review. Raising the issue of forward secrecy is a good point.

Regarding forward secrecy: yes the current implementation is not fully forward secret compliant (in the case the channel certificate is compromised).
More generally, we know (and as I understood admitted) that the node to node secure channel is not perfect security (some replay attacks are also possible) in current implementation.
I don't see the issue around re-using salt/nonce (should not happen IMO).
I suspect if we want to go further we would need to completely re-implement the node to node to be based on some messaging protocol that implements "perfect" security - this is probably not our focus and beyond our skills.
As far as I remember we deemed the current design as acceptable for now. Let's remember this happens withing a node-researcher TLS (forward secret) communication, so I suspect only malicious researcher might do this kind of advanced attacks.
Moreover, we do another (independant) DH exchange within the secure channel for setting up a training's secagg keys. These keys are specific to one experiment, and independant/unknown from the long term (n2n channel keys). In that sense our DH-in-DH achieves that property of forward secrecy that compromising long term n2n channel key does give experiment secagg keying material that would enable to decrypt previous session's local parameters from nodes.

So I would suspect that:
- all that you mention is very relevant, but residual risk is low
- if we want to address we probably want to totally reshape n2n communications to base on some secure messaging protocol (which may not be that simple to integrate ...)

We can discuss that tomorrow if you want.

I don't think we want to go further on that topic for now, but we can mention this in some issue...

PR #1261, opened 2025-01-07
Feature/1226 flamby optional

sharkovsky:
# Flamby is not required anymore

The goal of this PR is to
- remove flamby from fedbiomed dependencies
- fedbiomed can still use flamby code if the user has installed flamby separately
- fedbiomed should print out nice warnings to the user when they try to use flamby code but it was not installed

## Strategy

- If some code simply defines constants or other non-executing code related to flamby, keep it unchanged
- all flamby-related imports are done conditionally, only in the case where the user wants to use flamby-specific code
- before importing the actual flamby library, test whether the module exists and print out a nice error message in case of `ModuleNotFoundError`
- removed flamby from `pyproject.toml`

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/docs/tutorials/flamby/flamby-integration-into-fedbiomed.ipynb
+++ b/docs/tutorials/flamby/flamby-integration-into-fedbiomed.ipynb
@@ -2,8 +2,6 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "99b1c6a5",
-   "metadata": {},
    "source": [
     "# FLamby integration in Fed-BioMed\n",
     "\n",
@@ -43,7 +41,11 @@
     "pip install wget\n",
     "python $(find $CONDA_PREFIX -path */fed_heart_disease/dataset_creation_scripts/download.py) --output-folder ${FEDBIOMED_DIR}/data\n",
     "```\n"
-   ]
+   ],
+   "metadata": {
+    "collapsed": false
+   },
+   "id": "8cdd50d733af9d2f"
   },
   {
    "cell_type": "markdown",
@@ -137,7 +139,8 @@
    "source": [
     "from fedbiomed.common.training_plans import TorchTrainingPlan\n",
     "from flamby.datasets.fed_ixi import Baseline, BaselineLoss, Optimizer\n",
-    "from fedbiomed.common.data import FlambyDataset, DataManager\n",
+    "from fedbiomed.common.data.flamby_dataset import FlambyDataset\n",
+    "from fedbiomed.common.data import DataManager\n",
     "\n",
     "\n",
     "class MyTrainingPlan(TorchTrainingPlan):\n",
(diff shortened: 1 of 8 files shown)


sharkovsky:
A couple of choices to discuss:
- I removed the leading underscore from `flamby_dataset.py` because it needs to be imported directly by other modules, and the IDE prints warnings about importing a protected file. If you prefer, we can re-instate the underscore and live with the IDE's warning
- I chose to log the flamby not found error using the fedbiomed logger, but I realise my reasons for this were incorrect: I thought the error should be propagated to the researcher, but this is not the case since the error appears at dataset add time. Maybe a simple print would be sufficient?

srcansiz:
> A couple of choices to discuss:
> 
> * I removed the leading underscore from `flamby_dataset.py` because it needs to be imported directly by other modules, and the IDE prints warnings about importing a protected file. If you prefer, we can re-instate the underscore and live with the IDE's warning
> * I chose to log the flamby not found error using the fedbiomed logger, but I realise my reasons for this were incorrect: I thought the error should be propagated to the researcher, but this is not the case since the error appears at dataset add time. Maybe a simple print would be sufficient?

Hi @sharkovsky,

Thank you for the implementation. 

- I think it is ok to leave `flamby_dataset` without underscore. 
- About the logging, I find that raising error is enough because there will be two places that user may get this error 
    - CLI, while training to add dataset: Exception will print the error anyway.
    - User who uses `FlambyDataset` (see: https://fedbiomed.org/latest/tutorials/flamby/flamby/). 
    - The case where Flamby installed on the researcher but not on the node. Round will catch import error and report it researcher with an error message.   
    
  
  Regarding 2 last points, I guess there is only the import statement needs to me fixed in `tutorials/flamby`.

sharkovsky:
Sorry for missing the tutorials, they should be fixed now. 

Also, I removed logging of the error as suggested (I agree with your point). 

The PR is again ready for review :D

srcansiz:
> Sorry for missing the tutorials, they should be fixed now.
> 
> Also, I removed logging of the error as suggested (I agree with your point).
> 
> The PR is again ready for review :D

Thank you for the changes. There is one last thing for the tests. Since `Flamby` is not part of the dependencies. `tox` won't install it by default, because of that `tox.ini` needs to be updated accordingly.

```ini
[tox]
requires =
    tox>=4

envlist = py{310}
isolated_build = True

[testenv]
deps =
	pytest
	pytest-cov
	git+https://github.com/owkin/FLamby@main
extras=
```

- Another point (optional): `Falmby` can be part of `dev-dependencies` in `pyproject.toml`. This will allow developers to have `Flamby` in their environment by default but not in PyPi (I assume that PyPi wont be angry about it).  This can avoid getting module not found error if tests are executed using `pytest` only.  I am not sure about this modification so I will leave it to you to add if you think that it is useful.

srcansiz:
> Added dependency in tox. I don't think we should install it by default as developer dependency in pyproject for two reasons: it really isn't necessary for development, plus this way we make sure that the developer process stays as similar as possible to the user process.

Hi @sharkovsky,
Ok perfect! The PR is ok for me. I will squash your commit and rebase to base branch. Thanks again for the PR.

Issue #1308, opened 2025-03-18
Run multiple framework container instances simultaneously


mvesin:
As a user, I want to run multiple instances of Fed-BioMed (in containers/VPN mode) simultaneously on the same machine, each instance being part (or total) of a distinct Fed-BioMed federation.

I want not only to be able to deploy, handle images (shared or not between the instances), handle configurations for the instances, but also have containers for 2 different instances of Fed-BioMed running at the same time. For each federation all or only some components may be running on the machine (eg fed1 has vpnserver + researcher +2 nodes ; fed2 has only one node).

For example I want to be able to run node(s) participating in 2 different federations running on the same machine at the same time.

This requires (at least):
- [ ] add `docker compose` support for 2 instances of the services running at the same time (eg: use distinct service names for a component in different federation)
- [ ] add support for different network ports
- [ ] other remaining shared resources to distinguish between the instances ?

Issue #1306, opened 2025-03-18
improve container network isolation configuration


mvesin:
As a user, I want improved container network configuration to better isolate components running in VPN/container mode.

Currently, all components are launched using a `default_docker` network automatically created by `docker compose`

PR #1309, opened 2025-03-18
improve container network configuration + misc

mvesin:
**PR description**

- implement #1306 by creating one docker network per component, plus optionally use different networks if another instance of Fed-BioMed is running on the same machine
- use different names (but same service) for containers in 2 different instances. Please note this permits better identifying the related instance of the containers, but this is not enough to enable 2 instances to execute in parallel ( #1308 covers it)

- fix regression error in CI for testing docker containers introduced in https://github.com/fedbiomed/fedbiomed/commit/ad864605030d8afb6629f5bc3776d78f5de3640c (not related, but avoid another PR for one line fix)

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:


envs/vpn/docker/docker-compose.yml
```diff
@@ -246,3 +268,20 @@ services:
...
+    name: fedbiomed_node_${FBM_CONTAINER_INSTANCE_ID:-unknown}
+  fedbiomed_node2:
+    name: fedbiomed_node2_${FBM_CONTAINER_INSTANCE_ID:-unknown}
+  fedbiomed_gui:
+    name: fedbiomed_gui_${FBM_CONTAINER_INSTANCE_ID:-unknown}
+  fedbiomed_gui2:
+    name: fedbiomed_gui2_${FBM_CONTAINER_INSTANCE_ID:-unknown}
+  # for intermediate containers, usuallly not used
+  fedbiomed_misc:
+    name: fedbiomed_misc_${FBM_CONTAINER_INSTANCE_ID:-unknown}
```
srcansiz, line 287: Hi @mvesin,

Correct me if I am wrong but the interest of having docker container networks it to let containers talk to each other efficiently and securely. As I know, it useful if there are multiple containers launched in a machine and required to talk to each other.  

However, there I see each container has it is own network, and I don't see the point of having separated networks.  Correct me if I am wrong but each container will have it is own network, so this implementation will also cause problem for local testing deployments if one day we have an option to deploy without VPN network. Containers won't be able to talk to each other using localhost. 

Finally, please tell me if I miss anything and what this implementation will provide.

srcansiz (commented):
Hi @mvesin,

Thanks for the PR. I have only one comment for the PR about why this implementation necessary. You can see my comment below.

envs/vpn/docker/docker-compose.yml
mvesin: Hi @srcansiz 
The interest of partitioning networks is to increase security. It makes sense to use distinct network from other applications (protect from other apps) and to use distinct network for each component of the application (this limits the attack surface between components).

For example, in Fed-BioMed, if deploy researcher and node on the same machine, their communications are then limited to the published ports (while you could connect to any port from one to another if they share the same network).

Regarding the future support of containers that would not use VPN : they would not be able to communicate using `localhost` from one container to another with the current implementation. Yet they could use the docker network address, which won't be possible with separate network. In both implementations, you can use `localhost` from the host machine.

PR #1325, opened 2025-05-19
Exception for empty list of nodes before starting a training job.

### LINKED ISSUE #1127: Node sampling history of strategy is erased when nodes fail
When nodes fail during an experiment, they are also removed from the sampling history of the strategy class. 
This should not happen, because the purpose of the history is to keep track of which nodes had been selected by the strategy, not only those that didn't fail.  

### Steps to reproduce
on `develop` commit 850f73dc3f9691064cf8ba18f0e7778b41cfbc8d
Based on 101 notebook

1. insert `assert False` in the `training_step` of the training plan. This simulates a node failure. 
2. execute `exp.run`
3. the output of `exp.strategy()._sampling_node_history` is `{0: []}`, while it should contain the ids of the selected nodes.

Furthermore, the experiment fails with `FB401: aggregation crashes or returns an error. Aggregation aborted due to sum of the weights is equal to 0 {}. Sample sizes received from nodes might be corrupted.`. 
However, the fact that no nodes succeeded should be caught before the aggregation is even attempted.

leonard2103:
**PR description**

This closes #1127. 

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/fedbiomed/researcher/federated_workflows/_experiment.py
+++ b/fedbiomed/researcher/federated_workflows/_experiment.py
@@ -737,12 +737,18 @@ class Experiment(TrainingPlanWorkflow):
         if missing:
             raise FedbiomedExperimentError(ErrorNumbers.FB411.value + ': missing one or several object needed for'
                                            ' starting the `Experiment`. Details:\n' + missing)
+        
         # Sample nodes for training
-
         training_nodes = self._node_selection_strategy.sample_nodes(
             from_nodes=self.filtered_federation_nodes(),
             round_i=self._round_current
         )
+        if not training_nodes:
+            raise FedbiomedExperimentError(
+                "Empty list of nodes for training: no nodes replied to original "
+                "`search_request` or sampling strategy returned an empty list."
+            )
+
         # Setup Secure Aggregation (it's a noop if not active)
         secagg_arguments = self.secagg_setup(training_nodes)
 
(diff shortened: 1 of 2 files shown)


fedbiomed/researcher/federated_workflows/_experiment.py
```diff
@@ -737,12 +737,18 @@ def run_once(self, increase: bool = False, test_after: bool = False) -> int:
...
             raise FedbiomedExperimentError(ErrorNumbers.FB411.value + ': missing one or several object needed for'
                                            ' starting the `Experiment`. Details:\n' + missing)
+        
         # Sample nodes for training
-
         training_nodes = self._node_selection_strategy.sample_nodes(
             from_nodes=self.filtered_federation_nodes(),
             round_i=self._round_current
         )
+        if not training_nodes:
```
sharkovsky, line 746: On second thoughts, it might be better to be more explicit in the check:
```suggestion
        if len(training_nodes) == 0:
```

sharkovsky (changes requested):
Great job @leonard2103 ! 
I have a small comment about checking the length of the list instead. 

I don't know if we should require a test for this, maybe @srcansiz has an opinion. I would be fine merging without tests, but I defer to the team's opinion.

sharkovsky (approved):

PR #1328, opened 2025-05-22
Support for both Py3.10 and 3.11 

sharkovsky:
This PR allows Fed-BioMed to officially support Python 3.11 alongside 3.10. 
In addition, it significantly relaxes the constraints on several package dependencies, making Fed-BioMed compatible with more recent numpy, pandas, and torch versions. 

## User-facing versus internal dependencies
This PR introduces a notion of:
- **user-facing dependencies**, i.e. dependencies that are included in Fed-BioMed to support users in writing their federated training code) 
- **internal dependencies**, i.e. dependencies required by the internal Fed-BioMed infrastructure in order to work. 

Some examples:
- user-facing dependencies: torch, scikit-learn, pandas, torchvision, itk, monai, etc..
- internal dependencies: tinydb, grpc, requests, joblib, cryptography, msgpack, etc...

The proposed approach is:
- only specify a minimum version for user-facing dependencies, allowing users large flexibility in the installation of these packages
- continue giving very strict specifications for internal dependencies, pinning them to a very specific version number, trying to minimize the chance of breaking things due to third-party updates. 

### Testing policy

The tests will be run as follows:
1. on every PR to develop, we run the unit tests + mnist tests on the last supported python version (i.e. py3.11)
2. on every PR to master, we run the unit tests + mnist tests on all supported python versions (py3.10,py3,.11)
3. we also run unit tests + mnist tests, based on `develop`, weekly
4. on the two conditions above (2 and 3), we also run an additional mnist test based on an environment created with the `direct_minimal_versions` strategy, that installs the oldest package versions possible. 
5. endurance tests are on the same schedule as before, and run on all python versions
6. e2e tests are on the same schedule as before, but only run on py3.10 to minimize redundancy with point 3 above

### Known issues with third-parties

I had to exclude itk 5.4.4 due to this [issue](https://github.com/InsightSoftwareConsortium/ITK/issues/5399), and opacus 1.5.4 due to this [issue](https://github.com/pytorch/opacus/issues/769). 
As of today, itk has already published a `5.4.4.post1` version (I verified it is correctly installed), while opacus will integrate the changes in 1.5.5.

### Changes with pdm flow

When changing/updating packages, it is important to compute the lock file first by specifying both python  versions as follows:
```bash
pdm lock --python="<3.12,>=3.10" -G :all
```

And only execute `pdm install` after the lock file has been generated. 

This generates a lock file that is valid for both python versions, and that can be committed to github without loss of generality.

Of course, if the development activity does not require playing with package versions, the developer can directly `pdm install` using our github-committed lockfile, just like before.

### Compatibility

Some points have been raised about potential compatibility issues with opacus, flamby and concrete-ml. 

For opacus and flamby, I did some quick and dirty tests on my machine and everything seemed ok. 
For concrete-ml, I did not take the time to test. I would suggest treating this as a separate issue. 

#### ipython

Since recent ipython (`>=9.0.0`) explicitly requires python 3.11, pdm settles on recent `8.*` versions of ipython in order to  support both  python version with a single spec. 

#### conda

Anecdotically, on MacOS, in order to succesfully install Fed-BioMed, I needed to explicitly tell conda to create an environment with a python interpreter for the correct platform (arm) instead of x86.

```shell
conda create --platform osx-arm64 -n fedbiomed-dev-py3.11 python=3.11 -y
```

## Checklist

- [x] build both environments(py3.10 and py3.11)
- [x] run unit tests on both environments
- [x] execute 101-getting-started notebook on both environments
- [x] execute a notebook with DP on both environments
- [x] update testing configurations with both a minimal and a maximal version of packages
- [x] update documentation

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:


tests/test_models.py
```diff
@@ -588,7 +583,7 @@ def test_torchmodel_07_add_corrections_to_gradients(self):
         }
 
         # zeroes gradients model
-        self.model.model.zero_grad()
+        self.model.model.zero_grad(set_to_none=False)
```
srcansiz, line 586: Does it work for `torch < 2`?

tests/test_flamby.py
```diff
@@ -8,13 +8,17 @@
...
+try:
+    from fedbiomed.common.data.flamby_dataset import (
+        FlambyDataset,
+        FlambyDatasetMetadataBlock,
+        FlambyLoadingBlockTypes,
+        discover_flamby_datasets,
+    )
+except ModuleNotFoundError as e:
+    raise unittest.SkipTest(f"Flamby not found with error {e}. Skipping all tests in test_flamby.py")
+
```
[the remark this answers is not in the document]
sharkovsky: This is irrelevant to python version. 
If you prefer, we can create an issue and implement there instead!

tests/test_models.py
sharkovsky: yes, the argument already [existed](https://docs.pytorch.org/docs/1.13/generated/torch.nn.Module.html?highlight=zero_grad#torch.nn.Module.zero_grad) in previous torch versions, but the default value was different.

sharkovsky:
> Have you check it is compatible with opacus?

Thanks for the comment @ybouilla , after your suggestion I quickly checked the `differential-privacy-with-opacus-on-fedbiomed.ipynb` notebook. At least on my system, it works for both py3.10 and py3.11 versions.

tests/test_models.py
srcansiz: Ok great!

tests/test_flamby.py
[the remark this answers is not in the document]
srcansiz: Yes, It could be nice to have this in separate PR and merge into develop quickly since it is annoying to install Flamby each to is want to execute tests. Should I do it or do you want to do it ?

sharkovsky:
I realized I misunderstood a few crucial points. Here is a summary of the current situation:
1. Py 3.11 requires `torch >= 2.0.0`, therefore we need to include this in our requirements
2. This leads to installing `torch 2.0.1` for both py 3.10 and 3.11, since both python versions support torch 2
3. Therefore, the proposed changes to the torch versions are effectively equivalent to dropping `torch `1.13`, since it would never get installed

If we want to 

~~Pytorch is supposed to have a mutually-exclusive support matrix~~
EDIT: the sentence above was due to a misunderstanding on my part. 
Python 3.10 is [compatible](https://github.com/pytorch/pytorch/blob/main/RELEASE.md) with all torch versions, while 3.11 is only compatible with `torch >= 2.0.0`

ybouilla:
> Open question with pytorch: should we follow this conservative approach, or try a more liberal approach such as "torch  >=1.13.0,<3.0.0" ??
Just my 2 cents, but maybe a more conservative approach is safer, because we are also depending on libraries such as `declearn` that may stay idle for some time (ie not including last recent of torch)

PR #1373, opened 2025-06-26
Feature/1335 create base docker images without vpn installation

### LINKED ISSUE #1335: Create base docker images without VPN installation
Current docker images are configured to force VPN packages/modules to be installed and activated. This approach does not allow users to use docker images without VPN installation. Having a version where Fed-BioMed base/minimal docker images are created can help modularize Fed-BioMe docker components, thus, allowing  interoperability and  being more flexible for different deployment scenarios (e.g where users selected VPN network if their choice)

- [ ] Create base docker images that installs required packages and tools for Fed-BioMed components.
    - [ ] Use root privileges as minimum as possible. 
    - [ ] Optimize the size of images #1269 
    - [ ] Avoid reinstallation of same packages that are used different component to gain from build time. 

- [ ] Distribute docker images on Docker hub

This issue can be extended later with the following tasks. 

- [ ] Make VPN images are build on the top of these images.

srcansiz:
**PR description**
closes #1335 

This PR create base Fed-BioMed docker images without VPN installed. 

- It includes Github action that tests and pushed fedbiomed images to docker hub. 
- Adds documentation to explain what are the steps to built docker images from source. 
- Adds documentation for user about how to use ready to use docker images. 

For the review and being able to follow the documentation you may need to build images using: 

Please execute from fedbiomed repo root:
 
```
docker build -f docker/base/Dockerfile -t fedbiomed/base:latest .
docker build -f docker/node/Dockerfile -t fedbiomed/node:latest .
docker build -f docker/researcher/Dockerfile -t fedbiomed/researcher:latest 
```

This PR introduces there docker images `base`, `node`, and `researcher`. GUI comes with `node` images. All other functions and instructions are described in the documentation `docs/user-guide/deployment/docker.md`


docker/node/entrypoint.sh
```diff
@@ -0,0 +1,55 @@
...
+fedbiomed component create --component NODE --path /fbm-node --exist-ok
+
+
+if [ -n "${MNIST_OFF}" ]; then 
+  echo "[INFO] MNIST download and deploy operation is disabled."
+else
+  echo "[INFO] Deploying MNIST dataset by default"
+  fedbiomed node -p /fbm-node dataset add --mnist /fbm-node/data || \
+    printf "[ERROR] MNIST download or deploy operation failed. Please make sure you have an internet connection. Otherwise, please set '-e MNIST_OFF=True' in the docker run command.\n"
+fi
```
sharkovsky, line 38: Why deploy MNIST by default? Do we actually need to deploy something by default?

envs/vpn/docker/node/build_files/entrypoint.bash
```diff
@@ -43,7 +43,7 @@ su -l -c "export FBM_SECURITY_ALLOW_FEDERATED_ANALYTICS=\"${FBM_SECURITY_ALLOW_F
 su -l -c "echo \"$FBM_NODE_START_OPTIONS\" >/fbm-node/FBM_NODE_START_OPTIONS" $CONTAINER_USER
 
 # Launch node using node options
-$SETUSER fedbiomed node start $FBM_NODE_START_OPTIONS &
+su -l -c "fedbiomed node start $FBM_NODE_START_OPTIONS &" $CONTAINER_USER
```
sharkovsky, line 46: I've had issues with this exact line, thank you for fixing it.

notebooks/101_getting-started.ipynb
```diff
@@ -25,7 +25,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
```
sharkovsky, line 28: I don't think you intended to commit this notebook?

docs/user-guide/deployment/docker.md
```diff
@@ -0,0 +1,516 @@
...
+    --network fedbiomed-net \
+    fedbiomed/node
+```
+
+This setup allows the Node container to connect to the Researcher via its container name (`fbm-researcher`), thanks to Docker's internal DNS.
+
+
+It is up to the user to configure Docker networking according to their deployment scenario. For more information about Docker networking and available drivers, refer to the [Docker networking documentation](https://docs.docker.com/engine/network/drivers/).
+
+
```
sharkovsky, line 373: I would add a section here about connecting to the computers outside of the local network. 
My suggestion would be:
1) clarify that for sensitive data and secure deployments, the vpn approach is the recommended one
2) if the user knows what they're doing, provide instructions on how to connect two fedbiomed instances that are on public IPs on different networks (presumably only the researcher needs a public IP).

Also, I would suggest to update the network matrix documentation to include this scenario.

docker/node/entrypoint.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# Render supervisord config with actual user
+echo "Container user: ${FEDBIOMED_USER}"
```
sharkovsky, line 41: maybe also output the group information here, especially on Mac we've had issues with the group id sometimes (please do it for all container types).

sharkovsky:
Overall, this PR introduces some very nice and sorely-needed features such as automatic deployment to dockerhub and unprivileged containers. 

I'm a bit scared of the additional complexity of having some docker stuff in `docker` and other docker stuff in `envs/vpn/docker. In general, I'm a bit scared of having to maintain two versions of `Dockerfile` and `entrypoint.sh` that do somewhat similar things. As I mentioned, my ideal scenario would be that we can use the non-vpn images as base for the vpn ones, but I understand that this is maybe too complicated right now.

In the PR description, could you please explain what is `supervisord` and why you chose to use it? Also, how does the `pip install ... --find-links=/tmp/wheels` work and why did you chose to do it like that? 

I see you chose not to provide a `docker-compose.yml` file, but instead wrote extensive documentation. That's a valid choice in my opinion. 

If I understood correctly, these images do not require any privileges nor additional capabilities to run, right? Except for the node-gui, which (might? or always?) require root. In any case, that's great!

docker/node-gui/Dockerfile
```diff
@@ -0,0 +1,78 @@
+ARG FBM_IMAGE_VERSION
+
+# Start from Fed-BioMed base image
+FROM fedbiomed/node:${FBM_IMAGE_VERSION:-latest} AS fedbiomed-node
```
mvesin, line 4: Does the GUI really need to build on the node ? Isn't building on `base` enough ?

docker/base/entrypoint-base.sh
```diff
@@ -0,0 +1,167 @@
...
+    if [ "$uid" -gt 65535 ] || [ "$gid" -gt 65535 ]; then
+        log "ERROR: UID/GID must be <= 65535 (got UID=$uid, GID=$gid)"
+        return 1
+    fi
+    
+    return 0
+}
+
+# Only perform root operations if running as root
+if [ "$(id -u)" = "0" ]; then
```
mvesin, line 58: Do we have any scenario / use case for launching container not as root ?

docker/base/entrypoint-base.sh
```diff
@@ -0,0 +1,167 @@
+#!/bin/bash
+set -e
```
mvesin, line 2: Just a remark, because it is probably too late to address - time to merge:
The implementation add several nice features, but I suspect also the very "linear script in one file covering many case" implementation may make it harder to maintain.

docker/base/Dockerfile
```diff
@@ -0,0 +1,134 @@
+# Stage 1: Build Node.js frontend
```
mvesin, line 1: As Francesco, I would prefer to have it under `envs/docker` rather than `docker` to avoid mixing different path logics.

I also share the comment that we probably don't want to support 2 different ways of building images for a long time - an effort to reshape the `envs/vpn` images soon ?

docs/developer/docker-images.md
```diff
@@ -0,0 +1,84 @@
...
+```bash
+docker build \
+    --build-arg FEDBIOMED_USER=<user-name> \
+    --build-arg FEDBIOMED_UID=<user-id> \
+    --build-arg FEDBIOMED_GROUP=<group-name> \
+    --build-arg FEDBIOMED_GID=<group-id> \
+    -t fedbiomed/base:<tag> ./ -f docker/base/Dockerfile
+```
+
+Once the base image is built with a custom user, all other Fed-BioMed Docker images that use this base image will inherit that user configuration (except Node GPU). Therefore, if you change the user in the base image, make sure to rebuild the other images accordingly to ensure consistency across the stack. However, this is not the case for the Node GPU image, because it uses a multi-stage build where the base image is an NVIDIA-provided image. To rebuild a Node GPU image with a different user, you must redeclare the build arguments in the build command.
```
mvesin, line 38: Misc: Maybe remove reference to node-gpu as not supported yet ?

docs/user-guide/deployment/docker.md
```diff
@@ -0,0 +1,516 @@
...
+RUN pip install <a-package-required-for-your-setup>
+
+# Add your custom Docker instructions here, this will overwrite default entrypoint
+
+ENTRYPOINT ["/entrypoint.bash"] # Optional
+```
+
+Replace `<version-tag>` with the desired image tag (e.g., `latest`) and `<a-package-required-for-your-setup>` with the package(s) you need.
+
+## Torubleshooting
```
mvesin, line 508: Small typo => troubleshooting

docker/node/entrypoint.sh
```diff
@@ -0,0 +1,55 @@
+# !/bin/bash
+cat > /home/${FEDBIOMED_USER}/fbm.env <<EOF
+
+FBM_NODE_START_OPTIONS=${FBM_NODE_START_OPTIONS}
```
mvesin, line 4: Small important point: Here we probably need to update to match current practice, cf file from envs/vpn

FBM_SECURITY_ALLOW_DEFAULT_TRAINING_PLANS=${FBM_SECURITY_ALLOW_DEFAULT_TRAINING_PLANS:-False}

FBM_SECURITY_ALLOW_FEDERATED_ANALYTICS=${FBM_SECURITY_ALLOW_FEDERATED_ANALYTICS:-False}

FBM_SECURITY_ALLOW_PREPROC=${FBM_SECURITY_ALLOW_PREPROC:-False}

docs/user-guide/deployment/docker.md
```diff
@@ -0,0 +1,516 @@
...
+    fedbiomed/node
+```
+
+This setup allows the Node container to connect to the Researcher via its container name (`fbm-researcher`), thanks to Docker's internal DNS.
+
+
+It is up to the user to configure Docker networking according to their deployment scenario. For more information about Docker networking and available drivers, refer to the [Docker networking documentation](https://docs.docker.com/engine/network/drivers/).
+
+
+## Use Docker Compose to Run and Manage Multiple Fed-BioMed Instances
```
mvesin, line 374: Maybe a small section to explain the security model of these images:

- no VPN so expects to be run on a secure network or inside a custom VPN for sensitive deployments (as we currently use autosigned certificate for ciphering, ...)
- but parameters of FB set of operational deployement as for VPN/container mode (training plan approval activated by default, FA desactivated, ...)

mvesin (approved):
Great work @srcansiz !

I have a few comments directly in the code. I let you the last work to decide if they apply or not. OK for me to merge.

docker/node/entrypoint.sh
srcansiz: Hi @sharkovsky,
My intention was to give a ready to execute images to users where they can start containers and run 101 notebook directly without taking care of dataset deployment. It can also serve as test dataset for deployment.

docs/user-guide/deployment/docker.md
srcansiz: Hi @sharkovsky,

Thank you for the comment. Indeed, documentation needs more revising, I will keep in my mind while updating the docs.

srcansiz:
> Overall, this PR introduces some very nice and sorely-needed features such as automatic deployment to dockerhub and unprivileged containers.
> 
> I'm a bit scared of the additional complexity of having some docker stuff in `docker` and other docker stuff in `envs/vpn/docker. In general, I'm a bit scared of having to maintain two versions of `Dockerfile`and`entrypoint.sh` that do somewhat similar things. As I mentioned, my ideal scenario would be that we can use the non-vpn images as base for the vpn ones, but I understand that this is maybe too complicated right now.
> 
> In the PR description, could you please explain what is `supervisord` and why you chose to use it? Also, how does the `pip install ... --find-links=/tmp/wheels` work and why did you chose to do it like that?
> 
> I see you chose not to provide a `docker-compose.yml` file, but instead wrote extensive documentation. That's a valid choice in my opinion.
> 
> If I understood correctly, these images do not require any privileges nor additional capabilities to run, right? Except for the node-gui, which (might? or always?) require root. In any case, that's great!

Hi @sharkovsky, 

Thank you for the review! Relevant points and questions!

- Regarding `docker` and `envs/docker` I fully agree that we should not maintain two different docker. As soon as this branch is merged, I will focus on using these docker images in VPN images. I didn't want to do it in this branch  to avoid complicating things. 

- `supervisord` is process and service manager. Instead of running node in the command line as background process it allows to create a service that can be manageable through `supervisorctl` interface. It handles writing error and infor logs in files, auto-start etc. It is not an obligation but I find it convient to avoid handling process that managed manually through PIDs. 

- `--find-links` is some kind of optimization to reduce the time of installing Fed-BioMed dependencies. What used to happen before is that each image used to install base `fedbiomed` dependencies no matter if you specify gui, researcher, node. Now, `fedbiomed-base` resolves dependencies and downloads every image and puts them in the `/tmp/wheels` directory, and other images uses this directory to install Fed-BioMed. Thanks to this improvement building base image takes 4-5min max, and other images are build in 1min approx, but before every image build operation used to take around 4-5min. 

- Yes, I wanted to leave `docker-compose` configuration to end user by guiding them in the docs. However, VPN can have a `docker-compose.yml` because it is more like an application that requires multiple instances to be launched synchronously  to be functional.

- I had to launch `node-gui` as root because Nginx requires to be root. Ideal solution would be to run `node-gui` non-root and launch Nginx in a separate container and use reverse proxy but it didn't want to go that far. 

Thanks for the review again.

docker/node/entrypoint.sh
sharkovsky: +1

sharkovsky:
> > Overall, this PR introduces some very nice and sorely-needed features such as automatic deployment to dockerhub and unprivileged containers.
> > I'm a bit scared of the additional complexity of having some docker stuff in `docker` and other docker stuff in `envs/vpn/docker. In general, I'm a bit scared of having to maintain two versions of `Dockerfile`and`entrypoint.sh` that do somewhat similar things. As I mentioned, my ideal scenario would be that we can use the non-vpn images as base for the vpn ones, but I understand that this is maybe too complicated right now.
> > In the PR description, could you please explain what is `supervisord` and why you chose to use it? Also, how does the `pip install ... --find-links=/tmp/wheels` work and why did you chose to do it like that?
> > I see you chose not to provide a `docker-compose.yml` file, but instead wrote extensive documentation. That's a valid choice in my opinion.
> > If I understood correctly, these images do not require any privileges nor additional capabilities to run, right? Except for the node-gui, which (might? or always?) require root. In any case, that's great!
> 
> Hi @sharkovsky,
> 
> Thank you for the review! Relevant points and questions!
> 
>     * Regarding `docker` and `envs/docker` I fully agree that we should not maintain two different docker. As soon as this branch is merged, I will focus on using these docker images in VPN images. I didn't want to do it in this branch  to avoid complicating things.
> 
>     * `supervisord` is process and service manager. Instead of running node in the command line as background process it allows to create a service that can be manageable through `supervisorctl` interface. It handles writing error and infor logs in files, auto-start etc. It is not an obligation but I find it convient to avoid handling process that managed manually through PIDs.
> 
>     * `--find-links` is some kind of optimization to reduce the time of installing Fed-BioMed dependencies. What used to happen before is that each image used to install base `fedbiomed` dependencies no matter if you specify gui, researcher, node. Now, `fedbiomed-base` resolves dependencies and downloads every image and puts them in the `/tmp/wheels` directory, and other images uses this directory to install Fed-BioMed. Thanks to this improvement building base image takes 4-5min max, and other images are build in 1min approx, but before every image build operation used to take around 4-5min.
> 
>     * Yes, I wanted to leave `docker-compose` configuration to end user by guiding them in the docs. However, VPN can have a `docker-compose.yml` because it is more like an application that requires multiple instances to be launched synchronously  to be functional.
> 
>     * I had to launch `node-gui` as root because Nginx requires to be root. Ideal solution would be to run `node-gui` non-root and launch Nginx in a separate container and use reverse proxy but it didn't want to go that far.
> 
> 
> Thanks for the review again.

Thanks for the explanation. 

I didn't know about `supervisord` but it is a good choice IMO!

Also, nice optimization about putting all fedbiomed dependencies in the base image and then installing only the needed ones for each image.

let me know when it's ready for second review

docker/node-gui/Dockerfile
srcansiz: Hi @mvesin, 

In the feature `node` can additional packages that may be required in the GUI too. It is the reason I wanted to derive it from `node` image.

docs/developer/docker-images.md
srcansiz: Hi @mvesin, 

Thanks, yes I forgot to remove that. I can also add GPU supported image, either in this PR or in an other PR.

docker/node/entrypoint.sh
srcansiz: Good point ! I will update it

docs/user-guide/deployment/docker.md
srcansiz: Thanks @mvesin, I will update doc accordingly.

srcansiz:
Thank you @mvesin @sharkovsky for the review! I will address the suggested changes and push as soon as possible and let you know for a second look.

mvesin:
Thank you Sergen, if you want to complete quickly, feel free to skip my second review - as you  prefer

docker/base/Dockerfile
leonard2103: It seems like having a `/docker` folder is considered best practice. Taking into account the amount of Dockerfiles it seems the right approach to me.

docs/developer/docker-images.md
leonard2103: @srcansiz I think this still needs to be addressed

fedbiomed/common/logger.py
```diff
@@ -48,6 +48,7 @@
 import logging
 import os
 import socket
+import sys
```
leonard2103: This line seems unused.

hatch_build.py
```diff
@@ -24,12 +25,18 @@ def initialize(self, version, build_data):
...
-        finally:
-            os.chdir("../../")
+        for attempt in range(3):
+            try:
+                logger.info(
+                    "### Yarn: Installation front-end dependencies to prepare build.\n"
+                )
+                subprocess.run([yarn, "install"], check=True)
+                logger.info("\n### Yarn: Building front-end application run.\n")
+                subprocess.run([yarn, "build"], check=True)
```
leonard2103, line 35: It seems like you are missing a `break`. In case it is successful.

docker/researcher/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+
+USER root:root
+# Install socat
+RUN apt-get update && apt-get install -y socat
+
+# Create /fbm-researcher directory with proper ownership
+RUN mkdir -p /fbm-researcher && \
+    chown -R ${FEDBIOMED_USER}:${FEDBIOMED_GROUP} /fbm-researcher
+
+USER root:root
```
leonard2103: Line repeated L26.

.github/workflows/docker-deploy.yml
```diff
@@ -0,0 +1,66 @@
...
+name: Tests and Publish Fed-BioMed Docker Images
+description: |
+  Builds for testing and deployment of Fed-BioMed docker images. Deployment is done if
+  a new tag is pushed, or a new commit is pushed to develop. It is to keep "dev" version
+  pushed in docker hub. 
+
+on:
+  push:
+    branches:
+      - 'feature/1335-*'
```
leonard2103: This line should be remove before merging.

docker/base/entrypoint-base.sh
srcansiz: To be able to change user ownership  container has to be started as root.

docs/developer/docker-images.md
srcansiz: Done, thanks for reminding!

hatch_build.py
srcansiz: Good catch! thanks

leonard2103 (commented):
You can check my comments and merge after you fix where necessary. @srcansiz, good work.

envs/vpn/docker/node/build_files/entrypoint.bash
mvesin: Nota: introduces a subtle bug.
As the execution directory is changed with "su -l -c" then we need a `fedbiomed node -p /fbm-node`

PR #1384, opened 2025-07-31
Add codespell support (config, workflow to detect/not fix) and make it fix some typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.


mvesin (approved):
Dear @yarikoptic
Thank you very much for this great contribution !
It corrected so many mispellings, and also found 3 bugs.

We will merge it as soon as we can pass the CI test, currently delayed by some Codecov issues.

Unless you decline it, we will add you to AUTHORS.md to credit your contribution.

Cheers !

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,25 @@
+# Codespell configuration is within pyproject.toml
+---
+name: Codespell
+
+on:
+  push:
```
mvesin, line 6: One question out of curiosity. Is there any constraining reason why you applied codespell on push/PR for develop ?

Or is kind of generic/common configuration as I understand ?
In that case I will slightly adapt after merging (add `master` branch, remove `push` as we don't push directly to these branches)

yarikoptic: My script just did it for the default branch. Your are most welcome to extend to others. Testing branch, if no direct pushes, indeed shouldn't result in different results from what in PRs, bit also shouldn't hurt I guess

Issue #1444, opened 2025-09-26
GUI Bugs when opening FileBrowser when adding a dataset


alitolga:
As a user, I have noticed two bugs when I try to open the File Browser to add a dataset in the GUI. The first one can be seen in the screenshot below.

<img width="1704" height="803" alt="Image" src="https://github.com/user-attachments/assets/02630610-1668-42d5-9d34-9f98237c3bed" />

Where the files and folders were not visible. This was due to path being not properly transmitted to the frontend and remaining as null. 

When I wrote a workaround by manually setting the path as root ('/'), I have encountered another bug, where the FileBrowser would crash when trying to traverse "/mnt/c". After some debugging, I have discovered that the `file_stats` and `get_disk_usage` functions inside `"fedbiomed_gui/server/utils"` were trying to calculate the folder size by traversing all of the subfiles and subfolders in that folder; which is highly inefficient and unrobust.

I can confirm that after removing the call to those functions, the issue is fixed, and I can display the FileBrowser widget and add a dataset without a problem.

For the issue with the path, which is possibly caused by the environment variables not working properly on Windows and WSL, I will create another issue

PR #1445, opened 2025-09-26
Fix issue that caused the GUI server to crash when adding a new dataset

alitolga:
It was caused by the function call `file_stats `in `repository.py` which caused the traversal of all subfiles when calculating file size.

Details can be seen in [Issue #1444](https://github.com/fedbiomed/fedbiomed/issues/1444)

### DIFF
--- a/fedbiomed_gui/server/routes/repository.py
+++ b/fedbiomed_gui/server/routes/repository.py
@@ -3,11 +3,11 @@ import re
 
 from flask import request
 
-from .api import api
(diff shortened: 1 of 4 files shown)


srcansiz:
Hi @alitolga, 

Do you want to implement a version of GUI where the calculation is only done for the files but not for directories?

alitolga:
Hi @srcansiz ,

Yes, I can work on that :)

PR #1509, opened 2025-11-21
Add Unit Tests for SkLearnDataLoader. Increase coverage on Dataset and Tabular Dataset unit tests.

alitolga:


tests/test_sklearn_dataloader.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def test_dataset_getitem_exception(self):
+        failing_dataset = FailingGetItemDataset()
+        loader = SkLearnDataLoader(
+            failing_dataset, batch_size=1, shuffle=False, drop_last=False
+        )
+
+        it = iter(loader)
+        with self.assertRaises(FedbiomedError):
+            next(it)
```
srcansiz: Hi @alitolga,

We decided to use pytest way of writing test for the new tests that will be written. Could you please convert the tests to pytest?

PR #1537, opened 2025-12-10
Feature/tabular analytics

alitolga:
First implementation for the Tabular Analytics.

Feel free to leave comments/remarks.


tests/test_tabular_analytics.py
```diff
@@ -0,0 +1,234 @@
...
+    mock_controller = Mock(spec=TabularController)
+    mock_controller._reader = mock_reader
+
+    mock_dataset = Mock(spec=TabularDataset)
+    mock_dataset._controller = mock_controller
+
+    return mock_dataset
+
+
+class TestTabularAnalyticsMean:
```
srcansiz: @alitolga, I think you don't need t use test class here. The functions contains `self` but not used. Please use functions instead of classes

fedbiomed/common/analytics/analytics_dispatcher.py
```diff
@@ -0,0 +1,17 @@
...
+    registry = {ImageFolderDataset: ImageAnalytics, TabularDataset: TabularAnalytics}
+
+    @classmethod
+    def register(self, dataset_cls, analytics_cls):
+        self.registry[dataset_cls] = analytics_cls
+
+    @classmethod
+    def mean(self, dataset, params=None):
+        analytics_cls = self.registry[type(dataset)]
+        return analytics_cls().mean(dataset, params)
```
srcansiz: Hi @alitolga,

Thank you, very nice idea to implement `Dispatcher`!

srcansiz: Do you think that this class can be implemented as `Mixin` class. For example, this implementation assumes that the dataset is going to be given as argument. 
```
my_dataset = TabularDataset(...)

# I expect the usage will be
mean = AnalyticsDispatcher.mean(my_dataset, params={}) 

```

But what about making Analytics classes as mixin classes where it can be integrated directly in the dataset

```
class TabularDataset(.., _TabularAnalytics)
....

# Then, the usage will be 

my_dataset = TabularDataset(...)
my_dataset.mean(params)
```

fedbiomed/common/analytics/tabular_analytics.py
```diff
@@ -0,0 +1,48 @@
+from fedbiomed.common.dataset import TabularDataset
+from fedbiomed.common.dataset_controller import TabularController
+from fedbiomed.common.dataset_reader import CsvReader
+from fedbiomed.common.exceptions import FedbiomedError
+
+from .analytics_strategy import AnalyticsStrategy
+
+
+class TabularAnalytics(AnalyticsStrategy):
+    def mean(self, dataset: TabularDataset, params=None):
```
srcansiz: Can you please add return type.

srcansiz: And `type` for `params` argument

fedbiomed/common/analytics/tabular_analytics.py
```diff
@@ -0,0 +1,48 @@
...
+            raise FedbiomedError(
+                f"Expected TabularController, got {type(controller).__name__} instead."
+            )
+
+        # Initialize reader
+        reader = controller._reader
+        if not isinstance(reader, CsvReader):
+            raise FedbiomedError(
+                f"Expected CsvReader, got {type(reader).__name__} instead."
+            )
```
srcansiz: No need to check reader type it is already checked implicitly by checking the dataset type.

fedbiomed/common/analytics/tabular_analytics.py
```diff
@@ -0,0 +1,48 @@
...
+class TabularAnalytics(AnalyticsStrategy):
+    def mean(self, dataset: TabularDataset, params=None):
+        # Read data from dataset controller
+
+        # Initialize controller
+        controller = dataset._controller
+        if not isinstance(controller, TabularController):
+            raise FedbiomedError(
+                f"Expected TabularController, got {type(controller).__name__} instead."
+            )
```
srcansiz: Isn't possible to use the `dataset` directly. Here, you are accessing a private member `_controller` which is not an expected way of accessing class member.

fedbiomed/common/analytics/tabular_analytics.py
```diff
@@ -0,0 +1,48 @@
...
+
+        # Determine columns
+        if params is None or params.get("cols") is None:
+            cols = [c for c, t in df.schema.items() if t.is_numeric()]
+        else:
+            cols = params["cols"]
+
+        missing = [c for c in cols if c not in df.columns]
+        if missing:
+            raise KeyError(f"Columns not found: {missing}")
```
srcansiz: Isn't this logic already implemented in `TabularDataset`, like making sure that column required is existing or not , selecting a specific column etc.

fedbiomed/common/analytics/tabular_analytics.py
```diff
@@ -0,0 +1,48 @@
...
+
+        # Select only numeric columns
+        numeric_cols = [c for c in cols if df.schema[c].is_numeric()]
+        if not numeric_cols:
+            return {}
+
+        result = df.select(numeric_cols).mean()
```
srcansiz: I prefer to use `TabularDataset` directly rather than accessing `df` from controller.

fedbiomed/common/analytics/analytics_dispatcher.py
alitolga: I changed the implementation according to the mixin paradigm. I wait for your feedback :)

fedbiomed/common/analytics/analytics_strategy.py
```diff
@@ -0,0 +1,32 @@
...
+            count += 1
+
+        if count == 0:
+            return {"data": None, "target": None}
+
+        # Calculate means by dividing by count
+        data_mean = data_sum / count
+        target_mean = target_sum / count
+
+        return {"data": data_mean, "target": target_mean}
```
srcansiz: Hi @alitolga,

I think for the first version it could be nice to have only `TabularAnalytics` without base class. It is because we don't know which parts can be shared among datasets and analytics. Therefore, it is logic to have simple implementation first and improve it as we add other analytics.

srcansiz: Another thing, I think that `target` does not make sense for analytics. Can you please check if it is possible to create a dataset `TabularDataset` without specifying `target` argument, if not propose a version where it is possible. This will allow to use only `data` for analytics.

alitolga: Hi @srcansiz ,

I applied the suggested changes.

alitolga: Hi @srcansiz ,

The analytics return a dictionary with the column names now.

fedbiomed/common/analytics/tabular_analytics.py
```diff
@@ -0,0 +1,60 @@
...
+        for i, col in enumerate(self._input_columns):
+            result[col] = (
+                data_mean[i] if hasattr(data_mean, "__getitem__") else data_mean
+            )
+
+        # Add target column means
+        for i, col in enumerate(self._target_columns):
+            result[col] = (
+                target_mean[i] if hasattr(target_mean, "__getitem__") else target_mean
+            )
```
srcansiz: Hi @alitolga, 

How do you handle the case where there is no target columns provided? Given a target for `TabularDataset` in analytics does not make sense no? It may require to to modify `TabularDataset` implementations to provide columns only without separating `inputs` and `targets`.

srcansiz: or using only `inputs` and not forcing `target` to be provided.

alitolga: Hi @srcansiz ,

I have done the necessary modifications.

fedbiomed/common/analytics/tabular_analytics.py
```diff
@@ -0,0 +1,65 @@
...
+        Iterates through dataset items (tuples of data and target tensors/numpy arrays)
+        and computes the mean of both data and target features manually.
+
+        Returns:
+            Dictionary keyed by column names (input_columns + target_columns) with their
+            respective means. Preserves the input data type (Tensor or numpy array).
+        """
+        data_sum = 0
+        target_sum = 0
+        count = 0
```
srcansiz: Hi @alitolga, 

The implementation is lacking validation of column types. Single this classes intended to be used with `TabularDataset`, I think you can acces `self._controller` to validate if given columns are numeric not string.

fedbiomed/common/analytics/tabular_analytics.py
```diff
@@ -0,0 +1,65 @@
...
+
+        if count == 0:
+            # Return dict with None values for all columns
+            result = {}
+            for col in self._input_columns:
+                result[col] = None
+            if self._target_columns is not None:
+                for col in self._target_columns:
+                    result[col] = None
+            return result
```
srcansiz: I think you can return directly raise an error if there is no sample. You can do it in the very beginning of the method.
```
if not len(self):
    raise FedbiomedError("....")
```

fedbiomed/common/analytics/tabular_analytics.py
```diff
@@ -0,0 +1,65 @@
...
+            respective means. Preserves the input data type (Tensor or numpy array).
+        """
+        data_sum = 0
+        target_sum = 0
+        count = 0
+
+        for idx in range(len(self)):
+            data, target = self[idx]
+            data_sum += data
+            count += 1
```
srcansiz: No need to have `count`. It is equal to `len(self)`

fedbiomed/common/analytics/_tabular_analytics.py
```diff
@@ -0,0 +1,65 @@
...
+        data_mean = data_sum / count
+        target_mean = target_sum / count if target_sum != 0 else None
+
+        # Build result dict keyed by column names
+        result = {}
+
+        # Add input column means
+        for i, col in enumerate(self._input_columns):
+            result[col] = (
+                data_mean[i] if hasattr(data_mean, "__getitem__") else data_mean
```
srcansiz, line 71: Is `hasattr` check really needed here ?

fedbiomed/common/analytics/tabular_analytics.py
alitolga: Hi @srcansiz ,

I added the function to filter the non-numeric values using the schema.

fedbiomed/common/analytics/_tabular_analytics.py
alitolga: It is just to be error-proof in case there is only one input column and data_mean returns a scalar.

Issue #1548, opened 2025-12-19
Design & implement `PreprocJob` “class skeleton” on the node side


mvesin:
Design & implement `PreprocJob` “class skeleton” on the node side (similar to `Round`)

- Empty - no Fed-ComBat payload (+ no “harmonized/intermediate dataset save/load”, no “harmonization process state save/load”, ...)
- Add message handler for `PreprocRequest` in `Node.py` to handle harmonization requests and reply. 
- Probably implement as `PreprocJob` (similar to `Round`) that takes `PreprocRequest`, does nothing, and returns `PreprocReply`. 
- Repeat for each step defined in harmonization protocol exchanges

PR #1552, opened 2025-12-19
Refactor node side FAJob payload class

mvesin:
**PR description**

~~IMPORTANT: MERGE AFTER #1547~~ now merged

Refactor `FAJob`:
* split between `FAJob` and `_BaseJob` to be able to share some code with other jobs plus expose common job pattern
* move to fedbiomed.node.jobs

Preparation task for #1548

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:


mvesin:
@alitolga  as you pushed to this branch I suppose you reviewed it and I can merge ?

fedbiomed/node/jobs/_base_job.py
```diff
@@ -0,0 +1,154 @@
...
+    @abstractmethod
+    def _build_args_for_dataset(self, dataset_entry: dict) -> dict:
+        """Build arguments for dataset initialization.
+
+        Args:
+            dataset_entry: dataset entry from dataset manager
+
+        Returns:
+            Dict of arguments for dataset initialization
+        """
```
srcansiz, line 143: Hi @mvesin, 

Instead of creating a method for extracting dataset arguments. These arguments can be passed as argument in `_build_dataset`, and child classes can handle their own logic of extracting these parameters and providing it to `_build_dataset`. For example,  dataset arguments can be extracted in the `run` method for now.  So, I propose to remove `_build_args_for_dataset` method from the base class.

mvesin: Thank you @srcansiz for the review.

The thing is I don't see how I can have this working, because at the time we call `FAJob.run()` we didn't read yet the `dataset_entry` from the database. 

Am I missing something ?
Maybe we can discuss what you have in mind ?

srcansiz: Ahh,  ok.  sorry  I missed it.

fedbiomed/node/jobs/_base_job.py
```diff
@@ -0,0 +1,154 @@
...
+            data_types: list of accepted data types for the dataset
+
+        Raises:
+            _InternalJobError: if dataset cannot be recovered or initialized.
+        """
+        if not self._dataset_id:
+            raise _InternalJobError(
+                f"This job {self.__class__.__name__} has no associated `dataset_id` "
+                f"in message {self._message}."
+            )
```
srcansiz: `dataset_id` is mandatory argument that is already checked when the message class `FARequest` is instantiated again on the node side. I know that it was already there before this PR but if you want you can correct it.

fedbiomed/node/jobs/_base_job.py
```diff
@@ -0,0 +1,154 @@
...
+
+        # dataset manager to get metadata about dataset, dlp and loading block
+        dataset_manager = DatasetManager(self._db_path)
```
srcansiz: I think there is no need to instantiate a dataset manager again. Node already have it. Node can pass it directly to the node.

fedbiomed/node/jobs/_base_job.py
mvesin: Yes, fully true for `FAJob`

But `_build_dataset` is now a method of `_BaseJob`, so we probably cannot assume that every job class will (have and) set a `dataset_id`. So IMO it makes sense to keep it.

Am I missing something ?

fedbiomed/node/jobs/_base_job.py
mvesin: Yes, good idea, I will apply this change.

FYI did not touch it yet, as it was pre-existing in `FAJob` implementation

Question: was it assessed whether `DatasetManager` is thread-safe ? I suppose not (yet the underlying `TinyDBConnector` is a singleton to ensure low-level coherence. I don't think the suggested modification will change/break something regarding that point, but I'm not 100% sure.

fedbiomed/node/jobs/_base_job.py
srcansiz: Yes, that's true. Then, maybe we can instantiate `dataset_id` argument only in child classes not in the `_BaseJob`.

mvesin: Yes, it makes sense :-p => pushed a fix for that + also adapted unit tests for the review changes

@srcansiz can you check if you have other review comments before merging ? Thx !

srcansiz: Sorry, I got confused again :) we removed dataset_id from `_BaseJob`  but there is still a method that is called `_build_dataset` that expects `_dataset_id` present. So, the implementation become a bit confusing. I think better to let `dataset_id` in the `_BaseJob` without checking if it is exist or not assuming that that check is already done for `FARequest`. 

Then, we can think  about other solutions when there is a job that does not requires a dataset. In the end, you created this parent and child separation based on what will be executed for federated analytics and federated preprocessing. So let's assume that we will use dataset for now for every job.

Sorry for misleading you for passing dataset id to the child class.

mvesin: OK, no problem, I implemented that (revert to previous version + remove test for existence of `self_dataset_id`

Will proceed with merge now :-)

PR #1555, opened 2025-12-22
Invert logic for transforms and conversion on Tabular Dataset. Fix bu…

alitolga:
…g on Tabular Analytics when input_columns on Tabular Dataset is None or []

Suggested fix for handling Strings on Tabular Data. Also minor fix on Tabular Dataset logic when input_columns is None or Empty List.


alitolga:
This PR is also obsolete, as the Analytics api will likely change with #1595 .

**However,** the bug in case of empty columns might be remaining, so we might need to push a fix for the change on csv_reader and the __init__ function when input columns are empty.

PR #1558, opened 2025-12-30
PreprocJob class skeleton on node side

mvesin:
**PR description**

Continues and completes #1548 

Builds on node-side part of `poc/1550-and-1549-skeleton-implementation` as mentionned with co-author in commit

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/fedbiomed/node/jobs/_preproc_job.py
+++ b/fedbiomed/node/jobs/_preproc_job.py
@@ -1,0 +1,105 @@
+# This file is originally part of Fed-BioMed
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Implementation of Preprocess Job class of the node component
+"""
+
+from fedbiomed.common.constants import ErrorNumbers, PreprocStep, PreprocType
+from fedbiomed.common.message import ErrorMessage, PreprocReply, PreprocRequest
+from fedbiomed.node.dataset_manager import DatasetManager
+
+from ._base_job import _BaseJob
+
+
+class PreprocJob(_BaseJob):
+    """
+    This class represents the preprocessing part executed by a node in a given preprocessing job.
+    """
+
+    def __init__(
+        self,
+        root_dir: str,
+        dataset_manager: DatasetManager,
+        node_id: str,
+        node_name: str,
+        request: PreprocRequest,
+    ) -> None:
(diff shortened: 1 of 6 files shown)


fedbiomed/node/jobs/_preproc_job.py
```diff
@@ -0,0 +1,105 @@
...
+            PreprocReply message if successful, ErrorMessage otherwise.
+        """
+        # Further check message parameters if needed
+        try:
+            self._preproc_type = PreprocType(self._preproc_type_raw)
+        except ValueError:
+            return self._build_error_msg(
+                f"Received invalid preproc_type: {self._preproc_type_raw}",
+                errnum=ErrorNumbers.FB326.value,
+            )
```
srcansiz, line 62: Hi @mvesin,

How do you plan to execute different preprocessing types. As I understand, `PreprocJob` is a generic class that handles different type of preprocessing requests, or it is only for `FedCombat`. 

I am asking that to understand if variables such as `preproc_step` will be useful for other preprocesses, or they are more belong to `FedCombat`.  

It may be too early to ask this question but in the future as FedCombat implementation matured we may need to reconsider notion of `PreprocJob` to identify what belongs generic preprocessing and what belongs specific pre process jobs. 

More precisely, I would like to avoid having `PrerpocJob` that will end of like `Round` class that has very complex execution steps and has many responsibilities.

mvesin: Hi @srcansiz 

Yes, this is what you say.

The idea is that `PreprocJob` is a generic class, this is why it needs the `preproc_type` and `preproc_step` to instantiate the proper classes for each case. It could be per type (for now: one class for FedCombat) or per type+step (for now: one class per FedCombat step 2/4/5/6).

This can be implemented in practice with many class designs (eg subclass, factory, ...), and as you say I felt like it is to be decided in later step, by the people doing the implementation (as we discussed: don't try design everything in advance). So it will be reconsidered while advancing in the implementation.

And yes, of course, we don't want another `Round` class :-p

srcansiz:
Hi @mvesin, 

Thank you! The PR is okay for me. You can proceed to merge.

PR #1588, opened 2026-01-16
Feature/1564 image analytics

alitolga:
FIrst implementation for image analytics. This implementation is not finished yet.

 Missing:

- [ ] Integration with aggregation functions for histogram and quantile
- [ ] Unit tests
- [ ] Potentially other analytics such as Contrast, Signal to Noise Ratio, Mean Image...


alitolga:
I believe this PR is now obsolete, and will be followed by #1595

Issue #1569, opened 2026-01-05
Test and adapt using an “Experiment within Experiment” for training the harmonization model


mvesin:
Test and adapt using an “Experiment within Experiment” for training the harmonization model

**Step1**

- [x] basic run "skeleton class" for Fed-ComBat with training a *dummy* model within the pre-processing

**Step2** Misc/advanced pre-processing nested model support for dummy model

- [x] check handling of saved experiment data & experiment folder
    * use `/var/experiments/Experiment_0123_fedcombat` for `Experiment_0123`
    * using sub-directory of `/var/experiments/Experiment_0123` would require more complicated adaptations
- [x] check handling of breakpoints (for main experiment, no need to support for pre-processing)
- [x] check handling of monitor (scalar) messages & adaptable verbosity for pre-processing (and main experiment)
- [x] check handling of tensorboard (for main experiment, no need to support for pre-processing)
    * tensorboard works fine for main experiment; disabled (don't try to support) for nested experiment
- [x] check using non default `preproc_args` (eg for `training_args` of nested experiment)
- [x] verify format and values of covariates/phenotypes (empty covariates/phenotypes, duplicates phenotypes/covariates)
- [x] check behaviour with numeric/string/mixed covariates/phenotypes
    * scenarios: all values same format ; mixed int/str values in source/target/source+target; duplicate values; duplicate between source-target
    * solution: improve checks in `CsvReader` + more restrictive checks in `FedCombatPreproc`
- [x] check handling of filtered nodes
- [x] check handling of errors cases (node side, researcher side)
    * misc: when running nested exp, 2 "terminated due to an exception" messages, 1 for main exp + 1 for nested pre-processing exp
- [x] check with a sklearn notebook
- [x] check with 2 & 3 nodes

**Step3** Implement for real use case and model

 - [ ]  support with containers (model approval)

- [ ] refactor to avoid import loops (and local imports)

- [x] run "skeleton class" for Fed-ComBat with training the *Fed-ComBat model* within the pre-processing

- [x] refactor to match with Fed-ComBat class design (step 2-4-5-6) from #1568

PR #1597, opened 2026-01-22
Experiment in experiment with basic model

mvesin:
**PR description**

- partially addresses #1569 : basic model training in preprocessing (== Experiment in experiment with basic model)

Hint: can use the 2 "TEMPO" notebooks for testing the PR.

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:


fedbiomed/common/utils/_utils.py
```diff
@@ -53,27 +52,29 @@ def get_class_source(cls: Callable) -> str:
...
-
-    if status:
-        file = get_ipython_class_file(cls)
-        codes = "".join(inspect.linecache.getlines(file))
-
-        # Import only on IPython interface
-        module = importlib.import_module("IPython.core.magics.code")
-        extract_symbols = module.extract_symbols
-        print(extract_symbols)
-        class_code = extract_symbols(codes, cls.__name__)[0][0]
```
srcansiz, line 73: Hi @mvesin,

I am not sure it is a good idea to remove `status.is_ipython` get source can fail for different reasons not only because it is executed in IPython shell. The reason to have `is_python` is precisely detect if the code is execute in ipython shell. I would recommend to keep it.

fedbiomed/researcher/federated_workflows/preproc/_fedcombat.py
```diff
@@ -100,6 +101,14 @@ def _needs_harmonization(self) -> bool:
...
         return False
 
+    def set_nodes(self, nodes: list[str]) -> None:
+        """Set the list of nodes participating in the harmonization.
+
+        Args:
+            nodes: List of node IDs.
+        """
+        self._nodes = nodes
+
```
srcansiz, line 111: Hi @mvesin,

I don't know if you were obliged to implement it but I think it is better to use `_fds` to get the nodes to be consistent with the nodes that are searched and selected in the experiment (parent experiment). Setting nodes here independent from initial `_fds` can be risky.

fedbiomed/researcher/federated_workflows/preproc/_fedcombat_step3.py
```diff
@@ -0,0 +1,221 @@
...
+        return DataManager(dataset=dataset, **train_kwargs)
+
+    def training_step(self, data, target):
+        output = self.model().forward(data).float()
+        loss = nn.MSELoss()(output, target)
+        return loss
+
+
+# TODO: change design to match other preprocessing steps from issue 1568
+class _FedCombat_Step3:
```
srcansiz: Hi @mvesin,

I could be nice to use explicit name of the preprocessing that is going to be applied rather than `step1,2,3`. e.g `FedCombatHarmonizationModel`. This is much more clear for developers and users.

fedbiomed/researcher/federated_workflows/_federated_workflow.py
```diff
@@ -591,6 +595,11 @@ def set_nodes(self, nodes: Union[List[str], None]) -> Union[List[str], None]:
             )
             logger.critical(msg)
             raise FedbiomedTypeError(msg)
```
srcansiz, line 597: Hi @mvesin,

I think this part causing the problem and making it required to create `set_nodes` method for `FederatedProproc` class. It seems to it is time to an improvement to manage nodes that are participating to experiment from single point. For example, these filters can be handled from `_fds`, and other objects/classes uses `_fds` to access up to date list of nodes with filter or without filter.

srcansiz (commented):
Hi @mvesin,

Thank you for the PR! Very nice work. I have a few minor comments below.

fedbiomed/common/utils/_utils.py
mvesin: After discussion: we need to keep `inspect.getsource()` as first test, for the case of experiment in experiment (`is_ipython()` is True but TP needs to be picked from file).

Yet, we can test `is_ipython()` in second to be more explicit on errors when reading file fails and we are not running in iPython (script).

fedbiomed/researcher/federated_workflows/_federated_workflow.py
mvesin: Exactly !
I fully agree with you @srcansiz about the analysis and proposed solution.
I reached same conclusion while coding the branch, but just considered that further refactoring of `_fds` is beyond the current PR.

fedbiomed/researcher/federated_workflows/preproc/_fedcombat.py
mvesin: As mentioned in other comment: it is needed for now, but will be removed after further refactor of `_fds` (beyond the scope of the current PR)

fedbiomed/researcher/federated_workflows/preproc/_fedcombat_step3.py
mvesin: Thank you @srcansiz I will change the name as suggested
- `_FedCombat_Step3` to `_FedCombatTrainModel`
- etc.

PR #1607, opened 2026-02-05
Initial draft for the security logger and an example to log training round information.

alitolga:


fedbiomed/common/logger.py
```diff
@@ -207,8 +237,122 @@ def __init__(self, level: str = DEFAULT_LOG_LEVEL):
...
+        """
+        current = dict(
+            SECURITY_CONTEXT.get() or {}
+        )  # copy to avoid mutation across calls
+        current.update({k: v for k, v in ctx.items() if v is not None})
+        token = SECURITY_CONTEXT.set(current)
+        try:
+            yield
+        finally:
+            SECURITY_CONTEXT.reset(token)
```
srcansiz, line 344: Hi @alitolga, 

Better to check if the context is thread safe. For example, What happens if node receives search request while there is a training running. It is going to change the context of the logger that is used in the training? I think it could be nice to test: 

1 - Run a experiment training and add security log in round 
2 - Create another experiment in parallel and execute search request. 
3 - Check if the experiment id in these two different security has changed after the second experiment search executed.  The second one should not override the experiment id in the first one.

alitolga: Hi @srcansiz ,

Just to give you an initial update, I tested the scenario you specified and none of the fields seem to get mixed! 

The search request apparently doesn't have experiment id. However, I tested with 2 researchers and 1 node, and the researcher ids, operations, status, request ids, all seem to be logging perfectly fine.

We can discuss more scenarios to test as well.

fedbiomed/node/node.py
```diff
@@ -118,6 +120,26 @@ def __init__(
             tp_approval=self._config.getbool("security", "training_plan_approval"),
         )
 
+        # Initialize security audit logging
+        logger.configure_security(
+            node_id=self._node_id,
+            node_name=self._node_name,
+            fedbiomed_version=__version__,
+        )
+        security_log_dir = os.path.join(self._config.root, "log")
```
srcansiz: Hi @alitolga,

I think you can implement this logic in `logger` itself. For example, `logger.set_security_logs(self._config.root)`takes config  as argument or `self._config.root`and creates security log directory and file.

fedbiomed/node/node.py
```diff
@@ -427,73 +495,142 @@ def task_manager(self):
...
-                                ),
-                                secagg_active=self._config.getbool(
-                                    "security", "secure_aggregation"
+
+            # Set security context for all logs in this task
+            with logger.security_context(
+                researcher_id=item.researcher_id,
+                experiment_id=item.experiment_id,
+                request_id=item.request_id,
+            ):
```
srcansiz, line 502: I am not sure if you already implemented but we may also need a API that can be used without security context is set where developer gives these context variables one by one  as `security_event` arguments. This can be useful for `async` operation.  

And please remember to test this context with async operations.

fedbiomed/common/logger.py
```diff
@@ -207,8 +291,152 @@ def __init__(self, level: str = DEFAULT_LOG_LEVEL):
...
+    def configure_security(
+        self,
+        *,
+        node_id: Optional[str] = None,
+        node_name: Optional[str] = None,
+        fedbiomed_version: Optional[str] = None,
+    ) -> None:
+        """Configure default fields that must appear in every security log entry."""
+        if node_id is not None:
+            self._security_defaults["node_id"] = node_id
```
srcansiz: This is very specific to `node` component. I think you can make this function more generic that it can handles node and researcher specific defaults. For example, using `component_id` instead of `node_id` and also `component_name`. Then, you can add `component_type` as parameter (see `fedbiomed.common.constants.ComponentTypes`) and name parameters logs dynamically based on component type. So, `component_id` becomes `node_id` if it is node.

fedbiomed/common/logger.py
```diff
@@ -207,8 +291,152 @@ def __init__(self, level: str = DEFAULT_LOG_LEVEL):
...
+            if caller_frame
+            else "unknown",
+            "caller_module": os.path.basename(caller_frame.f_code.co_filename)
+            if caller_frame
+            else "unknown",
+            "caller_file": caller_frame.f_code.co_filename
+            if caller_frame
+            else "unknown",
+            "caller_line": caller_frame.f_lineno if caller_frame else 0,
+        }
```
srcansiz: I think it could be nice to take this block in try/except  to make sure that it does not break anything during execution. If there is exception during this operation it can log that exception in ->  `caller_parsing` = `success | failure`, `caller_parsing_error` = `exception`.

srcansiz (approved):
Hi @alitolga,

Very good job! Thanks for the PR. I left a few minor comments. Once they are addressed you proceed to merge.

fedbiomed/node/node.py
alitolga: Hi @srcansiz ,

Yes, it would work perfectly fine by just using the `security_event` function as well. I will add this note it in the developer documentation :)

As we discussed, I will test the `async` operations in more detail on `node2node` communication.

alitolga:
> Hi @alitolga,
> 
> Very good job! Thanks for the PR. I left a few minor comments. Once they are addressed you proceed to merge.

Thank you very much for the comment @srcansiz !

The suggested changes are implemented :)

srcansiz:
> > Hi @alitolga,
> > Very good job! Thanks for the PR. I left a few minor comments. Once they are addressed you proceed to merge.
> 
> Thank you very much for the comment @srcansiz !
> 
> The suggested changes are implemented :)

Hi @alitolga, 

Tests are failing could you please check them otherwise we won't be able to merge the branch.

Issue #1604, opened 2026-02-02
Add the Necessary Security Log Entries to the Source Code


alitolga:
- [x] All messages coming and going to/from the node. (This covers most of the specified legal requirements)
  - [x] Messages to/from researcher(s). **No logging on researcher side for now.**
  - [x] Creation, deletion, start, stop of a node.
- [ ] CRUD operations on all the datasets. - **PR Created**
- [ ] Node2node communication (For secagg, diffie-hellman key exchanges...)
- [x] Creation, deletion, start, stop, save of an experiment. - **No logging on researcher side for now.**
- [ ] CRUD operations and approvals/rejections for training plans. - **PR Created**
- [x] Information regarding training/round. Info on each training step, dataset split...
- [ ] CRUD operations on certificates. - **PR Created**
- [ ] CRUD operations on config files. - **PR Created**
- [ ] Optional: Info regarding secagg operations.

alitolga:
Has been completed with PR #1640 . The remaining CRUD operations have been covered by adding logging to the DB Table class.

Issue #1603, opened 2026-02-02
Implement Logger Class according to API design on #1602


alitolga:
### 1) New imports

Add Libraries:

- ContextVar (from contextvars)
- contextmanager (from contextlib)
- datetime, timezone (from datetime)

### 2) New module-level objects/helpers

Add:

- SECURITY_CONTEXT: ContextVar[dict] (holds the bound security context dict)

- **(Optional)** _utc_ts() → UTC ISO timestamp (e.g., 2026-02-05T12:34:56.789Z)

- **(Optional)** Redaction helpers:
  - Looks_sensitive_key(key) detects sensitive keys by substring match
  - Sanitize_fields(fields) redacts sensitive fields and (optionally) truncates large strings
  - Special rule: if config_key looks sensitive, redact old_value/new_value too
  - SecurityOnlyFilter that only lets record.is_security=True reach the security file handler

### 3) New FedLogger state

In FedLogger.__init__ add:

self._security_defaults = {"node_id": None, "fedbiomed_version": None}

### 4) New FedLogger methods

Add:

-  configure_security(node_id=..., fedbiomed_version=...)
- security_context(**ctx) context manager

**How it resets:** It sets context at entry, then in finally calls SECURITY_CONTEXT.reset(token) so the previous context is restored even if an exception happens (prevents context leaks).

- add_security_file_handler(filename, level=INFO)

Writes JSON-only lines ("%(message)s") to file
Uses _SecurityOnlyFilter so only security_event entries go there

- security_event(operation, status, researcher_id, **fields)

Security Defaults + bound context + per-call fields
Writes one JSON line with required fields always present.

5) Redaction helpers examples (what gets redacted)

Looks_sensitive_key() flags keys containing private_key, cert_pem etc.
- private_key_path="/..." → "***REDACTED***"
- cert_pem="-----BEGIN..." → "***REDACTED***"

PR #1612, opened 2026-02-11
Feature/361 security logs db operations

alitolga:
This PR starts implementing issue #1604. It follows up issue #1603 on writing the security logs after the security logger functionalities are implemented.

### DIFF
--- a/fedbiomed/node/dataset_manager/_db_tables.py
+++ b/fedbiomed/node/dataset_manager/_db_tables.py
@@ -20,6 +20,15 @@ class BaseTable(TinyTableConnector):
     # Dataclass is expected to have from_dict and to_dict methods
     _dataclass: Type = None
 
+    def _get_operation_prefix(self) -> str:
+        """Get operation prefix based on table name for security logging."""
+        table_mapping = {
+            "Datasets": "dataset",
+            "Dlps": "dlp",
+            "Dlbs": "dlb",
+        }
+        return table_mapping.get(self._table_name, "table")
+
     def _validate_and_convert_to_dict(self, data: dict) -> dict:
         """Generic validation using the table's dataclass"""
         try:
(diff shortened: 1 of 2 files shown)


fedbiomed/node/dataset_manager/_db_tables.py
```diff
@@ -30,18 +39,88 @@ def _validate_and_convert_to_dict(self, data: dict) -> dict:
...
     def get_validated_entry(self, entry_id: str) -> Optional[Any]:
         """Get entry by ID and return as validated model instance"""
         data = self.get_by_id(entry_id)
         return None if not data else self._dataclass.from_dict(data)
 
+    def all(self) -> List[dict]:
+        """Get all entries with security logging."""
+        result = super().all()
+
+        operation_prefix = self._get_operation_prefix()
```
srcansiz, line 99: Hi @alitolga,

I think the prefix construction can be can automatically rather than calling get_operation_prefix that relies on hard coded table names. I remember in the security logging event there is a function that gets where the function is called. I think similar approach can be implemented in a way to detect which  child class calls db functions. `DatasetTable, DLPtable` etc.

fedbiomed/node/dataset_manager/_db_tables.py
srcansiz, line 14: Hi @alitolga, 

I propose to implement security logging in the `TinyTableConnector` as very generic. Otherwise, we will have to add security logging everytime we add new dataset table even though the operations are same. 

I now that this way logger has precise info about the  action that is performed but I propose to keep information as minimal as possible and be more generic to minimize redundant code.

alitolga:
Hi @srcansiz ,

Thanks for the review. For simplicity, I created another PR with the security logging implemented in `TinyTableConnector `

Issue #1605, opened 2026-02-04
Improve error logging from Round class


sharkovsky:
In some instances, the `Round` class gobbles up the exceptions raised and returns only a super generic error message.

It happens:
- https://github.com/fedbiomed/fedbiomed/blob/6f33025119daec03317ad3f8c2254db014aa871a/fedbiomed/node/round.py#L258
- https://github.com/fedbiomed/fedbiomed/blob/6f33025119daec03317ad3f8c2254db014aa871a/fedbiomed/node/round.py#L285
- https://github.com/fedbiomed/fedbiomed/blob/6f33025119daec03317ad3f8c2254db014aa871a/fedbiomed/node/round.py#L296
- https://github.com/fedbiomed/fedbiomed/blob/6f33025119daec03317ad3f8c2254db014aa871a/fedbiomed/node/round.py#L295
- https://github.com/fedbiomed/fedbiomed/blob/6f33025119daec03317ad3f8c2254db014aa871a/fedbiomed/node/round.py#L304
- and many others, I didn't realize it would be so many.

This is super disruptive because it makes debugging impossible.

I understand that we may not want to communicate too many details to the researcher, but on the node side at least the error message should be logged somewhere!

PR #1623, opened 2026-02-17
Feature/1266 debug mode

### LINKED ISSUE #1266: Create a DEBUG mode in Fed-BioMed, and increase verbosity of DEBUG mode.
With the increasing number of users and diverse implementations, the lack of a DEBUG mode has become more noticeable. It has become increasingly difficult to diagnose errors in Fed-BioMed components, especially those arising from user requests. This issue proposes adding a DEBUG mode to Fed-BioMed components, which increases log verbosity to provide the most detailed output possible, making it easier to debug issues.

- [ ] Allow users to enable debug mode using `export FBM_DEBUG=true`.  
- [ ] Ensure that the debug mode does not expose private or sensitive information from nodes to researcher component.  
- [ ] Implement detailed logging in debug mode, enabling users to follow processes step by step. For example, from a training request to the corresponding training reply, log each block that processes the request on both the node side and the researcher side.  

This issue may also relate to #361. However, while #361 focuses on general security logs, this issue is specifically aimed at debugging purposes.

alitolga:
Fixes #1266 and #1605.

Implement detailed logging on a train request. 

When debug mode is enabled on the node side, logging is more verbose on each step of a training; such as Round creation, training plan initialization, processing optimizer aux variables and training routine initialization.

When debug mode is enabled on the researcher side, the exception handler is modified and prints tracebacks on FedbiomedExperimentErrors.

**Important Note:** I created two different variables to enable debug mode on node side and researcher side separately. They can be merged into one variable (or implemented completely differently), according to request/feedback.

### DIFF
--- a/fedbiomed/common/cli.py
+++ b/fedbiomed/common/cli.py
@@ -119,6 +119,8 @@ class ComponentDirectoryAction(ABC, argparse.Action):
         self.set_component(component_dir)
 
         # this may be changed on command line or in the config_node.ini
+        # FIXME: alitolga: this may remain for debugging purposes until node and researcher components are started
+        # however I don't think the logger level should be set here, and instead be set when node and researcher get started.
         logger.setLevel("DEBUG")
 
 
--- a/fedbiomed/common/cli.py
+++ b/fedbiomed/common/cli.py
@@ -119,9 +119,6 @@ class ComponentDirectoryAction(ABC, argparse.Action):
         self.set_component(component_dir)
 
         # this may be changed on command line or in the config_node.ini
-        # FIXME: alitolga: this may remain for debugging purposes until node and researcher components are started
-        # however I don't think the logger level should be set here, and instead be set when node and researcher get started.
-        logger.setLevel("DEBUG")
 
 
 class CommonCLI:
@@ -139,6 +136,10 @@ class CommonCLI:
         self._certificate_manager: CertificateManager = CertificateManager()
         self._description: str = ""
         self._args = None
+        if os.environ.get("FBM_DEBUG", "").lower() in ("1", "true", "yes"):
+            logger.setLevel("DEBUG")
+        else:
+            logger.setLevel("INFO")
 
     @property
     def parser(self) -> argparse.ArgumentParser:
--- a/fedbiomed/common/db.py
+++ b/fedbiomed/common/db.py
@@ -7,6 +7,7 @@ Interfaces with a tinyDB database for converting search results to dict.
 
 from __future__ import annotations
 
+from functools import wraps
 from typing import Any, Callable, Dict, List, Optional, Union
 
 from tinydb import Query, TinyDB
@@ -14,6 +15,8 @@ from tinydb.table import Document, Table
 
 from fedbiomed.common.exceptions import FedbiomedError
 from fedbiomed.common.logger import logger
+
+LOG_VALUE_MAX_LENGTH = 50
 
 
 def cast_(func):
@@ -58,8 +61,6 @@ def _security_log(operation: str, default_stacklevel: int = 3):
             Extracting/logging relevant information from args/kwargs for security events."""
 
             logging_stacklevel = kwargs.pop("stacklevel", default_stacklevel)
-
-            LOG_VALUE_MAX_LENGTH = 50
 
             # Reuse your helpers
             kwargs_stripped = _strip_forbidden_keys(kwargs)
@@ -180,6 +181,72 @@ def _truncate_for_logging(value: Any, max_len: int) -> Any:
     return value
 
 
+def _debug_db_operation(message: str, **fields: Any) -> None:
+    """Emit sanitized debug logging for database wrapper operations."""
+
+    sanitized = _truncate_for_logging(
+        _strip_forbidden_keys(fields), max_len=LOG_VALUE_MAX_LENGTH
+    )
+    logger.debug(f"{message}: {sanitized}")
+
+
+def _debug_result_summary(result: Any) -> Any:
+    """Build a compact debug summary for DB wrapper results."""
+
+    if isinstance(result, list):
+        return {"result_type": "list", "count": len(result)}
+    if isinstance(result, dict):
+        return {"result_type": "dict", "count": len(result)}
+    if result is None:
+        return {"result_type": "NoneType"}
+    return {"result_type": type(result).__name__, "value": result}
+
+
+def _debug_log(operation: str):
+    """Decorator for sanitized debug logging of database wrapper methods."""
+
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            bound_self = args[0] if args else None
+            call_args = args[1:] if len(args) > 1 else tuple()
+
+            context = {
+                "operation": operation,
+                "component": type(bound_self).__name__
+                if bound_self is not None
+                else None,
+                "table": getattr(
+                    bound_self, "_table_name", getattr(bound_self, "name", None)
+                ),
+                "id_name": getattr(bound_self, "_id_name", None),
+                "args": call_args,
+                "kwargs": kwargs,
+            }
+            _debug_db_operation(f"Starting {operation}", **context)
+
+            try:
+                result = func(*args, **kwargs)
+            except Exception as exc:
+                _debug_db_operation(
+                    f"Failed {operation}",
+                    **context,
+                    error=repr(exc),
+                )
+                raise
+
+            _debug_db_operation(
+                f"Completed {operation}",
+                **context,
+                result=_debug_result_summary(result),
+            )
+            return result
+
+        return wrapper
+
+    return decorator
+
+
 class DBTable(Table):
     """Extends TinyDB table to cast Document type to dict"""
 
@@ -226,6 +293,13 @@ class TinyDBConnector:
             cls._instance = super().__new__(cls)
             # Use DBTable as the default table class for this TinyDB instance
             cls._instance._db = TinyDB(db_path)
+            _debug_db_operation(
+                "Initialized TinyDB connector", db_path=db_path, reused_existing=False
+            )
+        else:
+            _debug_db_operation(
+                "Reusing TinyDB connector", db_path=db_path, reused_existing=True
+            )
         return cls._instance
 
     @property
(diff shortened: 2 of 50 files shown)


fedbiomed/researcher/federated_workflows/_federated_workflow.py
```diff
@@ -118,7 +122,19 @@ def payload(*args, **kwargs):
...
             return ret
 
-        return function(*args, **kwargs)
+        else:
+            try:
+                return function(*args, **kwargs)
+            except FedbiomedSilentTerminationError as exp:
+                traceback.print_exc(file=sys.stdout)
+                logger.debug(f"FedbiomedSilentTerminationError raised: {repr(exp)}")
+                logger.debug(f"Error details: {traceback.format_exc()}")
```
sharkovsky: For the SilentTerminationError, I think we only want to say that it was raised

```suggestion
                logger.debug(f"FedbiomedSilentTerminationError raised.")
```

fedbiomed/researcher/federated_workflows/_federated_workflow.py
```diff
@@ -118,7 +122,19 @@ def payload(*args, **kwargs):
...
+                raise
+            except FedbiomedExperimentError as exp:
+                traceback.print_exc(file=sys.stdout)
+                logger.debug(f"FedbiomedExperimentError raised: {repr(exp)}")
```
sharkovsky: This was probably a `logger.error`?

```suggestion
                logger.error(f"FedbiomedExperimentError raised: {repr(exp)}")
```

fedbiomed/common/cli.py
```diff
@@ -119,6 +119,8 @@ def _create_config(self, component_dir: str):
         self.set_component(component_dir)
 
         # this may be changed on command line or in the config_node.ini
+        # FIXME: alitolga: this may remain for debugging purposes until node and researcher components are started
```
sharkovsky: I'm not a big fan of leaving FIXME in the code.
Let's discuss how we can fix it.

fedbiomed/node/node.py
```diff
@@ -424,24 +427,37 @@ def parser_task_train(self, msg: TrainRequest) -> Union[Round, None]:
...
 
         dataset_id = msg.get_param("dataset_id")
         data = self.dataset_manager.dataset_table.get_by_id(dataset_id)
 
         if data is None:
             return self.send_error(
-                extra_msg="Did not found proper data in local datasets "
-                f"on node={self._node_id}",
+                extra_msg=(
+                    f"{ErrorNumbers.FB313.value}: Did not found proper data in local datasets "
```
sharkovsky: while we're here, let's fix this typo

```suggestion
                    f"{ErrorNumbers.FB313.value}: Did not find proper data in local datasets "
```

sharkovsky (dismissed):
Great work, and nice work identifying many relevant spots where a debug log could be useful. I could't find more so I think it's fine for now, we can always add new ones later.

fedbiomed/researcher/federated_workflows/_federated_workflow.py
alitolga: Yes, makes more sense. I was probably confused to avoid accidentally printing anything on the researcher side 😅

fedbiomed/common/cli.py
alitolga: I left that note to double-check. We set the logger level initially, and it gets overridden when Node or Researcher initializes. This felt incorrect to me, but I guess it has to be this way if we want to do some logging before Node or Researcher starts. 

I can remove it if you also agree that it makes sense.

fedbiomed/node/node.py
alitolga: Applied fix.

fedbiomed/researcher/federated_workflows/_federated_workflow.py
alitolga: Applied fix.

alitolga:
> Great work, and nice work identifying many relevant spots where a debug log could be useful. I could't find more so I think it's fine for now, we can always add new ones later.

Hi @sharkovsky !

Thank you for the review. I pushed a new commit with your suggested fixes.

srcansiz:
Hi @alitolga 

Thank you for the PR. Here are a few comment from my side. I think there are still some improvements to do. 

- [x] I think it is better to use one environ variable for both researcher and node `FBM_DEBUG`
- [x] On the researcher side `federated_workflow` if debug mode is enabled there is not need to catch he errors and raise silent error.
- [x] You may need to add `FBM_DEBUG=True` to the unit tests. You can do it by creating `confest.py` in the `tests/` directory, and setting environment variable in it.  Please keep in mind that this will raise many errors that says that expected exception is not raised on the researcher side because all those errors are caught as `sys.exit` while now it should be the original exception this raised by relavant class. You will have to change all `with pytest.raises`.


These are my first comments. I will be looking at the changes more. I will keep you updated. 

Thanks again for the PR!

alitolga:
Workflow to debug a requests which “hangs” or “disappears”:

1. Was the request created?
Look for logs in FederatedRequest.init / Request.send.

2. Was it queued on the researcher side?
Look in NodeAgent.send_async.

3. Did the node poll and receive it?
Look in GetTaskUnary and TaskListener._call_researcher.

4. Did Node.on_message run?
If no, transport delivery failed.
If yes, application logic received it.

5. Did the node create and send a reply?
Look in GrpcController.send / Sender._call_researcher.

6. Did the researcher receive that reply?
Look in ReplyTask.

7. Did NodeAgent.on_reply match it to a waiting request?
If no, maybe it was flushed early or request ids mismatched.

8. Did a policy terminate the request early?
Look in PolicyController.continue_all and PolicyController.has_stopped_any

9. Did cleanup flush it?
Look in FederatedRequest.exit.

The aforementioned functions are the main functions that are called during the lifecycle of a request, in this written order.

fedbiomed/node/node.py
```diff
@@ -185,6 +188,9 @@ def on_message(self, msg: dict):
         Args:
             msg: Incoming message from Researcher.
         """
+        ### (OPTIONAL) DEBUG THE RECEIVED MESSAGE FROM THE RESEARCHER
+        ### WITH MORE DETAILS THAN THE CURRENT DEBUG MESSAGES
```
srcansiz: +1

fedbiomed/node/node.py
```diff
@@ -424,24 +430,37 @@ def parser_task_train(self, msg: TrainRequest) -> Union[Round, None]:
             researcher_id=msg.researcher_id,
             send=self._grpc_client.send,
         )
+        logger.debug(f"History monitor initialized for experiment {msg.experiment_id}")
```
srcansiz: Hi @alitolga, 

I think it could be nice to add this debug in `HistoryMonitor` class. Additionally, better to add more details about `researcher_id`, `experiment_id`, and maybe a bit detail about where it is called from.

fedbiomed/node/node.py
```diff
@@ -424,24 +430,37 @@ def parser_task_train(self, msg: TrainRequest) -> Union[Round, None]:
...
             dlp_and_loading_block_metadata = self.dataset_manager.get_dlp_by_id(
                 data["dlp_id"]
             )
+            logger.debug(
+                f"DLP and loading block metadata with dlp_id={data['dlp_id']} fetched for dataset_id={dataset_id} on node={self._node_id}"
+            )
+        else:
+            logger.debug(
+                f"No DLP and loading block metadata found for dataset_id={dataset_id} on node={self._node_id}"
+            )
```
srcansiz: Hi @alitolga, 

I think this debug message can be called in the `dataset_manager` to avoid writing debug everywhere in node.py when there is dataset operation.

fedbiomed/node/node.py
srcansiz, line 692: Hi @alitolga, 

This is an important place to add logger debug with a traceback. If something un expected happens in any task execition it falls in here.

srcansiz:
Hi @alitolga, 

I didn't see any debug message on the node to node communication part. I believe it could be useful to have debug logs in the module `nodes/requests`. I don't have precise recommendation about where the logs can be placed. I think you will need to see and identify relevant places.

srcansiz:
Hi @alitolga,

Additionally, it could be nice to document debug mode in `README.md` and `https://fedbiomed.org/support/troubleshooting/` in the documentation.

srcansiz:
Hi @alitolga, 

Thank you for the PR! I think for the first version of the DEBUG mode it is quite good. The branch can be merged after the comments are addressed. Thanks!

fedbiomed/common/db.py
```diff
@@ -180,6 +181,72 @@ def _truncate_for_logging(value: Any, max_len: int) -> Any:
...
+                else None,
+                "table": getattr(
+                    bound_self, "_table_name", getattr(bound_self, "name", None)
+                ),
+                "id_name": getattr(bound_self, "_id_name", None),
+                "args": call_args,
+                "kwargs": kwargs,
+            }
+            _debug_db_operation(f"Starting {operation}", **context)
+
```
srcansiz: Hi @alitolga, 

You already have similar function for `security_log`. Would it be possible to combine these two. For example, if debug mode is activate security llogs for this part also logs for the debugs ? Why did you preferred to separate these two?

fedbiomed/researcher/requests/_policies.py
```diff
@@ -192,6 +194,17 @@ def continue_all(self, requests: List[TRequest]) -> PolicyStatus:
...
         )
 
+        for req in requests:
+            request_id = getattr(req, "_request_id", getattr(req, "request_id", None))
+            node = getattr(req, "_node", getattr(req, "node", None))
+            status_value = getattr(req, "status", None)
+            error_value = getattr(req, "error", None)
+            logger.debug(
+                f"PolicyController: Request {request_id} "
+                f"for node {node} "
```
srcansiz: `node` is an object here. It may be better to give node id instead.

fedbiomed/common/db.py
alitolga: Hi @srcansiz ,

I was looking for a way to make it more verbose. But I was also thinking of merging them in the end, just like you said.

fedbiomed/researcher/requests/_policies.py
alitolga: True, I am fixing it now

alitolga:
> Hi @alitolga,
> 
> Additionally, it could be nice to document debug mode in `README.md` and `https://fedbiomed.org/support/troubleshooting/` in the documentation.

Hi @srcansiz ,

I added a small documentation to the README and the troubleshooting page.

sharkovsky (approved):

Issue #1609, opened 2026-02-09
Create new entry type in the DB for pre-processed datasets


srcansiz:
Implement a pre-processed dataset entry and its handler in the database for saving preprocessed dataset after FedCombat preprocessing operation. The implementation should be generic that it can be used for other type of preprocessed dataset results as well. 

Here are a few points to consider during the implementation:

- Preprocessing happens via researcher (user) requests through experiments; therefore, it should be considered different from node-deployed datasets.
   * This includes saving these datasets to a different folder than the node-deployed datasets (which should not be writable to the node)
   * Also these datasets probably don't have tags (or they are optional), and so they are not advertised in a `search()` (only accessible through their `dataset_id`)
   * They may be attached to an `experiment_id` (so there may be an (?)optional field for that)
   * They may be deleted by researcher's request (from the correct `experiment_id`), so `DatasetTable.delete_by_id()` supports the case of pre-processed dataset 
- Preprocessed datasets are derived datasets from node-deployed datasets; therefore, the link between raw data and preprocessed data should be kept.
- Preprocessed data can also be derived from another preprocessed dataset. For example, Dataset A → preprocess → Dataset A' → preprocess → Dataset A''. This may require keeping two entries in the database as `dataset_id`: raw dataset ID, `preprocess_history`: <list-of-preprocessed-datasets-that-are-used-in-order>
- If not complicating implement with `DynamicDataset` terminology, as it may be re-used for other purposes (eg: post-processing, ...).

Here are the steps of possible implementation. 

- [x] identify the fields that will be added in the database for preprocessed dataset, and create `PreprocDatasetEntry` in `fedbiomed/node/dataset_manager/_db_dataclasses.py` Example possible fields: `preproc_id`, `preprocessed_dataset_id`, `raw_dataset_id`, `experiment_id`, `research_id`.
- [x] Implement `PreprocDatasetTable` in `fedbiomed/node/dataset_manager/_db_tables.` 
- [x] Implement a new method, something like `add_preprocessed_data`, in `DatasetManager`, similar to the existing `add_dataset` method.
- [x] Unittests

PR #1630, opened 2026-02-18
Implement dynamic dataset storage and deletion logic in database

lchambon:
**PR description**

First draft implementation of a new generic entry in the database for dynamic datasets (datasets created by researchers through pre-processing or post-processing operations for example).

**Questions / Feedback needed:**
- Need to cross-check all entries in the new `DynamicDatasetEntry`.
- Should the `DatasetManager` also be responsible for saving new dataset files? Currently, in this draft, DatasetManager saves a new CSV file when creating a Dynamic Dataset.
- Who should be responsible for setting the path for saving new files from a Dynamic Dataset?
- How should we manage different types of datasets? The current draft only considers CSV files.
- When deleting a dataset in `DatasetTable`, should we recursively delete all entries in `DynamicDatasetTable` and the corresponding files created from this dataset ? Currently, `delete_dataset_by_id` in `DatasetManager` handles recursive deletion for both datasets in `DatasetTable` and `DynamicDatasetTable`, but there are no changes in `delete_by_id` of DatasetTable. Should we replace `delete_by_id` by the new `delete_dataset_by_id` in gui or cli ?

**Missing:**
- [x] Unit tests

Related issue: #1609

### DIFF
--- a/fedbiomed/node/dataset_manager/_dataset_manager.py
+++ b/fedbiomed/node/dataset_manager/_dataset_manager.py
@@ -193,6 +193,16 @@ class DatasetManager:
         # Validate parent
         parent_entry = self.get_dataset_entry_by_id(parent_dataset_id)
 
+        # Get controller
+        data_type = parent_entry["data_type"]
+        controller = get_controller(
+            data_type,
+            controller_parameters={
+                "root": path,
+                **(dataset_parameters if dataset_parameters is not None else {}),
+            },
+        )
+
         # Insert partial DB entry
         entry = {
             "path": path,
@@ -205,28 +215,11 @@ class DatasetManager:
             "description": description,
             "dataset_id": dataset_id,
             "dataset_parameters": dataset_parameters,
+            "shape": controller.shape(),
+            "dtypes": controller.get_types(),
+            "data_type": data_type,
         }
         dataset_id = self.dynamic_dataset_table.insert(entry)
-
-        # Get controller
-        data_type = parent_entry["data_type"]
-        controller = get_controller(
-            data_type,
-            controller_parameters={
-                "root": path,
-                **(dataset_parameters if dataset_parameters is not None else {}),
-            },
-        )
-
-        # Update DB entry with path, shape and dtypes
-        self.dynamic_dataset_table.update_by_id(
-            dataset_id,
-            {
-                "shape": controller.shape(),
-                "dtypes": controller.get_types(),
-                "data_type": data_type,
-            },
-        )
 
         return dataset_id
 
--- a/fedbiomed/node/dataset_manager/_dataset_manager.py
+++ b/fedbiomed/node/dataset_manager/_dataset_manager.py
@@ -6,6 +6,9 @@ Interfaces with the node component database.
 """
 
 import csv
+import os
+import shutil
+from pathlib import Path
 from typing import Iterable, List, Optional, Tuple, Union
 
 import pandas as pd
@@ -15,7 +18,12 @@ from fedbiomed.common.constants import ErrorNumbers
 from fedbiomed.common.dataloadingplan import DataLoadingPlan
 from fedbiomed.common.dataset import get_controller
 from fedbiomed.common.exceptions import FedbiomedError
-from fedbiomed.node.dataset_manager._db_tables import DatasetTable, DlbTable, DlpTable
+from fedbiomed.node.dataset_manager._db_tables import (
+    DatasetTable,
+    DlbTable,
+    DlpTable,
+    DynamicDatasetTable,
+)
 
 
 class DatasetManager:
@@ -25,15 +33,22 @@ class DatasetManager:
     for the node. Currently uses TinyDB.
     """
 
-    def __init__(self, path: str):
+    def __init__(self, path: str, dynamic_datasets_path: str = None):
         """Initialize with database path."""
         self._dataset_table = DatasetTable(path)
+        self._dynamic_dataset_table = DynamicDatasetTable(path)
         self._dlp_table = DlpTable(path)
         self._dlb_table = DlbTable(path)
+        self._dynamic_datasets_path = dynamic_datasets_path
+        self.create_dynamic_datasets_folder()
 
     @property
     def dataset_table(self) -> DatasetTable:
         return self._dataset_table
+
+    @property
+    def dynamic_dataset_table(self) -> DynamicDatasetTable:
+        return self._dynamic_dataset_table
 
     @property
     def dlp_table(self) -> DlpTable:
@@ -42,6 +57,22 @@ class DatasetManager:
     @property
     def dlb_table(self) -> DlbTable:
         return self._dlb_table
+
+    def create_dynamic_datasets_folder(self):
+        """Creates the folder for storing dynamic datasets if it does not exist.
+
+        Raises:
+            FedbiomedError:
+            - If the folder creation fails.
+        """
+        if self._dynamic_datasets_path:
+            try:
+                os.makedirs(self._dynamic_datasets_path, exist_ok=True)
+            except Exception as e:
+                raise FedbiomedError(
+                    f"{ErrorNumbers.FB632.value}: Failed to create dynamic datasets folder at "
+                    f"{self._dynamic_datasets_path}: {str(e)}"
+                ) from e
 
     def get_dlp_by_id(self, dlp_id: str) -> Tuple[Optional[dict], List[dict]]:
         """Get data loading plan by ID and its associated data loading blocks."""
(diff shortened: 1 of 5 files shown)


lchambon:
**Updates after discussion with Marc, Ali Tolga and Jhonathan**

Following our discussion, here are the main conclusions and next steps regarding this draft PR:

### **1. Responsability for saving dataset files**
We agreed that `DatasetManager` should not be responsible for saving dataset files to disk.

This responsibility should likely belong to another dedicated class (to be defined), along with a clearly defined configuration variable specifying where new datasets must be stored.

In the revised version of this draft PR:

- It will be assumed that dataset files are already saved.
- `DatasetManager` will only be responsible for registering the dataset entry in the database, including its path.
- The current implementation will be simplified accordingly.

### **2. Current scope and impact**

At the moment, the concrete usage of dynamic datasets is not fully defined.
We expect the first real use case to come from the continued development of FedCombat.

For now, this implementation:
- Only affects the `dataset_manager` package
- Adds new logic in: `_dataset_manager.py`,  `_db_dataclasses.py`, `_db_tables.py`
- Should not break any existing behavior in Fed-BioMed

### **3. Future refactoring of `DatasetManager` usage**

We will later need to review how `DatasetManager` is used across the codebase (CLI, GUI, jobs, node.py, etc.).

For example (not exhaustive):

- Calls such as `dataset_manager.dataset_table.delete_by_id()` should likely be replaced by a method directly handled by `DatasetManager`, so that deleting a dataset also deletes all derived dynamic datasets (currently implemented as `delete_dataset_by_id` in this draft).
- Similarly, calls such as `dataset_manager.dataset_table.get_by_id()` may be replaced by a method on `DatasetManager` that searches in both `DatasetTable` and `DynamicDatasetTable` (currently implemented as `get_dataset_entry_by_id` in this draft).
- More generally, we may want to avoid direct access to internal tables and instead expose dedicated methods in DatasetManager to handle these operations.
- `dataset_manager.list_my_datasets()`: should this also return dynamic datasets?
- `dataset_manager.dataset_table.search_by_tags` : should this also return dynamic datasets?
- ...

### **4. Other questions**

We also discussed whether introducing a separate `DynamicDatasetTable` is the simplest long-term solution.

An alternative would be to store dynamic datasets directly in `DatasetTable`, with an additional field distinguishing raw and dynamic datasets.

For now, we keep the implementation with two distinct tables, but this may be reconsidered later.

fedbiomed/node/dataset_manager/_dataset_manager.py
```diff
@@ -146,6 +156,183 @@ def add_database(
...
+        )
+
+        # Update DB entry with path, shape and dtypes
+        self.dynamic_dataset_table.update_by_id(
+            dataset_id,
+            {
+                "shape": controller.shape(),
+                "dtypes": controller.get_types(),
+                "data_type": data_type,
+            },
```
alitolga: Hi @lchambon ,

Is there a specific reason why you don't add these during the insert operation as well?

alitolga (commented):
Thank you for the PR @lchambon !

I made a small comment, but I believe it's good to merge.

fedbiomed/node/dataset_manager/_dataset_manager.py
lchambon: Hi @alitolga, you're absolutely right. I can call the controller first and add these during insert as well. I had done it the other way in the first version of the PR, where I was also managing saving dataset files, and I forgot to update it. I’ll correct that. Thank you!

fedbiomed/node/dataset_manager/_dataset_manager.py
```diff
@@ -146,6 +156,176 @@ def add_database(
...
+            recursive: whether to recursively delete all descendant dynamic datasets (if any). Defaults to False.
+            force: whether to force delete the dataset if it has children, by reassigning its children to its parent dataset. Defaults to False.
+
+        Raises:
+            FedbiomedError:
+            - If the dataset has children and neither recursive nor force is True
+            - If no dataset is found with the given ID
+        """
+        dataset_entry = self.get_dataset_entry_by_id(dataset_id)
+        is_dynamic = "parent_dataset_id" in dataset_entry
```
sharkovsky: It might be better to have an explicit is_dynamic entry in the dataset, rather than implicitly inferring it from the existence of `parent_dataset_id`, because if in the future we design some other types of dataset that are not "dynamic" datasets but have a parent for whatever reason, this will generate a (maybe silent) bug.

fedbiomed/node/dataset_manager/_dataset_manager.py
```diff
@@ -146,6 +156,176 @@ def add_database(
...
+
+        children = self.dynamic_dataset_table.get_all_by_value(
+            "parent_dataset_id", dataset_id
+        )
+
+        # Case 1: dataset
+        if not is_dynamic:
+            if children:
+                if force:
```
sharkovsky: Would it be worth it to consider giving the `force` option another name, since in fact it fails to "force" the deletion in this case?

fedbiomed/node/dataset_manager/_dataset_manager.py
```diff
@@ -146,6 +156,176 @@ def add_database(
...
+
+        # Case 2: dynamic dataset
+        if children and not recursive and not force:
+            raise FedbiomedError(
+                f"{ErrorNumbers.FB632.value}: Dataset has children. Use recursive=True to delete subtree "
+                "or force=True to reassign children."
+            )
+
+        if recursive:
+            self._delete_dynamic_subtree(dataset_id, is_dynamic)
```
sharkovsky: why don't you need to also delete the dataset itself here?

sharkovsky (approved):

fedbiomed/node/dataset_manager/_dataset_manager.py
```diff
@@ -146,6 +156,176 @@ def add_database(
 
         return dataset_entry
 
+    def add_dynamic_dataset(
```
srcansiz, line 143: Hi @lchambon, 

Thank you for the PR! 

I only have one comment concerning the similarity between `add_dataset` and `add_dynamic_dataset` methods. It is a minor comment since it is just about reducing duplicated lines in the both methods. The idea is to handle common db fields from one place. 

They have many common properties, including retrieving the controller. So  I propose to implement a method `_build_dataset_entry` that can be used for both `add_dataset` and `add_dynamic_dataset`. Something like this : 

> BTW, I generated following code snippet using copilot :)

```
def _build_dataset_entry(
    self,
    *,
    data_type: str,
    path: str,
    dataset_id: Optional[str],
    dataset_parameters: Optional[dict],
    data_loading_plan: Optional[DataLoadingPlan] = None,
    extra_fields: Optional[dict] = None,
):


    controller_params = {
        "root": path,
        **(dataset_parameters or {}),
    }

    if data_loading_plan is not None:
        controller_params["dlp"] = data_loading_plan

    controller = get_controller(
        data_type,
        controller_parameters=controller_params,
    )
    # Creating common properties -------
    entry = {
        "dataset_id": dataset_id,
        "data_type": data_type,
        "path": controller._controller_kwargs["root"],
        "shape": controller.shape(),
        "dtypes": controller.get_types(),
        "dataset_parameters": {
            k: v
            for k, v in controller._controller_kwargs.items()
            if k not in ["root", "dlp"]
        },
    }

    if data_loading_plan is not None:
        entry["dlp_id"] = data_loading_plan.dlp_id
    # Adding extra fields. 
    if extra_fields:
        entry.update(extra_fields)

    return entry
```

Then,  it can be used  as follow for dynamic dataset by adding extra fields, and without `extra_fields` for normal datasets. 

```
    entry = self._build_dataset_entry(
        data_type=parent_entry["data_type"],
        path=path,
        dataset_id=dataset_id,
        dataset_parameters=dataset_parameters,
        extra_fields={
            "researcher_id": researcher_id,
            "experiment_id": experiment_id,
            "processing_id": processing_id,
            "parent_dataset_id": parent_dataset_id,
            "name": name,
            "tags": tags,
            "description": description,
        },
    )

 self.dynamic_dataset_table.insert(entry)
```

fedbiomed/node/dataset_manager/_dataset_manager.py
lchambon: Thank you @sharkovsky for this suggestion. In this case, I am wondering if we still need to have two separated tables: dataset table and dynamic dataset table. We could manage both types of datasets in the same table. What do you think ?

fedbiomed/node/dataset_manager/_dataset_manager.py
lchambon: Yes I will think about it. I added this force option mainly for the case where the dataset that must be deleted is a dynamic type.

fedbiomed/node/dataset_manager/_dataset_manager.py
lchambon: When you call the `_delete_dynamic_subtree` function for a **dynamic dataset**, it deletes the entire subtree and the dataset itself. This happens because the `collect_subtree` function of `DynamicDatasetTable` also includes the parent dataset ID.

In contrast, when `_delete_dynamic_subtree` is called for a **'raw' dataset**, it only deletes the dynamic datasets in the subtree. I made the choice to separately delete the raw dataset itself.

But there might be a smarter way to handle this. I will think about it :)

fedbiomed/node/dataset_manager/_dataset_manager.py
lchambon: Thank you @srcansiz ! I will update the PR with your suggestion.

fedbiomed/node/dataset_manager/_dataset_manager.py
srcansiz: Hi @sharkovsky, @lchambon, 

I understand the concern but the table is already called `dynamic_dataset_table` so it is already explicit. if there will be other dataset types that are not dynamic but has parents those datasets can go to main `dataset` table and main `dataset` can have another field called `parent_id`. 

What do you think ?

PR #1631, opened 2026-02-19
Feature/361 security logs gui

alitolga:
Initial implementation for GUI extension for security logs.

### DIFF
--- a/fedbiomed_gui/ui/src/pages/admin/SecurityLogs.js
+++ b/fedbiomed_gui/ui/src/pages/admin/SecurityLogs.js
@@ -3,6 +3,7 @@ import axios from 'axios'
 import {
   EuiBasicTable,
   EuiButton,
+  EuiButtonEmpty,
   EuiFlexGroup,
   EuiFlexItem,
   EuiFieldNumber,
@@ -12,13 +13,11 @@ import {
   EuiSelect,
   EuiSpacer,
   EuiSwitch,
-  EuiTablePagination,
   EuiText,
 } from '@elastic/eui'
 import moment from 'moment'
 
 import { EP_SECURITY_LOGS } from '../../constants'
-
 
 function normalizeItem(it) {
   if (!it || typeof it !== 'object') return { details: '' }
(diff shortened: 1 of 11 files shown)


srcansiz:
Hi @alitolga, 

Thank you for the PR. It is very nice the table to display security logs. There a few things to check and correct

- [x] Please add date time filter to filter out security logs. 
- [x] It is important that the security logs should be visible only for admins. 
- [x] It seems that log file is manually selected. What is there are hundreds of log files? It could be nice to select the logs files automatically for the selected date. If possible combine multiple files for given date time interval to display in the table. 
- [x] Please test security logs generation 10000 logs per files and multiple file to see how HTML view will bhave. We don't want t web browser crash because of long HTML pages.

fedbiomed_gui/ui/src/pages/admin/SecurityLogs.js
```diff
@@ -0,0 +1,460 @@
...
+    try {
+      const res = await axios.get(EP_SECURITY_LOGS, { params: buildParams() })
+      const payload = res?.data?.result || {}
+      const rawItems = payload.items || []
+      const newItems = rawItems.map((it) => normalizeItem(it))
+      setItems(newItems)
+      setLastBatchSize(newItems.length)
+    } finally {
+      setLoading(false)
+    }
```
srcansiz: Hi @alitolga, 

One last comment on these calls. In the other parts of software these called located in other place. `store` directory. Additionally, API calls and data is managed through `redux` API, to create global state. However, I see that security logs is quite separated, so no need to do it for security logs. Anyway, it could be nice to collect these API calls in a dedicated place. I recommend you to create `store/services` directory and create file call `securityLogsServices.js` and add this API call there and import in this module. What do you think?

fedbiomed_gui/ui/src/pages/admin/SecurityLogs.js
```diff
@@ -0,0 +1,460 @@
...
+  )
+
+  const cellNoWrap = { whiteSpace: 'nowrap' }
+  const cellWrap = { whiteSpace: 'normal', wordBreak: 'break-word' }
+
+  const columns = [
+    {
+      field: 'timestamp',
+      name: 'Timestamp',
+      width: '180px',
```
srcansiz: Hi @alitolga,

Hard coding width is not a good practice. We may end up width unnecessary large table, it could be nice to remove `width` and allow auto sizing.

fedbiomed_gui/ui/src/pages/admin/SecurityLogs.js
alitolga: Hi @srcansiz ,

Sure, I can do that. Or I can convert this flow also to a redux flow. Maybe, that's better as it's more consistent with the rest of the front-end?

PR #1634, opened 2026-02-23
Feature/1615 logistic regression example

alitolga:
I added a tutorial that uses Flamby Heart Disease dataset with a Logistic Regression model. The model seems to be training well and performing okay. 

I am creating the draft PR for now, however I believe the performance can be still improved. Feel free to add your suggestions/improvements.


mvesin:
Pushed a new version of the ADNI dataset logistic regression example.

- applying advices from @lchambon and @alitolga (constant LR, lower eta0) - thanks again !
- predicting an easier case (FAQ score from other cognition tests)
- purposely creating unbalanced datasets for the 2 clients (one with the high FAQ scores, one with the low FAQ scores)

It now gives reasonably good results on (distinct) test dataset + show similar performance to training centrally with full dataset + shows better performance than training with only with dataset of one client (which gives many false positive for the other class)

We probably can merge the 2 notebooks if they both reach satisfying results.

mvesin:
Nice work @alitolga !

Please find my notes from reviewing the flamby fedheart notebook

- some path still contain your homedir
- maybe add a small paragraph at proper place to mention "now add datasets to nodes"
- in TP `def testing_step` there is still some debug `print()`
- local training (centralized) does not work because it uses df_c1 df_c2 which are not existing

alitolga:
Hi @mvesin ,

Thank you for the review ! I added the notes you mentioned, you should be able to go through the whole tutorial now without any issues :)

> Nice work @alitolga !
> 
> Please find my notes from reviewing the flamby fedheart notebook
> 
> * some path still contain your homedir
> * maybe add a small paragraph at proper place to mention "now add datasets to nodes"
> * in TP `def testing_step` there is still some debug `print()`
> * local training (centralized) does not work because it uses df_c1 df_c2 which are not existing

mvesin:
Fine work @alitolga, great thanks ! Ready to merge.

mvesin (approved):

Issue #1643, opened 2026-02-27
`FedProx` does not accept integer `fedprox_mu`


michellewang:
This is very very minor but I get an error when I try to use `FedProx` if my `fedprox_mu` is an integer (e.g. `1` instead of a float (`1.`).

To reproduce:
1. Run 101 notebook with these training args:
```python
training_args = {
    "fedprox_mu": 1,  # --> ERROR
    "loader_args": {
        "batch_size": 1,
    },
    "optimizer_args": {"lr": 1e-3},
    # 'test_ratio' : 0.25,
    # 'test_batch_size': 64,
    # 'test_metric': MetricTypes.F1_SCORE,
    # 'test_on_global_updates': True,
    # 'test_on_local_updates': True,
    # 'test_metric_args': {'average': 'marco'},
    # 'use_gpu': True,  # automatically falls back to cpu on nodes that don't support gpu
    "epochs": 1,
    "dry_run": False,
    "batch_maxnum": 100,  # Fast pass for development : only use ( batch_maxnum * batch_size ) samples
}
```

Traceback for `Experiment` instantiation:
```
2026-02-27 12:16:40,812 fedbiomed INFO - Updating training data. This action will update FederatedDataset, and the nodes that will participate to the experiment.2026-02-27 12:16:40,857 fedbiomed DEBUG - Node: NODE_07412de2-c857-4d77-a5e7-43c68c4b8a76 polling for the tasks2026-02-27 12:16:40,868 fedbiomed INFO - Node selected for training -> test
Node ID is -> NODE_07412de2-c857-4d77-a5e7-43c68c4b8a762026-02-27 12:16:40,929 fedbiomed CRITICAL - FB414: bad type or value for training arguments: Expected `fedprox_mu` value is float, but got <class 'int'>. 

--------------------
Fed-BioMed researcher stopped due to exception:
FB414: bad type or value for training arguments: Expected `fedprox_mu` value is float, but got <class 'int'>. 
--------------------
```

This can be easily fixed by setting `fedprox_mu` to `1.` instead, but it feels like Fed-BioMed should be less strict in the type-checking here for better user experience.

srcansiz:
Hi @michellewang, 

Thank you for reporting. Having these feedbacks are important for us to increase the user experience. I agree that Fed-BioMed should be less strict on integer and float distinction. We are going to address this issue as soon as possible.

alitolga:
Hi @michellewang ,

We have fixed the issue in the latest PR.

PR #1649, opened 2026-03-03
Fix rule for checking fedprox value.

alitolga:
This PR fixes bug on #1643 when an int value is given to fedprox.

### DIFF
--- a/fedbiomed/common/training_args.py
+++ b/fedbiomed/common/training_args.py
@@ -238,18 +238,16 @@ class TrainingArgs:
     def _fedprox_mu_validator(
         val: Union[float, int, None],
     ) -> Union[Tuple[bool, str], bool]:
-        """Validates fedprox_mu value whether it None or float or int
+        """Validates fedprox_mu type whether it None or float or int
 
         Returns:
             Validation status  or/and error message
         """
-        if isinstance(val, float) or isinstance(val, int):
-            return True
-        elif val is None:
+        if val is None or isinstance(val, (int, float)):
             return True
         return (
             False,
-            f"Expected `fedprox_mu` value is float or int, but got {type(val)}. ",
+            f"Expected `fedprox_mu` type is float or int, but got {type(val)}. ",
         )
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


fedbiomed/common/training_args.py
```diff
@@ -235,17 +235,22 @@ def _metric_validation_hook(
...
+    def _fedprox_mu_validator(
+        val: Union[float, int, None],
+    ) -> Union[Tuple[bool, str], bool]:
+        """Validates fedprox_mu value whether it None or float or int
 
         Returns:
             Validation status  or/and error message
         """
-        if isinstance(val, float):
+        if isinstance(val, float) or isinstance(val, int):
```
leonard2103: This inline validates at once: 
```
if val is None or isinstance(val, (int, float)):
```

fedbiomed/common/training_args.py
```diff
@@ -235,17 +235,22 @@ def _metric_validation_hook(
...
             return True
         elif val is None:
             return True
-        return False, f"Expected `fedprox_mu` value is float, but got {type(val)}. "
+        return (
+            False,
+            f"Expected `fedprox_mu` value is float or int, but got {type(val)}. ",
```
leonard2103: type not value, you can update like this:
```
f"Expected `fedprox_mu` type to be float or int, but got {type(val)}. "
```

fedbiomed/common/training_args.py
```diff
@@ -235,17 +235,22 @@ def _metric_validation_hook(
 
     @staticmethod
     @validator_decorator
-    def _fedprox_mu_validator(val: Union[float, None]) -> Union[Tuple[bool, str], bool]:
-        """Validates fedprox_mu value whether it None or float
+    def _fedprox_mu_validator(
+        val: Union[float, int, None],
+    ) -> Union[Tuple[bool, str], bool]:
+        """Validates fedprox_mu value whether it None or float or int
```
leonard2103: Same as below, this is related to type, not value.
```
Validate fedprox_mu type to be ``None`` or ``float`` or ``int``
```

leonard2103 (changes requested):

alitolga:
Hi @leonard2103 ,

Thanks for the review. I added your suggested changes :)

PR #1658, opened 2026-03-06
Feature/1652 fa adaptation mfd

### LINKED ISSUE #1652: Adaptation of MFD to generate analytics schema

leonard2103:
- `fa_args` to `stats_args`
- Update of schema for image and tabular datasets.
- Addition of schema for mfd.

closes #1652 

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/fedbiomed/common/analytics/_orchestrator.py
+++ b/fedbiomed/common/analytics/_orchestrator.py
@@ -118,27 +118,29 @@ class AnalyticsOrchestrator:
         schema: Any,
         subschema: Optional[Any],
         stats: Optional[List[str]],
-        args: Optional[Dict[str, Any]],
+        stats_args: Optional[Dict[str, Any]],
         n_samples: int,
     ) -> Dict[str, Any]:
         """Validates inputs and builds the configuration tree in a single pass."""
         # Dispatch based on schema type
         if isinstance(schema, dict):
-            return self._handle_dict(schema, subschema, stats, args, n_samples)
+            return self._handle_dict(schema, subschema, stats, stats_args, n_samples)
         if isinstance(schema, (list, tuple)):
-            return self._handle_sequence(schema, subschema, stats, args, n_samples)
+            return self._handle_sequence(
+                schema, subschema, stats, stats_args, n_samples
+            )
 
         # Leaf Handling
         element_type = schema.type if isinstance(schema, DatasetElementSpec) else None
         if element_type == DatasetElementType.ROW:
-            return self._handle_row(schema, subschema, stats, args, n_samples)
+            return self._handle_row(schema, subschema, stats, stats_args, n_samples)
         if element_type == DatasetElementType.IMAGE:
             # Image does not support subschema selection
             if subschema is not None:
                 raise FedbiomedError(
                     "Subschema selection is not applicable for IMAGE type."
                 )
-            return self._handle_image(stats, args, n_samples)
+            return self._handle_image(stats, stats_args, n_samples)
 
         raise FedbiomedError(f"Unsupported schema type or element type: {type(schema)}")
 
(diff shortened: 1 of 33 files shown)


fedbiomed/common/analytics/_orchestrator.py
```diff
@@ -31,48 +32,56 @@ def compute_stats(
...
             raise FedbiomedError(
-                "Dataset does not implement 'get_analytics_schema' and 'get_analytics_item'."
+                "At least one of 'stats' or 'stats_args' must be provided."
             )
 
+        # Analytics currently only supports datasets that return data in numpy format
+        if dataset.to_format != DataReturnFormat.SKLEARN:
+            raise FedbiomedError(
+                f"Dataset format: '{dataset.to_format.value}' is not supported for analytics."
+            )
```
alitolga, line 62: Hi @leonard2103 ,

Since you have added the check in here, do you think it is still necessary to also add casting `value = np.asarray(value)` in the `update` methods of the data accumulator classes?

Just a minor suggestion.

fedbiomed/researcher/federated_workflows/_federated_analytics.py
```diff
@@ -456,57 +456,75 @@ def get_node_ids(self) -> list[str]:
...
 
-        missing = [s for s in stats if cached is None or not cached.has_stat(s)]
+        # When stats=None the request is entirely args-driven: only skip if already cached.
+        # When stats is a list, only request the individual stats not yet in the cache.
+        if stats is None:
+            need_request = cached is None
+            missing = None
+        else:
+            missing = [s for s in stats if cached is None or not cached.has_stat(s)]
+            need_request = bool(missing)
```
alitolga, line 522: I couldn't follow what is the expected default behaviour here.

If the user doesn't explicitly provide any stats, do we just return whatever we currently have in the cache? And what do we exactly return if we don't currently have anything in the cache?

tests/test_analytics/test_analytics_orchestrator.py
```diff
@@ -525,6 +569,19 @@ def test_resolve_and_validate_roots_conflicting_stat_args(mock_registry, orchest
...
+def test_resolve_and_validate_roots_dependency_conflict(mock_registry, orchestrator):
+    """Two stats imply the same dependency but with conflicting arguments."""
+    # stat1 depends on dep, stat2 depends on dep
+    mock_registry.get_dependencies.side_effect = lambda s, et: {"dep"}
+    mock_registry.validate_args.return_value = None
+
+    with pytest.raises(FedbiomedError, match="Conflicting arguments for dependency"):
+        orchestrator._resolve_and_validate_roots(
+            DatasetElementType.ROW, {"stat1": {"a": 1}, "stat2": {"a": 2}}
+        )
```
alitolga, line 663: What would be an example for this case?

alitolga (commented):
Very nice work @leonard2103 !

Thank you for the PR. I left a few comments to understand a bit better some changes.

fedbiomed/common/analytics/_orchestrator.py
leonard2103: You are right, but `np.asarray` acts as a protection for the operations (expect numpy arrays) that does not add complexity or relocation in memory. This lines in _orchestrator could change eventually if torch format would be enabled at some point.

fedbiomed/researcher/federated_workflows/_federated_analytics.py
leonard2103: This is basically an effort to cache some results that still needs to be improved. Cache results are not re requested from the nodes. However, all stats are stored together and that is not good as changing a parameter in `histogram` would not change what is received for `mean`. Still to be addressed, I was expecting to cache just stats without parameters or separate cache for stats with and without parameters.

tests/test_analytics/test_analytics_orchestrator.py
leonard2103: Does not happen at the moment, but is not allowed. If `mean` and `count` receive different parameters, the same function should not be used for both at once.

alitolga: For example, in such a scenario: The count calculated by the mean would be discarded, and there would be another request to calculate the count with the user given parameters, right?

leonard2103: In that case things would need to be calculated independently but aggregation would need to handle such cases as it is not the case at the moment.

PR #1673, opened 2026-03-19
Feature/1661 syslog

### LINKED ISSUE #1661: Adding security logs to the system logs.
Suggested by @mvesin, it could be useful to investigate libraries such as syslog to forward the security logs to the system as well.

alitolga:
This PR adds a field to the config named syslog, which gives the user the option to forward the logs. The logs will be forwarded to the syslog daemon by default, or to any other desired path/server if host and port information are given by the user.

When the node is created, syslog forwarding is disabled by default. To enable it, make enable=True under the [syslog] section in the configuration file. Under this section, the user can add "host", "port", "facility", "level" and "protocol". If host and port are not specified, Fed-BioMed forwards logs to localhost:514. Facility is the category of the syslog message. If facility is not specified, it defaults to "user". If level is not specified, it defaults to "info". The optional protocol field can be set to udp or tcp; the default is udp.


fedbiomed/common/config.py
```diff
@@ -76,6 +76,42 @@ def __init__(self, root: str) -> None:
         # This ensures security events are captured even before Node/Researcher initialization
         logger.set_security_logs(root_path=root)
         self.load(root)
+        self.add_syslog_from_config()
+
+    def add_syslog_from_config(self):
+        if not self.getbool("syslog", "enable", fallback="False"):
+            logger.info(
```
mvesin, line 83: Misc: You probably can merge the 2 `logger.info()` in one message ?

fedbiomed/common/config.py
```diff
@@ -76,6 +76,42 @@ def __init__(self, root: str) -> None:
...
+
+        host = self.get("syslog", "host", fallback="localhost")
+        port = self.getint("syslog", "port", fallback=514)
+        protocol = self.get("syslog", "protocol", fallback="udp").lower()
+        facility_name = self.get("syslog", "facility", fallback="user").lower()
+        level_name = self.get("syslog", "level", fallback="INFO").upper()
+
+        if protocol not in {"udp", "tcp"}:
+            raise FedbiomedConfigurationError(
+                f"Unsupported syslog protocol: {protocol}"
```
mvesin: Misc: we normally have the convention of specifying a ErrorNumbers.xxx for all errors we raise eg:

```
raise FedbiomedConfigurationError(f"{ErrorNumbers.FB600.value}: ....")
```

This applies to all the subsequent errors also.

fedbiomed/common/config.py
```diff
@@ -76,6 +76,42 @@ def __init__(self, root: str) -> None:
...
+
+    def add_syslog_from_config(self):
+        if not self.getbool("syslog", "enable", fallback="False"):
+            logger.info(
+                "Syslog configuration is disabled or not found in configuration"
+            )
+            logger.info("Disabling syslog logging.")
+            return
+
+        host = self.get("syslog", "host", fallback="localhost")
```
mvesin, line 89: This code will work, as it handles the case where the syslog entry does not exist in the config file.

Yet it would be nice to also integrate these FB practices for managing config file:

* update version number when we change config file content
   
   This is in fedbiomed.common.constants `__node_config_version__` and `__researcher_config_version__`

   Probably a minor version update (eg 3.1.0 => 3.2.0) because the file is backward compatible (we can us e previous file format with current version, as you can cope with the case the syslog does not exist)

   You can check code of `raise_for_version_compatibility` to understand how we check versions.

* update the `fedbiomed.{node,researcher}.config`  module `Config.migrate()` functions to raise warning & update file when the syslog entry does not exist.

   Migrate is a heuristic helper function to help update config files generated with older FB version.
   If you read the existing functions, this is rather self explanatory
   You can track in `common.Config()` the calls from `load()` ... => `migrate()` to understand
   how config file generation and reading is done - if you did not do that yet - to gain some
   background on the topic.

udp_server.py
```diff
@@ -0,0 +1,39 @@
+import queue
```
mvesin: I would propose not to merge the test script `udp_server.py`

Yet if you think it is useful, it would be nice to move it to the `scripts` directory, don't you think so ?

fedbiomed/common/logger.py
```diff
@@ -564,6 +588,48 @@ def _internal_level_translator(self, level: Any = DEFAULT_LOG_LEVEL) -> Any:
 
         return DEFAULT_LOG_LEVEL
 
+    def add_syslog_handler(
+        self,
```
mvesin, line 593: Nice point: I could test logging with syslog (127.0.0.1 UDP 514) on Linux Ubuntu - this works fine.

Issue: the current implementation logs **all log entries** in syslog, not only security logs.

From #1661 we want to add it for security logs only - and IMO it's enough for now.

Yet, if you would like to implement it for all logs, it would need to be via 2 different configuration entries eg

```
[syslog]
enable_sec = True
host_sec = 1.2.3.4
port_sec = 514
...
enable_log = True
host_log = localhost
port_log = 514
...
```

mvesin (commented):
Nice work @alitolga !

I added comments for a few points

Main ones are: (1) log only security logs in syslog (2) version/migration of config file

fedbiomed/common/config.py
srcansiz: +1

udp_server.py
srcansiz: I also would propose to exclude it.

alitolga: Sorry, I forgot to remove it. I agree to not merge it as well.

fedbiomed/common/config.py
alitolga: Hi @mvesin ,

Thanks for the comment ! I updated the versions and the migrate functions in the commit.

fedbiomed/common/logger.py
alitolga: Hi @mvesin ,

As we discussed, I am adding syslog logging only for the security logs for now.

alitolga:
Hi @mvesin ,

Thank you for the review !

I applied your suggested comments.

mvesin:
hi @alitolga 
I rebased on current develop & pushed to review a branch in sync with develop :-)

fedbiomed/node/config.py
```diff
@@ -111,6 +111,20 @@ def migrate(self):
 
             self._cfg["security"].update({"allow_federated_analytics": "True"})
 
+        if not self._cfg.has_option("syslog", "enable"):
```
mvesin: Hi @alitolga 
I recommend doing a few manual test after implementation :-)

This does not work because "syslog" section does not exist in file so you have to adapt like

```
if not self._cfg.has_section("syslog"):
...
  self._cfg.add_section("syslog")
```

mvesin (commented):
Thank you @alitolga !

- migration is not working yet
- also you probably need to adapt and complete unit tests - sorry for missing that at first review

Have a nice WE !

fedbiomed/node/config.py
alitolga: Hi @mvesin ,

Sorry for the oversight! I am pushing the fix now.

mvesin (approved):
Nice work @alitolga !
Now PR is ready for merging.

I rebased on current develop & will merge when tests have passed.

PR #1680, opened 2026-03-24
addition of minimal number of samples to dataset_manager and FA

### LINKED ISSUE #1679: Ensure minimum number of samples for Datasets in general
- Ensure minimum number of samples for datasets and ensure is used for analytics too

leonard2103:
This PR implements a minimum number of samples. This implementation concerns the addition of new datasets in the node and the consumption in FA operations. However, further changes are required to ensure minimum size of pre-existing dataset entries consumed in training.

closes #1679 

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/fedbiomed/common/dataset/_custom_dataset.py
+++ b/fedbiomed/common/dataset/_custom_dataset.py
@@ -19,52 +19,36 @@ class CustomDataset(Dataset):
     """
 
     def __init_subclass__(cls, **kwargs):
-        """Ensures that subclasses implement required methods"""
+        """Prevents subclasses from overriding reserved methods and enforces required ones."""
         super().__init_subclass__(**kwargs)
 
-        # Ensure __getitem__ is not overwritten
-        if (
-            "__getitem__" in cls.__dict__
-            and cls.__dict__["__getitem__"] is not CustomDataset.__getitem__
-        ):
+        if "__getitem__" in cls.__dict__:
             raise FedbiomedError(
                 "Overriding __getitem__ in CustomDataset subclasses is not allowed. "
                 "Please overwrite get_item instead."
             )
 
-        if (
-            "__init__" in cls.__dict__
-            and cls.__dict__["__init__"] is not CustomDataset.__init__
-        ):
+        if "__init__" in cls.__dict__:
             raise FedbiomedError(
                 "Overriding __init__ in CustomDataset subclasses is not allowed. "
                 "Please overwrite `read` method instead. Using path attribute to "
                 "access dataset location."
             )
 
-        # Ensure read and get_item are implemented by subclass
-        if "read" not in cls.__dict__ or not callable(cls.__dict__.get("read", None)):
-            raise FedbiomedError(
-                "CustomDataset subclass must implement a 'read' method. "
-                "This method is required to load the dataset and must be defined in your subclass."
+        for method, label in [
+            ("read", "read"),
+            ("get_item", "get_item"),
+            ("__len__", "__len__"),
+        ]:
+            implemented = any(
+                method in base.__dict__
+                for base in cls.__mro__
+                if base is not CustomDataset and base not in CustomDataset.__mro__
             )
-        if "get_item" not in cls.__dict__ or not callable(
-            cls.__dict__.get("get_item", None)
-        ):
-            raise FedbiomedError(
-                "CustomDataset subclass must implement a 'get_item' method. "
-                "This method is required to retrieve samples and must be defined in your subclass."
-            )
-
-        # Ensure __len__ is implemented by subclass
-        if "__len__" not in cls.__dict__ or not callable(
-            cls.__dict__.get("__len__", None)
-        ):
-            raise FedbiomedError(
-                "CustomDataset subclass must implement a '__len__' method. "
-                "This method is required to return the number of samples and must be "
-                "defined in your subclass."
-            )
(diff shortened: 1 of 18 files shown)


fedbiomed/common/dataset/_custom_dataset.py
```diff
@@ -19,52 +19,36 @@ class CustomDataset(Dataset):
...
-            cls.__dict__.get("__len__", None)
-        ):
-            raise FedbiomedError(
-                "CustomDataset subclass must implement a '__len__' method. "
-                "This method is required to return the number of samples and must be "
-                "defined in your subclass."
+        for method, label in [
+            ("read", "read"),
+            ("get_item", "get_item"),
+            ("__len__", "__len__"),
```
srcansiz, line 41: Hi @leonard2103, 

Why  use tuple for same method and label `('read, 'read')`, and not just tuple of `('read', 'get_item', '__len__'`)` ?

fedbiomed/node/jobs/_fa_job.py
```diff
@@ -153,6 +153,13 @@ def _build_dataset(
                 f"Cannot initialize dataset on node='{self._node_id}': {repr(e)}"
             ) from e
 
+        try:
+            self._dataset_manager.validate_samples(len(dataset))
+        except FedbiomedError as e:
+            raise _InternalJobError(
+                f"Dataset '{self._dataset_id}' does not meet minimum sample requirement: {repr(e)}"
+            ) from e
```
srcansiz, line 161: Hi @leonard2103, 

As I remember we wanted to apply same rule for trainings. However, sample size is only checked for FA. Correct me if sample size not intended to be used for training as well.

fedbiomed/common/dataset_controller/_medical_folder_controller.py
```diff
@@ -64,7 +64,7 @@ def __init__(
             FedbiomedError:
             - if one in `tabular_file` and `index_col` is given and the other is not
         """
-        DataLoadingPlanMixin.__init__(self)
+        super().__init__()  # initialises DataLoadingPlanMixin (self._dlp = None)
```
srcansiz, line 67: is calling `super().__init__()` needed? Isn't already called by default in class inheritance.

fedbiomed/node/config.py
```diff
@@ -60,6 +60,7 @@ def add_parameters(self):
             "allow_federated_analytics": os.getenv(
                 "FBM_SECURITY_ALLOW_FEDERATED_ANALYTICS", "True"
             ),
+            "minimum_samples": os.getenv("FBM_SECURITY_MINIMUM_SAMPLES", "0"),
```
srcansiz, line 63: I think this variable should `1` no? you can not do anything with 0 sample. Additionally, it is better to check if given value from ENV can be converted to integer. If it is not integer(able) it may raise expected errors.

fedbiomed/node/dataset_manager/_dataset_manager.py
```diff
@@ -31,13 +31,35 @@ class DatasetManager:
...
+    def validate_samples(self, n_samples: int) -> None:
+        """Raise FedbiomedError if n_samples is below the configured minimum.
+
+        Args:
+            n_samples: Number of samples in the dataset to validate.
+
+        Raises:
+            FedbiomedError: If n_samples is below the configured minimum.
+        """
+        if self._min_samples > 0 and n_samples < self._min_samples:
```
srcansiz, line 57: following my other comment concerning making min default samples `1`, it will be enough to check if `n_sample >= self._min_samples`.

fedbiomed/node/dataset_manager/_dataset_manager.py
```diff
@@ -31,13 +31,35 @@ class DatasetManager:
...
+            min_samples: Minimum number of samples required when adding a dataset.
+                Defaults to 0 (no minimum enforced).
+        """
+        self._min_samples = min_samples
         self._dataset_table = DatasetTable(path)
         self._dynamic_dataset_table = DynamicDatasetTable(path)
         self._dlp_table = DlpTable(path)
         self._dlb_table = DlbTable(path)
 
+    def validate_samples(self, n_samples: int) -> None:
```
srcansiz, line 48: Is there any reason that you implemented this method in `DatasetManager` and not in `DataManager`. Because `DataManager` is used for both training, testing, fa. It creates final dataset that will be used for data operations. I would recommend to implement it in `DataManager` if there is no specific reason to implement in `DatasetManager`.

srcansiz: Ok, after carefully checking, I  see that `DataManager` is not used in FAJob. So, for now, it may make sense to keep in `DatasetManager` and call from `Round` or `FAJob` after dataset instantiated. However,  once minimum sample size is checked for training we should keep in mind that we have to check both for testing and training dataset.

srcansiz (commented):
Hi @leonard2103, 

Thank for the PR. I have a few minor comment. And one comment about implementing minimum sample size for training and testing as well. I thought that it what we decided to do after the discussion on discord.

fedbiomed/node/config.py
srcansiz: Anyway, 1 does not make sense. Keeping it 0 to indicate "no-limit" is  much better. Sorry for the confusion.

Issue #1628, opened 2026-02-18
Implement support for locally-persistent parameters in Round class


sharkovsky:
This issue implements the "backend" for the feature of locally-persistent parameters.

The implementation supports two kinds of new behaviours, in addition to the default behaviour of aggregating all parameters. It allows users to define `"local"` parameters that are not shared with the aggregator, but are not necessarily saved between rounds, and `"persistent"` parameters that are saved between rounds, but may or may not be shared with the aggregation server.

The implementation leverages the node state manager to save persistent parameters. 

The general implementation flow is described in this [design](https://github.com/fedbiomed/fedbiomed/issues/1620#issuecomment-3919676066).

The implementation requires:

- [ ] Extend the Node State with a `model_state_%s_%s` file. 
- [ ] Implement loading and saving of `"persistent"` parameters in Round class (in `load_round_state` and `save_round_state` functions)
- [ ] Implement a `Round.reconcile_round_state` function, or equivalent, to handle the integration of aggregated params, local-only params (that need to be reinitialized) and persistent params (that have been loaded from state)
- [ ] Strip `"local"` parameters from the set of parameters to be aggregated inside the `after_training_params` function.

Issue #1627, opened 2026-02-18
Implement support for locally-persistent parameters in TrainingPlan


sharkovsky:
This issue implements the Researcher API for defining locally-persistent parameters. 

The implementation is intended to be extensible to new types of parameter behaviours in the future, which is why it creates the concept of a `parameter_tag` that defines its behaviour. 
The tags that are considered valid for now are:
- `"persistent"`, meaning the parameter values should be stored locally at the end of a round, and reloaded at the beginning of the next round
- `"local"`, meaning the parameter values should not be communicated to the aggregator
- Implicitly, we also define a global tag for parameters that are normally aggregated and not persisted locally. But this should be the default behaviour and we should not explicitly define this tag.

The implementation requires

- [ ] Adding a `tag_parameters(self, name: str) -> Set[str]` function to the interface defined in `BaseTrainingPlan`, with a default implementation that returns an empty set. The docstrings and semantics can be inspired by this [design](https://github.com/fedbiomed/fedbiomed/issues/1620#issuecomment-3919457990)

Issue #1620, opened 2026-02-13
Design feature allowing model classes to store some parameters locally instead of aggregating them


sharkovsky:
- [x] review the `exclude_buffers` functionality of `TorchModel` to see if and how it is relevant
- [x] review how node state works and whether it would be easy to use it for storing the values of the "local-only" parameters
- [x] design the researcher API: how can the researcher specify if and which parameters are "local-only", and how can they specify whether they should be stored on the node, or they can simply be reinitialized at every round (or other behaviours?)
- [x] design integration with aggregator: if we only send the shared parameters, then as Michelle said we don't need to create a custom aggregator. But potentially there could be other strategies that we may wish to explore.
- [ ] design implementation. Make sure that it can be applied to both torch and sklearn models.

sharkovsky:
`exclude_buffers` can be used to exclude a whole set of mode weights that are explictly handled differently by Pytorch, with the idea that they can be destroyed and reinitialized at every round. 

In the proposed new functionality, we would allow selecting parameters by name (and not by whether they are buffers or not), and we would retain their value across rounds. The new functionality is therefore an extension of `exclude_buffers`, which in theory becomes obsolete. 

However, since buffers are a specific concept in Pytorch and allows avoiding having to write all the names of the target buffers, I would suggest to maintain both functionalities.

sharkovsky:
## Researcher API

I am struggling between these ideas:
1. adding an argument to `model_args` or `training_args`
2. making this configurable directly in the model
3. An optional `list_local_only_parameters(self) -> List[str]` function in the `TrainingPlan` that returns a list of parameter names. 

### Adding an `*args` 

Pros: 
- consistent with `exclude_buffers`
- simple to use

Cons: 
- `*args` are already bloated
- requires nicer documentation
- implementation is "far away" from its usage

### extending the `Model` class

example: 
```python
experiment.training_plan().model().set_local_only_params(['conv1.weight', 'conv2.weight'])
```

Pros: 
- allows researcher more flexibility (e.g. user could provide a lambda function to identify params based on their name)
- implementation is closer to its usage

Cons:
- with current code, it could be very difficult because the `Model` class is very hidden to the user

### Optional `list_local_only_parameters` function in `TrainingPlan`

Pros: 
- flexibility

Cons:
- it could be dangerous to condition some logic on whether an attribute exists or not

sharkovsky:
ChatGPT likes the third option, and suggests a more generic

```python
[...]

Issue #1618, opened 2026-02-13
Add compatibility with models with node-specific parameters that should not be aggregated


michellewang:
I have a PyTorch [disease progression model](https://github.com/neurodatascience/fl-prog/blob/e24fc587ecc8e011a5e01e19a6cf311b948f2da8/fl_prog/model.py) that required a few hacks to work with Fed-BioMed.

The model is particular in that one of the parameters (`time_shifts`) is node-specific (there is one per subject, and the subjects are different between nodes), so it should not be aggregated by Fed-BioMed, and should be preserved on the node side between training rounds.

The changes @sharkovsky, @lchambon and I have had to make are:
1. Create a custom aggregator class that ignores some params: https://github.com/neurodatascience/fl-prog/blob/e24fc587ecc8e011a5e01e19a6cf311b948f2da8/fl_prog/aggregator.py
   1. This is only needed because, in this implementation, the node sends all parameters (even the ones that should not be aggregated) to the researcher. If those parameters were not sent, then `FedAverage` could have been used as usual
1. Override `_configure_model_and_optimizer` in the training class to use a custom variant of `TorchModel` which writes and loads the local params to disk in `get_weights` and `set_weights` respectively: https://github.com/neurodatascience/fl-prog/blob/e24fc587ecc8e011a5e01e19a6cf311b948f2da8/fl_prog/training_plan.py#L26-L60
   1. The main drawback with this method (beyond that it is overriding a private method) is that the file is always written to the current working directory. This could cause problems in cases where multiple nodes are running on the same machine, and also makes it hard to clean up leftover `weights.pth` files if they are left anywhere on the filesystem
1. Model initialization needs node-specific information (namely the number of subjects in the dataset). The workaround was to store that number in a config file for every node: https://github.com/neurodatascience/fl-prog/blob/e24fc587ecc8e011a5e01e19a6cf311b948f2da8/fl_prog/training_plan.py#L87-L92 (+ some extra handling of model initialization on the researcher side)
   1. This has the same limitation as the point above regarding current working directories
   1. This would be completely solved by #1617

From prior discussions, it seemed like the node state could be leveraged to handle the second point. For my use-case, the number of params is not very large since there is only one time-shift per subject, but of course it's possible to have use-cases where this is not the case.

PR #1681, opened 2026-03-24
Feature/1618 node specific parameters

lchambon:
**PR description**

This draft PR introduces a first attempt to address issue #1618 regarding node-specific parameters, implementing the design discussions from issues #1620, #1627 and #1628.

**Summary of changes**

Researcher can now tag parameters in a training plan as persistent and/or private:

- private: parameters that are never shared with the researcher (i.e., never aggregated).
- persistent: parameters that are saved locally on the node after each round.

In simple scenarios, this implementation works correctly for both Scikit-learn and PyTorch examples.
After fixing a small bug in breakpoint loading, breakpoints also appear to work correctly.

**Open questions & discussion points**

About Differential Privacy:

- Currently, OPACUS modifies model parameter names by adding a `"_module"` prefix, which breaks the current implementation because parameters cannot be loaded by the original names specified in `tag_parameters`.
- Should DP be applied to private parameters that are never shared to researcher? My intuition is no, but it’s unclear how this affects implementation. What are the consequences if the set of parameters subject to DP is smaller than expected
- While modifying the `after_training_params` function in `TorchTrainingPlan`, it seems that both SecAgg and DP cannot be applied simultaneously:

```
def after_training_params(self, flatten: bool = False) -> Dict[str, torch.Tensor]:
    ...

    # Run (optional) DP controller adjustments as well.
    params = self._dp_controller.after_training(params)
    if flatten:
        params = self._model.flatten(
            exclude_buffers=not self._share_persistent_buffers,
            private_params=self._private_params,
        )
    return params
 ```
 At least, DP seems to be ignored if SecAgg is applied in the `if flatten` statement.

About Declearn:

- How does the private parameter concept impact Declearn and more complex optimizers?
- Currently, using Scaffold triggers warnings, which require investigation.

About Secure Aggregation:

- The different use cases tested with secure aggregation (PyTorch and Scikit-learn) seem to work fine.
- However, the `secure_aggregation` function in `Aggregator` does not appear to be used anywhere. Should it still be updated to reflect the new parameter access logic, or can it be deprecated?

**Missing**

- [x] Implement unit tests.
- [x] Update documentation.
- [x] Create e2e test

### DIFF
--- a/docs/user-guide/researcher/training-plan.md
+++ b/docs/user-guide/researcher/training-plan.md
@@ -394,3 +394,102 @@ model_args = {
     }
 }
 ```
+
+## Advanced: tagging model parameters
+
+Fed-BioMed provides a mechanism to customize how individual model parameters are handled during federated training through the `tag_parameters` method. By default, all parameters follow the standard federated learning workflow: they are aggregated globally across nodes and updated each round from the researcher-aggregated values. Overriding `tag_parameters` allows you to exclude specific parameters out of this default behavior.
+
+### Available tags
+
+The `tag_parameters` method currently supports two tags:
+
+- **`local`**: The parameter is never shared with the researcher nor included in aggregation. It remains entirely local to the node.
+- **`persistent`**: The parameter is saved locally on the node at the end of each round.
+
+Tags can be combined: a parameter tagged as both `local` and `persistent` will be kept local *and* saved across rounds, and will be restored at the start of every round. A parameter tagged as `local` only (without `persistent`) is reset to its initial value at the start of every round.
+
+### Defining `tag_parameters`
+
+The `tag_parameters` method takes a parameter name as input and returns a set of tags. Return an empty set (the default) for standard federated behavior.
+
+#### PyTorch example
+
+```python
+import torch
+import torch.nn as nn
+from fedbiomed.common.training_plans import TorchTrainingPlan
+
+class MyTrainingPlan(TorchTrainingPlan):
+
+    class Net(nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.shared_layer = nn.Linear(128, 64)
+            self.local_bn = nn.BatchNorm1d(64)
+            self.local_bias = nn.Parameter(torch.zeros(64))
+
+        def forward(self, x):
+            x = self.shared_layer(x)
+            x = self.local_bn(x)
+            return x
+
+    def init_model(self, model_args):
+        return self.Net()
+
+    def tag_parameters(self, name):
+        if name == "local_bias":
+            return {"local", "persistent"}
+        if name == "local_bn":
+            return {"local"}
+        return set()
+
+    # ... rest of the training plan
+```
+
+#### Scikit-learn example
+
+```python
+from fedbiomed.common.training_plans import FedSGDClassifier
+
+class MyTrainingPlan(FedSGDClassifier):
+
+    def tag_parameters(self, name):
+        # Keep the intercept local and persistent across rounds
+        if name == "intercept_":
+            return {"local", "persistent"}
+        return set()
+
+    def training_data(self):
+        # returns a Fed-BioMed DataManager object
+        pass
+
+    def init_optimizer(self, optimizer_args):
+        pass
+
+    def init_dependencies(self):
+        pass
+```
+
+### What happens on the node side during training
+
+The way tagged parameters are handled is determined at the node level during each round, across three key steps.
+
+**At the start of a round**, the node reconstructs the full set of model parameters before loading them into the model. Parameters received from the researcher (i.e., the aggregated global parameters) serve as the base. The node then overrides specific entries depending on their tags:
+
+- Parameters tagged as both `local` and `persistent` are restored from the values saved at the end of the previous round. If no previous round exists (i.e., the first round), they are initialized from the model's initial weights as defined in `init_model`.
+- Parameters tagged as `local` only (without `persistent`) are always reinitialized from the model's initial weights at each round, regardless of any training that may have occurred in previous rounds.
+- All other parameters are taken as-is from the researcher-provided aggregated weights.
+
+**At the end of a round**, the node saves the current values of all parameters tagged as `persistent`. These saved values will be loaded and restored at the beginning of the next round.
+
+**When sending parameters back to the researcher**, parameters tagged as `local` are excluded from the payload. The researcher therefore never sees or aggregates them, and they play no role in the global model update.
+
+### Caution with advanced optimizers
+
+When using a `declearn`-based optimizer, the federated engine also exchanges optimizer auxiliary variables between the node and the researcher at each round (e.g., control variates for Scaffold). Fed-BioMed handles `local` parameters specifically in this exchange:
+
+- **When collecting auxiliary variables** to send back to the researcher (at the end of a round), auxiliary variables corresponding to `local` parameters are stripped from the payload. The researcher therefore never receives nor aggregates the optimizer state associated with local parameters.
+- **When processing auxiliary variables** received from the researcher (at the beginning of a round), the auxiliary variables for `local` parameters are reset to zero vectors on the node side. This prevents any cross-round accumulation of optimizer state for parameters that are not globally shared.
+
+!!! warning "Interaction with Declearn optimizers"
+    Using parameters tagged as `local` in combination with more complex federated optimizers such as those provided by `declearn` (e.g., Scaffold, or other stateful `OptiModule`-based optimizers) may lead to unexpected or incorrect behavior. These optimizers maintain internal auxiliary variables (e.g., gradient moments, control variates) that are tracked per parameter. When a parameter is tagged as `local`, its associated auxiliary variables are not shared nor aggregated. In practice, this may cause training instability or degraded convergence. It is therefore strongly recommended to use `local` parameters only with simple, stateless optimizers, unless you are fully aware of the implications for the optimizer state.
--- a/docs/user-guide/researcher/training-plan.md
+++ b/docs/user-guide/researcher/training-plan.md
@@ -470,7 +470,7 @@ class MyTrainingPlan(FedSGDClassifier):
         pass
 ```
 
-### What happens on the node side during training
+### Handling of parameters tagged as `local` or `persistent` during training
 
 The way tagged parameters are handled is determined at the node level during each round, across three key steps.
 
@@ -483,6 +483,11 @@ The way tagged parameters are handled is determined at the node level during eac
 **At the end of a round**, the node saves the current values of all parameters tagged as `persistent`. These saved values will be loaded and restored at the beginning of the next round.
 
 **When sending parameters back to the researcher**, parameters tagged as `local` are excluded from the payload. The researcher therefore never sees or aggregates them, and they play no role in the global model update.
+
+From the researcher's perspective, this has the following practical implications:
+
+- `local` parameters are absent from aggregation results: they will not appear in `experiment.aggregated_params()`, nor in the raw per-node replies accessible via `experiment.training_replies()`.
+- `local` parameters are still present in the researcher-side training plan: they remain accessible through `experiment.training_plan().get_model_params()` and `experiment.training_plan().model()`. However, these values originate from `init_model`: they reflect the initial weights as defined by the researcher, not the values that were actually trained and retained on each node. In particular, they do not correspond to the local parameter values that nodes will restore at the start of the next round (which, for parameters tagged as both `local` and `persistent`, are the node's own saved values from the previous round).
 
 ### Caution with advanced optimizers
 
--- a/fedbiomed/common/models/_sklearn.py
+++ b/fedbiomed/common/models/_sklearn.py
@@ -130,8 +130,10 @@ class BaseSkLearnModel(Model, metaclass=ABCMeta):
 
         unknown = local_params_set - set(self.param_list)
         if unknown:
-            raise FedbiomedModelError(
-                f"{ErrorNumbers.FB622.value}: Unknown local parameters: {unknown}"
+            logger.warning(
+                "'BaseSkLearnModel.get_weights' received unknown local parameters "
+                "that are not part of the model; unknown parameters: %s. ",
+                unknown,
             )
 
         # Gather copies of the model weights.
--- a/fedbiomed/common/models/_sklearn.py
+++ b/fedbiomed/common/models/_sklearn.py
@@ -102,7 +102,7 @@ class BaseSkLearnModel(Model, metaclass=ABCMeta):
         self,
         only_trainable: bool = False,
         exclude_buffers: bool = True,
-        private_params: Optional[List[str]] = None,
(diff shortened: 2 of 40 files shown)

### CODE REVIEW

**Comment by mvesin** [comment]
Great work @lchambon !

The open questions you raise are all very relevant and advanced points - may need some in depth team discussion :-)

My 2 cents at first read, HTH

> Should DP be applied to private parameters that are never shared to researcher? My intuition is no, but it’s unclear 
> how this affects implementation. What are the consequences if the set of parameters subject to DP is smaller than 
> expected

We need a DP expert on this :-)

Probably local DP and central DP are 2 different cases. My first intuition would be

1. in local DP: as it is applied in each batch training, the private parameters value would influence non-private parameters in next batch training - so it should be applied to avoid breaking model privacy.

AI assistants suggest that it is often not needed, as this case is often handled by DP applied to shared layers in next rounds. Leakage that may result from non-clipped gradients with DP are general issues, not introduced by private layers. Main risk would come from indirect leakage from the private layers (eg reuse local model for inference, distillation; client side attacks) which are probably out of our scope.

All in all, applying noise to all layers would be a reasonable first implementation choice (simple to implement and conservative) though sub optimal (noising private layers in many scenario where it is not needed).

More advanced (future) implementation may not noise private layers by default, with a flag to change behaviour.

2. in central DP: as it is applied after training, it's applying noise to params that won't be shared with the researcher. Yet, if these params are persistent it matters. In that case these params will influence non-private layers in next rounds. But then they will be subject to the local/central DP policy of the node. So it may not be necessary to apply central DP to the private layer. So clip + noise only shared layers is OK.

AI assistant suggest that noising private layers is probably not needed. Main risks are indirect leakage mentioned for local DP (probably out of our scope).

So we may choose not noise the private layer if simple enough to implement.
We may also noise them as first simple implementation (suboptimal utility) - and maybe document this in user documentation .
Or have optional param to do it in more advanced implementation.

>  While modifying the after_training_params function in TorchTrainingPlan, it seems that both SecAgg and DP cannot be applied simultaneously (at least, DP seems to be ignored if SecAgg is applied). Is this intended?

Sorry I could not find that. Can you point to the code where it happens to remind us ?

>  The different use cases tested with secure aggregation (PyTorch and Scikit-learn) seem to work fine.
>  However, the secure_aggregation function in Aggregator does not appear to be used anywhere. Should it still be 
> updated to reflect the new parameter access logic, or can it be deprecated?

Good point, I could not find any use of this function too.
@srcansiz had begun a refactor of secagg that could not be completed - maybe linked to this ?

HTH !

File: fedbiomed/node/round.py [hunk 1]
```diff
@@ -1093,3 +1168,61 @@ def _split_train_and_test_data(self, test_ratio: float = 0) -> DataManager:
...
+        Args:
+            initial_model_weights: initial values of all model parameters (before any training)
+            weights_from_researcher: parameters received from the researcher
+            persistent_model_weights: locally stored parameters saved from previous rounds
+
+        Returns:
+            the complete set of parameters to be loaded into the model before training
+        """
+        # Deal with the case where persistent_model_weights is None
+        persistent_model_weights = persistent_model_weights or {}
```
sharkovsky, line 1201: Handle edge cases where the weight itself evaluates to false (e.g. empty dict, but maybe the weight was actually an empty dict. Or the weight we a single scalar boolean set to False)

```suggestion
        persistent_model_weights = persistent_model_weights if persistent_model_weights is not None else {}
```

fedbiomed/node/round.py
```diff
@@ -1093,3 +1168,61 @@ def _split_train_and_test_data(self, test_ratio: float = 0) -> DataManager:
...
+
+        # Start from parameters received by researcher
+        full_params = dict(weights_from_researcher)
```
sharkovsky, line 1204: why is this needed? Isn't `weights_from_researcher` already a dict?

fedbiomed/node/round.py
```diff
@@ -1093,3 +1168,61 @@ def _split_train_and_test_data(self, test_ratio: float = 0) -> DataManager:
...
+        local_only = self.training_plan.filter_model_params_by_tags(
+            initial_model_weights,
+            required_tags={"local"},
+            forbidden_tags={"persistent"},
+        )
+
+        for name in local_only.keys():
+            full_params[name] = initial_model_weights[name]
+
+        return full_params
```
sharkovsky, line 1228: please DOUBLE-CHECK what I'm saying, but I think that all this code could be achieved with a one-liner

```suggestion
        return {**initial_model_weights, **persistent_model_weights, ****weights_from_researcher}
```

It should achieve the following behaviour:
- local + persistent parameters are overwritten with their persistent value
- persistent-only parameters are overwritten with their aggregated value
- automatic first-round handling: since the local+persistent parameters will not be in either the persistent or researcher dictionaries, they will automatically be randomly initialized

Example
```python
In [14]: weights_from_researcher = {'agg': 3, 'agg4': 4, 'perst': 1}
In [15]: persistent_model_weights = {'perst': 11, 'perst_local': 2}
In [16]: initial_model_weights = {'local': 0, 'perst': 11, 'perst_local': 222, 'agg': 33, 'agg4': 44}

In [17]: {**initial_model_weights, **persistent_model_weights, **weights_from_researcher}
Out[17]: {'local': 0, 'perst': 1, 'perst_local': 2, 'agg': 3, 'agg4': 4}
```

docs/user-guide/researcher/training-plan.md
```diff
@@ -393,4 +393,103 @@ model_args = {
...
+        pass
+
+    def init_optimizer(self, optimizer_args):
+        pass
+
+    def init_dependencies(self):
+        pass
+```
+
+### What happens on the node side during training
```
sharkovsky, line 473:
```suggestion
### Handling of parameters tagged as `local` or `persistent` during training
```

docs/user-guide/researcher/training-plan.md
```diff
@@ -393,4 +393,103 @@ model_args = {
...
+
+- Parameters tagged as both `local` and `persistent` are restored from the values saved at the end of the previous round. If no previous round exists (i.e., the first round), they are initialized from the model's initial weights as defined in `init_model`.
+- Parameters tagged as `local` only (without `persistent`) are always reinitialized from the model's initial weights at each round, regardless of any training that may have occurred in previous rounds.
+- All other parameters are taken as-is from the researcher-provided aggregated weights.
+
+**At the end of a round**, the node saves the current values of all parameters tagged as `persistent`. These saved values will be loaded and restored at the beginning of the next round.
+
+**When sending parameters back to the researcher**, parameters tagged as `local` are excluded from the payload. The researcher therefore never sees or aggregates them, and they play no role in the global model update.
+
```
sharkovsky, line 486: Please add some explanation for the researcher:
- the `local` parameters are not available in the `experiment.aggregated_params()`, nor in the `training_replies`
- however, the `local` parameters are still there in `experiment.training_plan().get_model_params()` and inside `experiment.training_plan().model()`. But clarify that these values are taken from `init_model`, and do not correspond to what the node will see in the next round.

fedbiomed/common/models/_sklearn.py
```diff
@@ -121,10 +125,21 @@ def get_weights(
                 f"{ErrorNumbers.FB622.value}. Attribute `param_list` is empty. You should "
                 f"have initialized the model beforehand (try calling `set_init_params`)"
             )
+
+        local_params_set = set(local_params) if local_params else set()
+
+        unknown = local_params_set - set(self.param_list)
+        if unknown:
+            raise FedbiomedModelError(
```
sharkovsky, line 133: it seems harsh to raise an exception here. 
Maybe simply log a warning, since it doesn't seem to me like the rest of the execution would fail if there are unknown parameters (but maybe I'm wrong, in which case it makes sense to raise an exception here)

fedbiomed/common/models/_torch.py
```diff
@@ -66,25 +69,39 @@ def get_weights(
...
+
+        param_iterator = list(
             self.model.named_parameters()
             if exclude_buffers
             else self.model.state_dict().items()
         )
+
+        unknown = local_params_set - {name for name, _ in param_iterator}
+        if unknown:
+            raise FedbiomedModelError(
```
sharkovsky, line 88: same comment as before for exception

fedbiomed/common/training_plans/_base_training_plan.py
```diff
@@ -671,6 +698,117 @@ def evaluate(data, target):
...
+        Args:
+            name: The name of the parameter whose tags are being retrieved.
+
+        Returns:
+            The validated set of tags associated with the parameter.
+        """
+        tags = self.tag_parameters(name)
+        return self._validate_parameter_tags(name, tags)
+
+    def filter_model_params_by_tags(
```
sharkovsky, line 773: this should be a static method

sharkovsky:
Great work @lchambon !

I've added some review comments, when fixed I think this is ready to be merged. 
I confirm the PR implements the intended behaviour.

I have a general comment about the implementation. Currently the interface for handling model parameters through the training plan (and through the underlying Model object) is becoming more and mode complex. 
It was already quite complex with all the `only_trainable` and `exclude_buffers` logic, but now with `filter_model_params_by_tags` it becomes necessary to think about whether we need to filter the model parameters every time we use them. 

The current implementation follows what was done before and fits within the rest of the state of affairs, however I think a better design and refactor is becoming overdue for this part. I will open a separate issue for that.

docs/user-guide/researcher/training-plan.md
lchambon: Thank you Francesco, I applied this change

docs/user-guide/researcher/training-plan.md
lchambon: Thank you Francesco, I added this paragraph:

```markdown
From the researcher's perspective, this has the following practical implications:

- `local` parameters are absent from aggregation results: they will not appear in `experiment.aggregated_params()`, nor in the raw per-node replies accessible via `experiment.training_replies()`.
- `local` parameters are still present in the researcher-side training plan: they remain accessible through `experiment.training_plan().get_model_params()` and `experiment.training_plan().model()`. However, these values originate from `init_model`: they reflect the initial weights as defined by the researcher, not the values that were actually trained and retained on each node. In particular, they do not correspond to the local parameter values that nodes will restore at the start of the next round (which, for parameters tagged as both `local` and `persistent`, are the node's own saved values from the previous round).
```

fedbiomed/common/models/_sklearn.py
lchambon: You are totally right. As suggested, I changed to log a warning instead:

```python
if unknown:
    logger.warning(
        "'BaseSkLearnModel.get_weights' received unknown local parameters "
        "that are not part of the model; unknown parameters: %s. ",
        unknown,
    )
```

fedbiomed/common/models/_torch.py
lchambon: Same for _sklearn, I changed to log a warning instead:

```python
if unknown:
    logger.warning(
        "'TorchModel.get_weights' received unknown local parameters "
        "that are not part of the model; unknown parameters: %s. ",
        unknown,
    )
```

fedbiomed/node/round.py
lchambon: Thank you Francesco, you are totally right ! I did not think about it. I double-checked and this works perfectly fine. 

I have replaced the function by:

```python
def _reconstruct_full_params(
    self,
    initial_model_weights: Dict[str, Any],
    weights_from_researcher: Dict[str, Any],
    persistent_model_weights: Optional[Dict[str, Any]],
) -> Dict[str, Any]:
    """Reconstructs the full set of model parameters before training.

    This method merges parameters received from the researcher (typically
    aggregated global parameters) with locally managed parameters based on
    their associated tags ("local", "persistent").

    The reconstruction follows these rules:
        - Parameters tagged as both "local" and "persistent" are restored
        from `persistent_model_weights` if available. If not (e.g., first
        round), they are initialized from `initial_model_weights`.
        - Parameters tagged as "local" but not "persistent" are always
        reinitialized from `initial_model_weights` at each round.
        - All other parameters are kept as provided in `weights_from_researcher`.

    Args:
        initial_model_weights: initial values of all model parameters (before any training)
        weights_from_researcher: parameters received from the researcher
        persistent_model_weights: locally stored parameters saved from previous rounds

    Returns:
        the complete set of parameters to be loaded into the model before training
    """
    # Deal with the case where persistent_model_weights is None
    persistent_model_weights = (
        persistent_model_weights if persistent_model_weights is not None else {}
        )

    return {**initial_model_weights, **persistent_model_weights, **weights_from_researcher}
```

fedbiomed/node/round.py
lchambon: It was to create a shallow copy of `weights_from_researcher`. But it is useless now, as I simplified a lot the function thanks to your other comment.

fedbiomed/node/round.py
lchambon: Thank you, you are right. I corrected as suggested.

fedbiomed/common/training_plans/_base_training_plan.py
lchambon: I am a bit confused about this comment.

Do you mean doing something like this:

```python

@staticmethod
def filter_model_params_by_tags(
    params: Dict[str, Any],
    get_tags: Callable[[str], Set[str]],
    required_tags: Optional[Set[str]] = None,
    forbidden_tags: Optional[Set[str]] = None,
) -> Dict[str, Any]:
    """Filters a given parameter dict based on required or forbidden tags.

    Args:
        params: Dictionary of model parameters (e.g. output of get_model_params)
        get_tags: Function that returns tags for a given parameter name
        required_tags: Set of tags that parameters must include (AND logic)
        forbidden_tags: Set of tags that parameters must NOT include

    Returns:
        Filtered dictionary of parameters
    """
    filtered_params = {}

    required_tags = required_tags or set()
    forbidden_tags = forbidden_tags or set()

    if not isinstance(required_tags, set) or not isinstance(forbidden_tags, set):
        raise FedbiomedTrainingPlanError(
            f"{ErrorNumbers.FB605}: required_tags and forbidden_tags must be sets"
        )

    for name, value in params.items():
        tags = get_tags(name)

        if required_tags and not required_tags.issubset(tags):
            continue

        if forbidden_tags and not tags.isdisjoint(forbidden_tags):
            continue

        filtered_params[name] = value

    return filtered_params
````

And then using in `round.py` like this:

```python
BaseTrainingPlan.filter_model_params_by_tags(...)
```

in order to separate how to get tags from how to filter parameters ?

lchambon:
Thank you @sharkovsky for your review. I applied your suggested changes, except for one comment I’d like to discuss with you, if you don’t mind.
I also added end2end tests if you want to review them.

fedbiomed/common/training_plans/_base_training_plan.py
sharkovsky: Ah, forget what I said. I think I missed the fact that you were using `self.get_parameter_tags` inside the function. Forget what I said!

PR #1698, opened 2026-04-07
node/cli

leonard2103:
Series of modifications to comply with mypy. This PR concerns `node/cli` files

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/fedbiomed/node/cli_utils/_database.py
+++ b/fedbiomed/node/cli_utils/_database.py
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import os
-import warnings
 from pathlib import Path
 from typing import List, Optional, Union
 
@@ -11,7 +10,11 @@ from fedbiomed.common.exceptions import (
     FedbiomedDatasetManagerError,
 )
 from fedbiomed.common.logger import logger
-from fedbiomed.node.cli_utils._io import validated_data_type_input, validated_path_input
+from fedbiomed.node.cli_utils._io import (
+    _prompt_path_cli,
+    validated_data_type_input,
+    validated_path_input,
+)
 from fedbiomed.node.cli_utils._medical_folder_dataset import (
     add_medical_folder_dataset_from_cli,
 )
@@ -71,10 +74,8 @@ def add_database(
     """
     data_loading_plan = None
 
-    # if all args are provided, just try to load the data
-    # if not, ask the user more information
     need_interactive_input = (
-        interactive is True
+        interactive
         or path is None
         or name is None
         or tags is None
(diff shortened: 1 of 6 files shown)


fedbiomed/node/cli_utils/_database.py
```diff
@@ -168,10 +169,12 @@ def add_database(
             data_type = "default"
 
         # Validate path
-        if not os.path.exists(path):
+        if path is not None and not os.path.exists(path):
             logger.critical("provided path does not exists: " + path)
 
     # Ensure path is absolute
+    if path is None:
+        raise FedbiomedDatasetManagerError("Dataset path is not set")
```
alitolga: There are multiple scenarios where the path input can given differently (from Tkinter, from file or directly from input).

Instead of checking if the path is valid per case, I think we can just do the check at the very end.

alitolga: I just tested, and if you exit tkinter without giving a path, it doesn't return None, but an empty tuple. And it crashes the next line when trying to convert it to an absolute path.

fedbiomed/node/cli_utils/_tkinter_utils.py
```diff
@@ -2,33 +2,33 @@
...
+# None if tkinter is unavailable on this platform
+filedialog: Optional[ModuleType] = None
+messagebox: Optional[ModuleType] = None
+available: bool = False
+
 try:
-    import tkinter.filedialog
-    import tkinter.messagebox
-    from tkinter import _tkinter
+    import tkinter._tkinter as _tkinter_internal
```
alitolga: This import is incorrect and crashes the tkinter.

And also I think the custom class definition for TclError should be outside the try block. So, I believe it can be something like this:

```python
from tkinter import TclError

class FedbiomedTclError(TclError):  # type: ignore[no-redef]
    """Raised in place of tkinter. TclError when tkinter is not installed."""
    pass
```

and below

``` python

except ImportError:
    raise FedbiomedTclError("Import Error: Tkinter is not available.")

except Exception as e:
    # Catch any other exception that may occur during import (e.g., on unsupported platforms)
    raise FedbiomedTclError(f"Unexpected exception caught: Tkinter is not available: {e}") from e
```

fedbiomed/node/cli_utils/_io.py
```diff
@@ -40,86 +40,75 @@ def validated_data_type_input() -> str:
...
-        logger.debug(path)
-
-    except (RuntimeError, TclError):
-        path = None
-        error_msg = "[ERROR] GUI failed. Falling back to CLI"
-        messagebox.showerror(title="Error", message=error_msg)
-
-    return path
+            logger.warning(empty_msg)
+        return path
```
alitolga, line 66: The same comment for here. We should check if the user selected a valid path from tkinter; it can be either here, or in the function above.

alitolga (changes requested):

fedbiomed/node/cli_utils/_database.py
```diff
@@ -186,19 +170,18 @@ def add_database(
             data_loading_plan=data_loading_plan,
         )
     except (AssertionError, FedbiomedDatasetManagerError) as e:
-        if interactive is True and messagebox is not None:
+        if interactive and messagebox is not None:
             messagebox.showwarning(title="Warning", message=str(e))
         else:
-            warnings.warn(f"[ERROR]: {e}", stacklevel=1)
+            logger.error(str(e))
```
alitolga, line 208: I forgot to mention these, thanks for converting these messages to use the logger !

alitolga (commented):
Thanks for the PR @leonard2103 !

I added small checks to prevent empty tags, name and description when adding datasets. If it is okay with you as well, please feel free to merge.

alitolga (approved):
Thanks for the PR again @leonard2103 . Looks great to me !

PR #1704, opened 2026-04-07
mypy_series_07

leonard2103:
Efforts to implement mypy. This PR includes changes to files in `fedbiomed/common/` including: `validator`, `training_args` and `data_loading_plan`

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/fedbiomed/common/dataloadingplan/_data_loading_plan.py
+++ b/fedbiomed/common/dataloadingplan/_data_loading_plan.py
@@ -4,7 +4,7 @@
 import uuid
 from abc import ABC, abstractmethod
 from importlib import import_module
-from typing import Any, Dict, List, Optional, Tuple, Type, TypeVar, Union
+from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
 from fedbiomed.common.constants import DataLoadingBlockTypes, DatasetTypes, ErrorNumbers
 from fedbiomed.common.exceptions import (
@@ -22,9 +22,6 @@ from fedbiomed.common.validator import (
     ValidatorError,
     validator_decorator,
 )
-
-TDataLoadingPlan = TypeVar("TDataLoadingPlan", bound="DataLoadingPlan")
-TDataLoadingBlock = TypeVar("TDataLoadingBlock", bound="DataLoadingBlock")
 
 
 class SerializationValidation:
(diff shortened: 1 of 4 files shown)


fedbiomed/common/dataloadingplan/_data_loading_plan.py
```diff
@@ -171,9 +168,13 @@ def _target_dataset_type_validator(
     @staticmethod
     @validator_decorator
     def _loading_blocks_types_validator(
-        loading_blocks: dict,
+        loading_blocks: object,
```
srcansiz, line 171: Hi @leonard2103, 

Why it is object and no dict?

fedbiomed/common/dataloadingplan/_data_loading_plan.py
```diff
@@ -195,9 +196,13 @@ def _loading_blocks_types_validator(
     @staticmethod
     @validator_decorator
     def _key_paths_validator(
-        key_paths: Dict[DataLoadingBlockTypes, Union[tuple, list]],
+        key_paths: object,
```
srcansiz, line 199: Why not using TypeVar instead of `object` object is too general everything is object in Python.

fedbiomed/common/dataloadingplan/_data_loading_plan.py
leonard2103: Most of the time is because the variables are not strictly treated as the expected type. One thing is the expected and another the possible input. In internal exchanges there should be no difference. However, when we cannot control the input, it is possible that the variable is not what it is expected (like a dict). In the logic there are parts of code that validate (`isinstance(val, type)`). mypy is more strict about that and `object` basically indicates that we care about the type but at that point it could be different things.

fedbiomed/common/dataloadingplan/_data_loading_plan.py
leonard2103: Same as above.

fedbiomed/common/dataloadingplan/_data_loading_plan.py
srcansiz: Ok, what I understand that this method is for validating the type so it make sense that it takes `object` type and because it can be anything ?  If it is the case, why don't we use `Any` ?

leonard2103: Object and Any somehow behave in the same way but there is a subtle difference. Any basically lets you do anything and does not raise a problem so for example doing `x.keys()` just after would not complain. With object it is necessary to have `isinstance(x, dict)` just before.

PR #1735, opened 2026-04-29
Feature/1734 fix fa dtype bug

### LINKED ISSUE #1734: [BUG] Federated Analytics Mean TypeError ufunc 'isfinite' not supported for the input types
I get this error when computing a federated analytics mean
```
2026-04-29 12:00:50,423 fedbiomed ERROR - Node NODE_fb410134-6d61-41de-8c49-4a706448b853 analytics error [FB325: Node federated analytics error]: Error during execution of analytics '['mean']' on node='NODE_fb410134-6d61-41de-8c49-4a706448b853': TypeError("ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''")
```

After investigation, I found that the issue happens because the dtype of the numpy array used during calculation of the mean is `object` instead of some numeric type.

Specifically, this happens when:
1. the original dataset contains both numeric and string (categorical) variables
2. I use `exp.analytics.mean(dataset_schema=<numeric only>)` to select the numeric variables only
3. Optionally, I have tried also subsetting the `input_columns` of `TabularDataset`, but it doesn't change anything
4. When the `AnalyticsOrchestrator` is in the `Iterate and Accumulate` phase, it first loads the full sample, and only after that it slices based on the columns set by the schema. This means that the returned numpy array inherits the generic `object` dtype of the full sample.

sharkovsky:
**PR description**
Fixes #1734.

The implementation relies on attempting to convert all columns to floating point type, which assumes that all columns were numeric. This should be a reasonable assumption since we're asking for `mean`. It raises a nicely interpretable `Value Error: could not convert string to float` in case the user requested the computing the mean on a column with strings. 

Note that this PR only fixes the issue for `mean`, but the problem may still persist in case of `count`, although the fix is harder there because we cannot make the assumption that the columns should be converted to float. 

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:

### DIFF
--- a/fedbiomed/common/analytics/accumulators/_operations.py
+++ b/fedbiomed/common/analytics/accumulators/_operations.py
@@ -65,11 +65,11 @@ class MeanAccumulator(BaseStatAccumulator):
         self._validate_shape(val)
 
         # Sum (replace non-finite with 0, initialize as float32)
-        zeroed = np.nan_to_num(val, nan=0.0, posinf=0.0, neginf=0.0).astype(np.float32)
+        zeroed = np.nan_to_num(val.astype(float), nan=0.0, posinf=0.0, neginf=0.0).astype(np.float32)
         self.sum_val = zeroed if self.sum_val is None else self.sum_val + zeroed
 
         # Count (finite values only)
-        increment = np.isfinite(val).astype(np.int32)
+        increment = np.isfinite(val.astype(float)).astype(np.int32)
         self.counts = increment if self.counts is None else self.counts + increment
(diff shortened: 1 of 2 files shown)


sharkovsky:
After discussion with @leonard2103, I will close this PR since the current assumption is that datasets are ready-to-use, and transformations such as OneHotEncoding are not necessarily supported by Fed-BioMed. So we need a deeper discussion on how to handle such cases.

PR #1743, opened 2026-05-06
Fed-BioMed Node - restful node api for Node management operations

### LINKED ISSUE #1717: Creation of NodeProcessManager class and Integration in CLI
This issue aims to delegate the responsibility of starting the Federated Node process from the CLI to a separate class.

### LINKED ISSUE #1718: Add State and User Info to the Database

srcansiz:
Closes #1717 and #1718 

Changes made: 

- Add Node Process Manager class. The class controls the federation-node-process lifecycle with start/stop/restart methods.
- Modify CLI startup to use NodePM class.
- Add different Node Process States starting/running/stopping etc. to keep track of the process state.
- Add a Node Process State Table and Node Process History Table to db.py class to write current and ancient process state history.
- Add set_state to various locations during the process lifecycle to write the info to the db. Table for current state upserts while the history table appends/inserts.
- Add unit tests.

What is being logged into the database: 

- node_id
- node_name
- state (Starting/Running/Stopping/Stopped/Unknown)
- pid
- last_action
- last_reason
- actor (Source (Local or GUI) / local_username or user_id / role etc.)
- updated_at
- started_at
- stopped_at
- exit_code (if available)


fedbiomed/node/dataset_manager/_db_tables.py
```diff
@@ -229,3 +229,37 @@ class DlbTable(BaseTable):
...
+class NodeProcessStateHistoryTable(TinyTableConnector):
+    """Append-only database table for node process state history."""
+
+    _table_name = "NodeProcessStateHistory"
+    _id_name = "pid"
+
+    def insert(self, entry: dict) -> int:
+        """Insert a history entry, allowing multiple entries for the same pid."""
+        if not isinstance(entry, dict):
+            raise TypeError(f"Expected entry to be dict, got {type(entry)}.")
```
srcansiz: Hi @alitolga,

I think you can remove this check here. This is an internal use code. We can expect user to interact with it os if `dict` is passed it is probably developer mistake and bug that should be detected before PR.

fedbiomed/node/dataset_manager/_db_tables.py
```diff
@@ -229,3 +229,37 @@ class DlbTable(BaseTable):
...
+        if self._id_name not in entry:
+            raise ValueError(f"Entry must contain '{self._id_name}' field.")
```
srcansiz: Hi @alitolga,

I propose to use `dataclass` for the `entry` where you identify every field in the dict along with their types. Then, you can use this dataclass pass `dict` to validate its fields.

fedbiomed/node/node_pm.py
```diff
@@ -0,0 +1,284 @@
...
+                "state": state.value,
+                "pid": self._process.pid if self._process else None,
+                "action": action,
+                "reason": reason,
+                "actor": self._build_actor(actor),
+                "updated_at": now,
+                "started_at": existing.get("started_at"),
+                "stopped_at": existing.get("stopped_at"),
+                "exit_code": exit_code,
+            }
```
srcansiz: Hi @alitolga,

As I mentioned in the old comment, you can use `dataclass` for this entry.

fedbiomed/node/cli.py
```diff
@@ -498,45 +588,18 @@ def initialize(self):
...
+            fedbiomed_root=self._node.config.root,
+            node_root=self._node.config.root,
+            data_folder="",
+            host="localhost",
+            port="8484",
+            development=False,
+            key_file=None,
+            cert_file=None,
+            gui_debug=False,
+            node_args=_build_node_args(args),
```
srcansiz: Hi @alitolga, 

For this first version, I think it is better to define only node PM. Then, we can merge it in develop. It will be backward compatible and not break anything. Therefore, I propose to you `NodePM` here instead of `start_gui`. The CLI is going to work as it used to work before.  Node and GUI are started separately. This allow merging this branch into develop.

fedbiomed_gui/server/application.py
```diff
@@ -50,6 +51,8 @@
 jwt = JWTManager(app)
 app.register_blueprint(api)
 app.register_blueprint(auth)
+start_node_for_gui(config.node_config)
+atexit.register(stop_node_for_gui)
```
srcansiz: Hi @alitolga,

You can comment it for now.

fedbiomed/node/node_pm.py
```diff
@@ -0,0 +1,284 @@
...
+        """Spawn a node subprocess.
+
+        Args:
+            config: NodeConfig object (must already be initialised on disk).
+            node_args: Dict of arguments forwarded to ``start_node``.
+            actor: Optional user/source metadata for process state attribution.
+        """
+        # Lazy import to avoid a circular dependency at module load time:
+        # cli imports this module to access the singleton.
+        from fedbiomed.node.cli import start_node  # noqa: PLC0415
```
srcansiz: Why don't you take `start_node` in `NodePM`. NodePM is responsible of starting node and it is used in `CLI` but start node comes form CLI. Adding `start_node` in `NodePM` will resolve circular dependency issue.

fedbiomed/node/node_pm.py
```diff
@@ -0,0 +1,284 @@
...
+        except Exception as e:
+            logger.warning(f"Could not persist node process state: {e}")
+
+    # ------------------------------------------------------------------
+    # Public interface
+    # ------------------------------------------------------------------
+
+    def start(
+        self,
+        config,
```
srcansiz: I think the config should be managed by `NodeProcessManager` as well.  It should instantiate `Node` class and access config from there. What I see now, this responsibility is still in the `start_node`. So, current situation is a bit complex. It is not clear what `NodeProcessManager` does and what `start_node`, `Node` does.

fedbiomed/node/node_pm.py
```diff
@@ -0,0 +1,284 @@
...
+    """Return the current UTC time as an ISO-8601 string."""
+    return (
+        datetime.now(timezone.utc)
+        .replace(microsecond=0)
+        .isoformat()
+        .replace("+00:00", "Z")
+    )
+
+
+class NodeProcessManager:
```
srcansiz, line 162: As I already mentioned in the other comment, this class should responsible of keeping the `start_node` that is implemented in the `cli.py`.

It should be also aware of it is instantiated for which node. So while initializing, it should take component path, or component config, or node class, and it should be instantiated in `cli.py` not in `node_pm.py`.  OR in `application.py` if it is used for GUI.

srcansiz (commented):
Hi @alitolga,

Thank you for the PR, I left some comments that needs to addressed. 

- I think we should first have version that `NodeProcessManager` used in CLI, and not combining GUI start and Node start for now. So you can leave it as it was before, launching node and gui separated as it was before.  
- `NodeProcessManager` responsibilities are no very clear. I think it still requires refactoring to better separate responsibilities between `NodeProcessManager` and `cli`. You can find more details in the comments.

fedbiomed/node/cli.py
alitolga: Hi @srcansiz ,

I created another PR to separate the PRs for issues #1717 and #1718 . In this branch, you are seeing the changes for both of them, that's the main confusion.

fedbiomed/node/node_pm.py
```diff
@@ -0,0 +1,284 @@
...
+    One process at a time. Intended to be used as a module-level singleton
+    so that both the CLI and the API server share the same instance.
+    """
+
+    def __init__(self) -> None:
+        self._process: multiprocessing.Process | None = None
+        self._node_id: str | None = None
+        self._node_name: str | None = None
+        self._state_table: NodeProcessStateTable | None = None
+        self._history_table: NodeProcessStateHistoryTable | None = None
```
srcansiz: Hi @alitolga, 

Thank you for the implementation. I did a few test. You did a very nice job creating `NodeProcessManager` overall, however, I think this implementation is not what we want from `NodeProcessManager` because `NodeProcessManager`:

- Should be able start node process and also allow tracking node processes that are already started. `multiprocessing.Process` creates an internal process and its PID can not be handled outside from the CLI. 

- Should allow taking PIDs as input and allow tracking its status, starting, restarting and stopping them. This is the use case that we will have on the GUI. No matter where the process of node is started Node GUI should be able to see that process. The PID is saved in the DB, but  it is created using `multiprocessing.Process` and `NodeProcessManager`  can not take the PID and perform start, stop or restart operation. 

Therefore, I suggest following changes: 

- [x] Please use `subprocess.Popen` that will execute `start_node_process`. 

- [x] `subprocess.Popen` expect a script or command to execute. It means that  you will have to find a way to pass `config` and `node_args`. An example `Popen` process creation.
```
        process = subprocess.Popen(
            [
                os.sys.executable,
                "-m",
                "fedbiomed.node.node_pm",
            ],
            stdout=subprocess.DEVNULL,
            stderr=subprocess.DEVNULL,
            text=True,
            start_new_session=True,
        )
```
- [x] Implement `foreground` and `background` version of node start. Foreground runs the process and waits for it (it prints the log while waiting for it). Background is more like `docker run -d` it runs the process at the background and saves logs in a log file. It means that for background version you will also provide log file input to keep the logs/outputs of the process. 

- [x] `NodeProcessManager` should be able to take PID and set it as property. This will allow/activate already existing methods like  start/stop for given PID. An example use case: (1) imagine that a node is started using CLI, (2) its PID is saved in the database (3) GUI used process manager again to check the status of the process (4) It stopped the process `foreground` or `background`

- [x] You can keep database operations as it is. PID has to be saved and it status has to be updated at any change. 

Please let me know if you have any question for the points above.

fedbiomed/node/node_pm.py
```diff
@@ -0,0 +1,531 @@
...
+        """Initialize the NodeProcessManager with the given configuration.
+
+        Args:
+            config: Node configuration object that the manager will use to spawn the node subprocess and manage its state.
+        """
+        self._config = config
+        self._node_id: str | None = config.get("default", "id")
+        self._node_name: str | None = config.get("default", "name")
+        self._state_table: NodeProcessStateTable | None = None
+        self._history_table: NodeProcessStateHistoryTable | None = None
```
srcansiz: Hi @alitolga, 

Why don't you instantiate directly `state_table` and `history_table`. You have the config and the path to the db?

fedbiomed/node/node_pm.py
```diff
@@ -0,0 +1,531 @@
...
+
+    # ------------------------------------------------------------------
+    # Internal helpers
+    # ------------------------------------------------------------------
+
```
srcansiz, line 203: Hi @alitolga,

I think a method is required to get the PID from database. This will be important especially to check if the status of process when node started and `NodeProcessManager` instantiated later to check the status of the node. Your implementation relies on PID that can only obtained after the node is started. 

An example use case: 
- I start a node process in a CLI
- I start a ipython kernel and create a `NodeProcessManager` and check the status of the process.
- or I go to GUI and check the status of the running process. (for that NodeProcessManager should be able to take the PID from database and check its status using psutil)

alitolga: Hi @srcansiz ,

I believe the PID should be given to the Process Manager. Because my understanding was that there can be multiple processes in the Node, and the Process Manager should be able to manage all (for example from the GUI).

srcansiz: Well, but `NodeProcessManager` should only manage processes that it has started and these process info is kept in the database. Plus, we need `config` and ~~`node_args`~~ to be able to build `NodePM` which means that it is already associated to a node.

fedbiomed/node/node_pm.py
```diff
@@ -0,0 +1,531 @@
...
+            if state == NodeState.RUNNING:
+                entry["started_at"] = existing.get("started_at") or now
+                entry["stopped_at"] = None
+                entry["exit_code"] = None
+            elif state == NodeState.STOPPED:
+                entry["stopped_at"] = now
+            elif state == NodeState.STARTING:
+                entry["started_at"] = None
+                entry["stopped_at"] = None
+                entry["exit_code"] = None
```
srcansiz: Hi @alitolga, 

You can use `match` `case` for such condition matching : https://www.geeksforgeeks.org/python/python-match-case-statement/

fedbiomed/node/node_pm.py
```diff
@@ -0,0 +1,531 @@
...
+        """
+        # In case we try to start/restart an existing node process.
+        if pid and (
+            self.get_status(pid) == NodeState.RUNNING.value
+            or self.get_status(pid) == NodeState.STARTING.value
+        ):
+            logger.warning(
+                f"Node process 'pid={pid}' is already running. Ignoring start request."
+            )
+            return pid
```
srcansiz: Hi @alitolga, 

What is I already started a process and  executed again `.start()` without `pid` ? Don't you think better to check database to see if node process has started already ?

fedbiomed/node/node_pm.py
```diff
@@ -0,0 +1,531 @@
...
+            logger.warning(f"Could not persist node process state: {e}")
+
+    # ------------------------------------------------------------------
+    # Public interface
+    # ------------------------------------------------------------------
+
+    def start(
+        self,
+        node_args: dict,
+        pid: Optional[int] = None,
```
srcansiz: Hi @alitolga,

This `PID` is not necessary, I think you should keep PID all the time as property and check database if a process already started with this PID. You can do the same for restart. Apparently, PID is only provided for while executing restart.

fedbiomed/node/node_pm.py
```diff
@@ -0,0 +1,531 @@
...
+                "state": state.value,
+                "node_id": self._node_id,
+                "node_name": self._node_name,
+                "action": action,
+                "reason": reason,
+                "actor": self._build_actor(actor),
+                "updated_at": now,
+                "started_at": existing.get("started_at"),
+                "stopped_at": existing.get("stopped_at"),
+                "exit_code": exit_code,
```
srcansiz: Hi @alitolga 

Please type this entry which will simplify its consistency and correctness. 

```
@dataclass
class NodeProcessEntry:
     pid: str, 
     state: ....
     .....
```

fedbiomed/node/node_pm.py
```diff
@@ -0,0 +1,531 @@
...
+            }
+
+            if state == NodeState.RUNNING:
+                entry["started_at"] = existing.get("started_at") or now
+                entry["stopped_at"] = None
+                entry["exit_code"] = None
```
srcansiz: Please use `entry.update({'x':'x'})` which simpler and safer.

fedbiomed/node/node_pm.py
srcansiz: any comment for this one ?

fedbiomed/node/cli.py
```diff
@@ -9,21 +9,17 @@
 import importlib
```
srcansiz, line 9: Hi @alitolga, 

Thank you for the corrections. I think with the new version of `NodeProcessManager` we can extend capabilities of CLI.  For example: Could you please add following options to CLI: 

- `fedbiomed node status`
- `fedbiomed node restart`
- `fedbiomed node stop`


These command can build and execute corresponding methods of `NodeProcessManager`.

fedbiomed/node/node_pm.py
alitolga: Hi @srcansiz ,

The dataclass was there, but apparently I forgot to change it in the implementation. I added it now 👍

fedbiomed/node/node_pm.py
alitolga: I used the dataclass for these updates as well.

fedbiomed/node/node_pm.py
```diff
@@ -0,0 +1,550 @@
...
+        # Start should raise an error if the user is trying to start/restart an existing node process.
+        status = self.get_status()
+        if status == NodeState.RUNNING:
+            logger.warning("Node process is already running. Ignoring start request.")
+            return
+
+        if background:
+            log_path = os.path.join(self._config.root, "log", "node_process.log")
+            os.makedirs(os.path.dirname(log_path), exist_ok=True)
+            output_target = open(log_path, "a")
```
srcansiz, line 389: Hi @alitolga,

Thank you for implementing background option! I think it could be nice to let user know that the process is started at the background. Additionally, following my other comment concerning CLI adding `fedbiomed node status` etc, it can be nice to add this information in the message to show user how they can display and manage this process using cli. What do you think?

alitolga: Hi @srcansiz ,

Of course! I am working on adding the other CLI commands, I can add them with the appropriate messages :)

fedbiomed/node/node_pm.py
```diff
@@ -0,0 +1,550 @@
...
+        Args:
+            node_args: Dict of arguments forwarded to the node subprocess on start.
+            background: Whether to start the node in the background.
+            actor: Optional user/source metadata for process state attribution.
+        """
+        background = background or node_args.get("background", False)
+        self.stop(actor=actor, reason="restart_requested")
+        self.start(node_args=node_args, background=background, actor=actor)
+
+    def get_status(self) -> NodeState:
```
srcansiz, line 560: Hi @alitolga, 

Thank you for extending this method to check PID as well! However I did the following test but it didn't work: 

- Started a node using `fedbiomed node start --background`
- Then, I started a Ipython kernel and instantiated a `NodeProcessManager`
```
config = NodeConfig(<path-to-component-root>)
pm = NodeProcessManager(config)
pm.get_status()
# And it returned stopping
> <NodeState.STOPPING: 'stopping'>
```

It seems that when running on background it changes it status. It may be due to exit on CLI that updates the status of the NodePM.

srcansiz: Additionally, when I execute `stop()` is says it already stopped while it is still running.

srcansiz: I have also realized that the PID is not correct. When I execute `get_status` it shows different PID than the one displayed in `ps aux`.

alitolga: Hi @srcansiz ,

I fixed the cause of your issue, however I discovered another bug.

Apparently, if I try to stop the node process started from another process (e.g. trying to stop CLI started node process from Jupyter Notebook or vice versa), the node process becomes defunct, but it doesn't get removed from the process tree. This causes an issue to keep track of the process state.

I am investigating possible solutions now.

alitolga: I have pushed a potential fix.

srcansiz: Hi @alitolga, 

Thank you very much for the fix. There is stil a minor issue. I start node process then I go to ipython and build `NodeProcessManager` and check  the status. It correctly returns the status as running. I can also stop.  

However, when I start again the node from CLI. If I go back to `NodeProcessManager` in the ipython kernel (same object wtihout rebuilding) it returns the status `STOPPED`. It seems that `get_status` does  understand that the processed stopped and restarted again. You may need to check again reading value from database and checking PID. I think another check is required for: if state is running -> check registered PID using psutil -> if running  -> check old pid -> if not running -> update PID.

fedbiomed/node/node_pm.py
```diff
@@ -0,0 +1,570 @@
...
+            match state:
+                case NodeState.RUNNING:
+                    entry.started_at = existing.get("started_at") or now
+                    entry.stopped_at = None
+                    entry.exit_code = None
+                case NodeState.STOPPED:
+                    entry.stopped_at = now
+
+            self._state_table.update_or_insert_by_id(self._node_id, entry.to_dict())
+            self._history_table.insert(entry.to_dict())
```
srcansiz: Hi @alitolga, 

It could be nice to clear some entries from time to as well. For example, you can remove the entries older than one month. This will avoid filling the db excessively.

fedbiomed/node/node_pm.py
```diff
@@ -0,0 +1,570 @@
+# This file is originally part of Fed-BioMed
```
srcansiz, line 1: Hi @alitolga, 

We may need to document new option of launching node database

- background option 
- where the logs are kept when background option enabled. 
- You can also document process history in the database, and how often entries are removed (see comment below)
- You don't need document `NodeProcessManager` in the main documentation but it could be nice to mention it in the developer documentation -> explaining that it can be used to manage node processes and it can be useful for embedding fedbiomed node in other softwares or PMs.

fedbiomed/node/node_pm.py
alitolga: Thanks for highlighting the issue @srcansiz ! I have pushed a potential fix now.

fedbiomed/node/node_pm.py
alitolga: Just pushed a commit for this as well.

fedbiomed/node/node_pm.py
alitolga: Hi @srcansiz ,

I added the documentation.

docs/developer/managing_nodes.md
```diff
@@ -0,0 +1,156 @@
...
+| reason: `Optional[str]`   | Optional reason for the state change.
+| exit_code: `Optional[int]` | Optional exit code for the process.
+
+Normally the pid and state information shouldn't be given by the user or developer, unless specifically for debugging purposes.
+
+For the `actor` parameter, if the parameter is not set by the user or contains partial empty values, the function `_build_actor()` is used to automatically set up actor information. The CLI currently passes `{"source": "cli"}`.
+
+To retrieve the process state, similarly the helper function `_get_process_state` can be used. It retrieves all the aforementioned information of the node process from the node database, plus the additional datetime information which is set automatically by `_set_process_state` when writing an entry to the database. It raises a `FedbiomedError` if the table, or the state for that node in the table, does not exist. Full list of the dataclass `NodeProcessStateEntry` can be seen in the next section.
```
srcansiz: Hi @alitolga, 

I missed this but in the documentation you describe `private` methods. `_get_process_State` or `_set_process_state` They start with underscore and private methods. Which means that they not intended to be used by user or developer, I mean outside of the class `NodeProcessManager`. 

So I propose to remove them from documentation. I think it too much detail.

docs/developer/managing_nodes.md
```diff
@@ -0,0 +1,156 @@
...
+2. If the PID no longer exists, it records `STOPPED` with reason `process_not_found`.
+3. Otherwise it sends the signal SIGTERM `terminate()` and waits up to 5 seconds.
+4. If the process does not exit, it sends the signal SIGKILL `kill()` and waits up to 5 more seconds.
+5. It records `STOPPED` with the exit code if one is available.
+
+The node subprocess also installs `SIGTERM` and `SIGINT` handlers. When the node is connected, those handlers emit a `node_stopped` security event and send an error message to researchers before exiting.
+
+## Foreground Exit Handling
+
+When the node is run in the foreground, the node process is a child of the foreground process. The foreground process calls `NodeProcessManager._wait()`, which essentially waits until the node process terminates gracefully. After the node terminates, the `_wait()` function will verify the database status to ensure that the node process was killed properly (using `fedbiomed node stop`). To ensure this it will check the database status, and if the database doesn't have a 'STOPPED' status, it will update it as:
```
srcansiz: Same comment for `_wait`. It is private method. You can have to describe `_wait` method, but you can just explain `foreground` and `background` logic. Without referring to private methods.

PR #1785, opened 2026-06-04
Feature/1692 pr2 node state gui

alitolga:
Should be merged after #1784 . PR still under draft.

First implementation of a template page to display node state information.


fedbiomed/node/node_pm.py
```diff
@@ -594,12 +610,13 @@ def get_status(self) -> NodeState:
...
-        if state and state.get("state") == NodeState.RUNNING.value:
-            return state.get("pid")
-        return None
+    def get_process_state(self) -> NodeProcessStateEntry:
+        """Get the current persisted node process state.
+
+        Returns:
+            Current node process state entry.
+        """
+        return self._get_process_state()
```
srcansiz: Hi @alitolga, 

Maybe you can make `get_process_state` public instead of calling same method as private?

fedbiomed_gui/server/routes/node_management.py
```diff
@@ -147,7 +147,7 @@ def node_status():
 def node_process_state():
     """Return current persisted node process state."""
     try:
-        return response(node_process_manager._get_process_state().to_dict()), 200
+        return response(node_process_manager.get_process_state().to_dict()), 200
```
srcansiz, line 150: Hi @alitolga, 

If node is created and gui is started before starting any node process. Node management UI show this error: 

```
Could not get node process state: FB327: Node Process Manager error: No process state found for node_id: NODE_1e8af174-6f01-4a9e-89d9-9f8143911b1b.
```

I think it seems that since node is not started yet there is no record in the database, and it return error. This should not be the case. It should be considered as node is not started, and the view should show a info to say that more info will come once no is started. 

Additionally, this view does not include start and stop. I think you will be adding it later ? 

One last comment about the design of the page. I think something like this rather than a simple table would be much professional. 

Image

srcansiz: Plus, you can put an info icon in the `PID` box and display PID (actor) information in an modal window instead of adding it at the end of the page.

srcansiz: or something like this

image

fedbiomed_gui/ui/src/components/layout/SideNav.js
```diff
@@ -21,6 +22,7 @@ const SideNav= (props) => {
             { key: '4', label: 'Manage Datasets', path: '/datasets/', icon: DataIcon, action: null },
             { key: '5', label: 'Add New Dataset', path: '/datasets/add-dataset/', icon: PlusIcon, action: null },
             { key: '6', label: 'Node Configuration', path: '/configuration/', icon: ConfIcon, action: null },
+            { key: '7', label: 'Node Management', path: '/node-management/', icon: LaunchIcon, action: null },
```
srcansiz: Hi @alitolga, 
I think an icon something like this would fit better.

Image

fedbiomed_gui/ui/src/pages/node-manager/NodeManagement.js
```diff
@@ -0,0 +1,280 @@
...
+            }
+        } catch (error) {
+            setProcessState(null)
+            setProcessStateError(
+                error?.response?.data?.message || 'Could not get node process state'
+            )
+        } finally {
+            setLoading(false)
+        }
+    }, [])
```
srcansiz: Hi @alitolga, 

Thank you for this implementation. However,  as I mentioned also before. I think these API calls should integrate existing API call structure. In the use interface API cals are made from `store/actions` component. I think you can add these calls there and call from the component.    

And please remember that you don't have to use `redux` (global state management, the code use is as `dispatch` to update the global state ) but it could be nice to collect them in one place rather then spreading in all components.

alitolga: Hi @srcansiz ,

I have added the `actions` and `reducer` components as you have suggested.

fedbiomed_gui/ui/src/components/layout/SideNav.js
alitolga: Hi @srcansiz ,

The icon has been replaced now in the latest commit.

fedbiomed_gui/server/routes/node_management.py
alitolga: Hi @srcansiz ,

I updated the GUI according to your suggestions.

fedbiomed_gui/ui/src/pages/node-manager/NodeManagement.js
srcansiz, line 1: Hi @alitolga, 

There is also another possible improvement for better user experience. 

Image

Here while node is running, I can turn on the debug mode as shown below 

Image

However, this does not change anything (which is normal) because node is still running and it has to be restarted. Therefore, It could be nice to add little message below these option radio buttons. If node is runing and user changes radio button state we need to tell them that they have to restart the node to make the changes active.

alitolga: Hi @srcansiz ,

Thank you for the suggestion ! I added the message for the user.

PR #1797, opened 2026-06-18
Feature/1796 display application logs in the gui

### LINKED ISSUE #796: [New issue]: Enhancement for node GUI - node process lifecycle
As a data provider and GUI user, 
I would like to be able to:
- launch the node (`node start`) associated with the GUI
- view the output of the node in the GUI instead of the console
- stop the node

Identifying if a node with the same id/config file has already been started by a CLI process may be too complex and out of the scope of this issue, but if instead it can be done easily it would be a nice addition!

### LINKED ISSUE #1691: Start/Stop/Restart the Node from the GUI
As a user, I would like to be able to manage the more node from the GUI. As mentioned in issue #1610 , an example use case is to change the node configuration from the GUI. To be able to that, a feature to starting, stopping and restarting the node from the GUI is required.

### LINKED ISSUE #1796: Display logs in the GUI
As a user, I would like to be able to see the application logs on the GUI. This should include scenarios where the node is started from the GUI, or the node is started from the CLI in background mode.

To implement this feature: 

- [x] Application logs should always be written into a file as well. This file should rotate daily (or hourly), in order to avoid bloated log files.
- [x] The contents of the file should be displayed accordingly in a dedicated section in the GUI. This includes filtering on the logs to be displayed, to avoid potentially long loading times.

alitolga:
Add rotating application logs to node management

- Add a file handler to the logger for application logs. Keep `add_file_handler` for non-rotating compatibility use cases.
- Add a dedicated application log setup for `<node-root>/log/application.log`. Write node application logs to daily rotating `application.log` files
- Remove the old `node_process.log` file
- Add a separate helper file so that security logs and application logs use the same helper functions for log parsing, filtering, and pagination.
- Expose `/api/node/logs` for filtered application log retrieval
- Add **Application Logs tab** beside Process Details in Node Management
- Add frontend log filters, refresh, and pagination controls
- Add unit tests to logger, node process manager, node management API, and UI for the coverage of log behavior


fedbiomed_gui/server/routes/log_utils.py
srcansiz, line 1: Hi @alitolga, 

This is very nice advance implementation but it is a bit overkill for the process logs. I understand the motivation of showing logs in table formats categorized where it is easy to follow and filter. However, I don't think that it is necessary. My concern is that such complicated parsing actions are not very bullet proof for future changes in logging. It may break quite easily.

I don't think that you need to implement table view in the process logs. It can be plain text where logs are shown line by line as it is in Jupyter notebook for example. 

- This requires only parsing log files line by line and adding the GUI view, no log level or argument parsing only time stamp is enough. 
- By default, web page always shows the latest log, as user scrolls in the scroll view other logs are also loaded. For example: Page shows 100 lines first, then user scrolls down then other 100 lines are loaded it goes like this. So pagination is implemented through scrolling. 
- This will simplify a lot users effort to just see some logs without complex query search.  
- Another important functionality missing is being able download raw log file ;)

alitolga: Hi @srcansiz ,

Thank you for the comment. I agree that parsing and filtering for the application logs may not be necessary.

I changed the view according to your selections, plus also adding the download button for the log file !

Issue #1678, opened 2026-03-24
Node side saving of data for dynamic dataset


mvesin:
Design & implement (node side) saving of *data* for a dynamic dataset
- Probably: general design, implement only for CSV

Issue #1592, opened 2026-01-19
Implement full Fed-ComBat, integrating all part


mvesin:
Implement full Fed-ComBat, integrating all parts (FC code, FA mean/std dev, experiment-in-experiment, dynamic datasets, node state extension)

  * [ ] final refactor
    - for `_FedCombatParameters` create a dataclass/enum for each step, one for the inputs, one for the outputs. The goal is twofold (1) clarify the inputs/outputs of each step for readability (2) check the types of inputs/outputs

  * [ ] Run with notebook
  * [ ] Re-test breakpoints ; containers ; cleaning of temporary datasets  

  * [ ] Compare results with simulation implementation
     * test numerical results, eg comparing to https://gitlab.inria.fr/greguig/ComBat-tuto/ for same model with fixed alea

Issue #1591, opened 2026-01-19
Design & implement (node side) “harmonization process (temporary) state sav


mvesin:
* TO BE DETAILED LATER

PR #1798, opened 2026-06-18
FedCombat integrate all components

mvesin:
**PR description**

PR is now ready to be reviewed and merged :
* addresses and completes #1591 - saving node state between steps
* addresses and completes #1678 - saving harmonized dataset in DB & associated data 
* completes #1569 - support for experiment in experiments
* partially address #1592
    * compare & fix numerical results regarding reference implementation
    * fix & extend unit tests
    * refactors
    * bug fixes
    * provides a (tested) demonstration notebook
* support and test for container mode

Items from #1592 delayed to an additional PR:
* documentation & tutorial
* add end2end test scenario

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:


fedbiomed/common/preproc/_fedcombat_wrapper.py
```diff
@@ -15,22 +15,19 @@ class FedCombatModelWrapper(nn.Module):
...
-            n_phenotypes: Number of phenotypes in the dataset.
+            bias_model: PyTorch model representing the bias effects in Fed-ComBat.
 
         Raises:
             FedbiomedExperimentError: if the provided biological model contains bias parameters."""
         super().__init__()
-        self.n_phenotypes = n_phenotypes
-        self.n_covariates = n_covariates
         self.biological_model = biological_model
+        self.bias_model = bias_model
```
GhilesReguig, line 30: Leaving the bias model to be specified by the user is a bit odd in this context. I don't know if this choice was discussed on your side, but I would have forced a linear module instead. I would think that the ComBat parameters estimated using a non-linear bias model would not behave as expected for this algorithm.

fedbiomed/node/jobs/_fedcombat_jobs.py
```diff
@@ -166,73 +285,230 @@ def _compute_fedcombat_params(self, params) -> dict:
...
             # `dim` is the native PyTorch name for `axis`
             delta2_star_ig = theta_bar_i + 0.5 * torch.sum(
                 (z - gamma_star_ig) ** 2, dim=0
             )
-            delta2_star_ig /= 0.5 * self.n_samples + lambda_bar_i - 1
+            delta2_star_ig /= 0.5 * n_samples + lambda_bar_i - 1
 
         harmonized_data = (
             sigma_hat_g / delta2_star_ig.sqrt() * (z - gamma_star_ig)
         ) + pred
```
GhilesReguig: The harmonized data computed this way is still standardized. It might be better to return the non-standardized version (as in undoing the standardization previously done by multiplying by the std and adding the mean)

GhilesReguig (approved):
Hello @mvesin, 
this looks good to me. I did 2 minor comments, but the implementation should be correct. I will do some more tests as soon as I get a proper testing environment set up. 

Thank you for this implementation !

mvesin:
Hi @leonard2103 
Thank you for raising the point about cleaning the data from the memory after the preprocessing.

1. on the researcher side: I don't think something is needed
     - in `Experiment` the whole `FedCombatPreproc` is de-referenced is we do a `set_preprocessing()` with another value
     - until then we want to keep the `Experiment._fed_preproc` as it may be re-used (if conditions change is re-applying preprocessing is needed)
     - plus most of the potentially memory intensive data is instantiated in `FedCombatPreproc.execute()` via `_FedCombatParameters`, so it is short-lived (only while applying the preprocessing algorithm to a dataset)

2. on the node side: 
    - as `Node.task_manager()` loops forever, the `preproc_job` lives until *the next pre-processing is triggered*. So there may be some useless memory consumption
    - to improve, I will `del` it there https://github.com/fedbiomed/fedbiomed/blob/5ec0c96307da55dea7aac0fc14d59017f21ef5dd/fedbiomed/node/node.py#L683
    - also will add the same for the `FAJob` ~~and `Round`~~ (EDIT] it already exists for Round) to clean a little better the node memory

fedbiomed/node/jobs/_fedcombat_jobs.py
mvesin: Thank you @GhilesReguig for your review !

Very good point, it is now pushed - we were asking ourselves the same.

fedbiomed/common/preproc/_fedcombat_wrapper.py
mvesin: Thank you @GhilesReguig for this important precision !

The idea of making `FedCombatBiasModel` a distinct class provided as a parameter is not at this point to let the user choose the bias model. (S)he currenly has currently no way of doing it.

It was introduced as technical refactoring to avoid having too many hardcoded dependencies, as we also need to know the model structure on the node for inference https://github.com/fedbiomed/fedbiomed/blob/44c577603919d72293349bd8b62f8c8640801ea5/fedbiomed/node/jobs/_fedcombat_jobs.py#L206

I added your point as a comment in the `FedCombatBiasModel` to discourage such future extension (or at least raise a caveat).

tests/dummy_db1
leonard2103, line 1: @mvesin , I think you would want to delete this file

fedbiomed/node/jobs/_fedcombat_jobs.py
```diff
@@ -139,100 +255,281 @@ def _compute_standardized_residuals_params(self, params) -> dict:
...
+        return covariates, phenotypes
+
+    def _delete_temporary_dataset(self, dataset_id: str):
+        """
+        Deletes a temporary dataset from the dataset registry
+
+        Args:
+            dataset_id: the id of the dataset to delete
+        """
+        self._dataset_manager.delete_dataset_by_id(dataset_id)
```
leonard2103, line 403: Does this function need to delete files from memory too?

fedbiomed/node/jobs/_fedcombat_jobs.py
```diff
@@ -139,100 +255,281 @@ def _compute_standardized_residuals_params(self, params) -> dict:
...
+            phenotypes_name: the names of the phenotypes columns
+
+        Returns:
+            a tuple containing the covariates and phenotypes tensors
+        """
+        dataset = TabularDataset(
+            input_columns=covariates_name,
+            target_columns=phenotypes_name,
+        )
+        dataset.complete_initialization(
```
leonard2103: method `complete_initialization` is deprecated. You can use `load` and pass the dict as arguments

fedbiomed/node/jobs/_fedcombat_jobs.py
mvesin: Good pick, thank you @leonard2103 !

Yes, we need to delete files from the disk => I am adding it.

tests/dummy_db1
mvesin: Thank you @leonard2103 

Argh I thought I had already deleted it ;-) 

Thanks for your vigilance !

fedbiomed/node/jobs/_fedcombat_jobs.py
mvesin: Thank you @leonard2103 this is now updated to match new signature.

leonard2103 (approved):
LGTM

PR #1815, opened 2026-06-26
Add Node Configuration to GUI - Feature/1610 node configuration gui

### LINKED ISSUE #1610: Allow changing node configuration dynamically from GUI
Currently, node configuration can only be changed from the `config.ini` file located in `<fbm-node-name>/etc/`. However, some node configuration values can be updated from the user interface, which will simplify node governance from the GUI and allow configuration to be changed dynamically.

There are several ways of integrating it:

- [ ] Option 1: Keep node configuration in the database dynamically, and make the `Config` class check the database first and return the result from the database. This will not require the node to be restarted.
- [ ] Option 2: The GUI modifies/re-writes the config file if the config is changed from the GUI. The Config class keeps track of file change history, and if the last file change date differs from the one that Config initiated, it refreshes itself with the new configuration. This guarantees the consistency of the config where `Database` option can result different values in DB and Config file. 
- [ ] Option 3: This option is a combination of the two options above. At each launch, the node updates the database to keep it consistent with the config file. The Config class always checks the database first and returns the requested config parameter. Once a config is changed from the GUI, it updates the config file and refreshes the DB record by reading the new configuration file. This solves the consistency problem of the database approach.

alitolga:
**Should be merged after #1796 !**

closes #1610 

Changes introduced:

- Modify configuration fields from the UI, under the Node Configuration page. Save/Write the modified changes to the config.ini file.
- Add both options Save and Save&Restart.
- Warn the user if the changes in the file has not been taken into effect yet. (Restart required)
- Warn the user if the config.ini file content has been changed while the node has been running.


fedbiomed/node/config.py
```diff
@@ -80,6 +139,71 @@ def add_parameters(self):
             "port": os.getenv("FBM_RESEARCHER_PORT", "50051"),
         }
 
+    @staticmethod
+    def _infer_gui_field_type(value: str) -> str:
```
srcansiz: Hi @alitolga, 

If this is specific to GUI or API please implement this in API section, and keep core config operation in the core.

fedbiomed/node/config.py
```diff
@@ -80,6 +139,71 @@ def add_parameters(self):
...
+        if normalized in {"true", "false", "1", "0", "yes", "no"}:
+            return "boolean"
+
+        try:
+            int(value)
+            return "integer"
+        except ValueError:
+            return "string"
+
+    def get_gui_config_sections(self) -> Dict[str, Dict[str, Any]]:
```
srcansiz: Same comment for this function as well. It seems to me that they belong to fedbiomed_gui/server more than core config. And please use more generic naming rather than specifying GUI.

fedbiomed_gui/ui/src/pages/node-manager/Configuration.js
srcansiz, line 1: Hi @alitolga, 

I think this page can be moved to `Node Management` entirely as a new tab. There is no need to keep it seperated from Node management page.

fedbiomed_gui/server/routes/node_management.py
```diff
@@ -37,6 +45,400 @@ def _actor_from_request() -> dict:
...
+                schema,
+            )
+
+    return section, updates, normalized_base_values, force
+
+
+def _config_update_conflicts(
+    section: str,
+    updates: Dict[str, bool | int | str],
+    base_values: Dict[str, bool | int | str],
```
srcansiz, line 303: Hi @alitolga, 

Base values are the values to shown user at the moment that display the page. However, if we provide an only API (REST) in the future users will not have initial (base) config values. They are going to directly set config values by providing section values. 

What you implemented is very nice. It checks if the config is updated in the mean time. However, I suggest to make it optional for the cases where user integrate an API to simplify override exiting config in any case.

srcansiz (changes requested):

fedbiomed/node/config.py
alitolga: Hi @srcansiz ,

Thank you for the comment. I relocated and renamed both functions.

fedbiomed_gui/ui/src/pages/node-manager/Configuration.js
alitolga: Hi @srcansiz ,

I moved the configuration page as a tab under node management. I removed the old configuration page and the old route from the side bar, and I changed the home page link to node management.

fedbiomed_gui/server/routes/node_management.py
alitolga: Hi @srcansiz ,

I added an `is not None` check to the necessary functions. The user does not have to provide initial/base config values in the request now.

fedbiomed_gui/ui/src/pages/node-manager/Configuration.js
srcansiz, line 1: Hi @alitolga, 

It may make sense to take this file and put it under `node-manager` since it displayed other that page.

alitolga: Hi @srcansiz,

I moved the `Configuration.js` file under `node-manager` folder.

PR #1827, opened 2026-07-03
Feature/1823 grpc mtls

leonard2103:
## This PR changes:

`+9611 / −3794` across 52 files, but only ~2500 lines are real source change. A
large share of the test churn is a `unittest` → `pytest` conversion that rewrites
tests without changing what they assert (`tests/test_node.py` goes 771 → 702 LOC
but shows 1267 diff lines).

| Group | Files | Lines | Suggested reading |
|---|---|---|---|
| **mTLS core** | `common/certificate_manager.py` (357→943), `transport/client.py` (799→1323), `transport/server.py` (662→905), `node/node.py`, `researcher/requests/_requests.py`, `node/config.py`, `researcher/config.py`, `common/config.py` | ~1900 | Line by line — this is the change |
| **CLI / config surface** | `common/cli.py` (535→750), `common/utils/_config_utils.py`, `common/constants.py`, `cli.py` | ~470 | Line by line — contains breaking signature changes in `fedbiomed.common.utils` |
| **Unrelated to mTLS** | MNIST `download` default flip (4 files), `common/db.py` log redaction, `common/logger.py` field rename, tmpdir fixes in `test_torchnn.py` / `test_csv_reader.py` | ~130 | Self-contained; can be reviewed independently of the mTLS work |
| **Tests** | 17 files | ~5000 | Split first: `test_transport_mtls.py` (new, +1188), `test_certificate_manager.py` (286→1249), `test_common_cli.py` (314→826), `test_transport_client.py` (755→1470) are new coverage. `test_node.py`, `test_requests.py`, `test_config.py`, `test_common_utils.py`, `test_transport_server.py` are mostly reformat — compare the set of `def test_*` names rather than reading the diff |
| **Docs** | new `mutual-tls.md` (+390), deleted `secagg/certificate-registration.md` (−112), `matrix.md`, `troubleshooting.md`, `mkdocs.yml` | ~570 | Skim |

### Points that need a reviewer's opinion

1. **Config version bumped without a migration entry.** `__node_config_version__`
   2.3.0 → 2.4.0 and `__researcher_config_version__` 3.1.0 → 3.2.0, but neither
   `migrate()` gained an `[authentication]` clause. Every reader uses
   `fallback="False"`, so existing configs keep working — so is the bump needed?
2. **Certificates generated before this branch have no SAN.** `ResearcherConfig`
   moves from `subject={"CommonName": host}` to `san=[host]`, and the code now
   raises on a certificate without SAN names. Existing researcher certificates
   need regeneration on upgrade.
3. **`GRPC_VERBOSITY=ERROR` is set at `import fedbiomed`** (top-level
   `__init__.py`). Required because declearn imports `grpc` before
   `fedbiomed.transport`, but it is a process-global env mutation on package import.
4. **`fedbiomed/cli.py` rejects unknown args via `cli.parser.parse_args()`**,
   re-parsing `sys.argv` for its `SystemExit` side effect.

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:


fedbiomed/common/config.py
```diff
@@ -295,13 +296,20 @@ def generate(self, id: Optional[str] = None) -> None:
...
+            mtls_db_path = os.path.join(
+                self.root, CERTS_FOLDER_NAME, f"{DB_MTLS_BASENAME}.json"
             )
+
+            self._cfg["default"]["db"] = os.path.relpath(db_path, config_dir)
             self._cfg["syslog"] = {"enable": "False"}
+            self._cfg["mtls"] = {
+                "enabled": "False",
+                "db": os.path.relpath(mtls_db_path, config_dir),
+            }
```
srcansiz: Hi @leonard2103, 

I propose to use component database rather than creating another json. Both researcher and node can use their already existing dbs? what do you think?

fedbiomed/node/node.py
```diff
@@ -146,6 +144,54 @@ def __init__(
...
+                certificates are missing.
+        """
+        credentials = ResearcherCredentials(
+            port=self._config.get("researcher", "port"),
+            host=self._config.get("researcher", "ip"),
+            certificate=self._config.get("researcher", "certificate", fallback=None),
+        )
+
+        if not is_mtls_enabled(self._config):
+            return credentials
```
srcansiz: node config was not containing `certificate` of researcher, therefore, node was trusting any certificate provided by gRPC. However, now, there is mTL mode. If researcher certificate is present in the db and mTL is not enabled `credentials` will return this certificate. While this may work if researcher certificate is correct but if researcher has different certificate than the one in node config and mTLS mode is not activated this will fail. 

I would suggest first check mTLS, if enabled provide certificate registered. If not give None, meaning that registered certificate won't be taken into account.

fedbiomed/node/node.py
```diff
@@ -146,6 +144,54 @@ def __init__(
...
+
+        etc = os.path.join(self._config.root, CONFIG_FOLDER_NAME)
+        private_key = read_file(
+            os.path.join(etc, self._config.get("certificate", "private_key"))
+        )
+        certificate_chain = read_file(
+            os.path.join(etc, self._config.get("certificate", "public_key"))
+        )
+
+        certificate_manager = CertificateManager(db_path=mtls_db_path(self._config))
```
srcansiz: certificates should be kept in component's db. This will simplify also integrating it in the GUI.

fedbiomed/node/node.py
```diff
@@ -146,6 +144,54 @@ def __init__(
...
+        if not researcher_certificates:
+            raise FedbiomedCertificateError(
+                f"{ErrorNumbers.FB619.value}: Mutual TLS is enabled but no researcher "
+                "certificate is registered. Please register the researcher certificate "
+                "with `fedbiomed node certificate register`."
+            )
+
+        credentials.mtls = True
+        credentials.private_key = private_key.encode("utf-8")
+        credentials.certificate_chain = certificate_chain.encode("utf-8")
```
srcansiz: These two does not belong to `ResearcherCredentials` it is node credential, better to keep them in different places. For example,  Node reads it and passes to `GRPCServer`? An implementation of `NodeGrpcCredentials` ?

fedbiomed/researcher/requests/_requests.py
```diff
@@ -351,14 +361,40 @@ def __init__(
...
+                    "No node will be able to connect until certificates are registered "
+                    "with `fedbiomed researcher certificate register`."
+                )
+            for party_id, expiry in certificate_manager.expiring_certificates(
+                30, ComponentType.NODE.name
+            ):
+                logger.warning(
+                    f"Node certificate `{party_id}` expires on {expiry:%Y-%m-%d}; "
+                    "register an updated certificate to avoid connection failures."
+                )
```
srcansiz: Very good idea to control this!!

fedbiomed/researcher/requests/_requests.py
```diff
@@ -351,14 +361,40 @@ def __init__(
...
+            trusted_node_certificates = "\n".join(node_certificates).encode("utf-8")
```
srcansiz: Here as you can see certificates are join one an other. Maybe I am wrong but once server is started you can not update these certificate, you will have to restart it. 

I don't know how reliable but I came across this implementation. 

```
    dynamic_creds = grpc.dynamic_ssl_server_credentials(
        cert_provider.get_config_callback(),
        require_client_authentication=True,
    )
```

https://grpc.github.io/grpc/python/grpc.html#grpc.dynamic_ssl_server_credentials

And this is certificate config provider. 

```
class CertificateConfigProvider:
    """
    Holds the latest certificate configuration and exposes a callback
    that grpc calls whenever it needs the current cert state
    (e.g. on new connection attempts).
    """
    def __init__(self):
        self._lock = threading.Lock()
        self._config = build_certificate_configuration()

    def refresh(self):
        """Call this whenever a node cert is added/removed/updated."""
        with self._lock:
            self._config = build_certificate_configuration()
            print("[CertProvider] Certificate configuration reloaded.")

    def get_config_callback(self):
        # This matches the signature grpc expects:
        # callback() -> (bool, ssl_server_certificate_configuration) or just config
        def _callback():
            with self._lock:
                return self._config
        return _callback

    cert_provider = CertificateConfigProvider()

```

I don't know it maybe overkill but once we separate researcher-backend and front-end being able to register and update certificate without restarting server could be nice. However, what you have implemented seems decent for the current use case.

fedbiomed/transport/client.py
```diff
@@ -79,6 +84,46 @@ def is_server_alive(host: str, port: str):
...
+    context.check_hostname = False
+    context.verify_mode = ssl.CERT_NONE
+    try:
+        with socket.create_connection(
+            (host, int(port)), timeout=GRPC_CLIENT_CONN_RETRY_TIMEOUT
+        ) as sock:
+            with context.wrap_socket(sock, server_hostname=host):
+                return False
+    except ssl.SSLError:
+        return True
```
srcansiz: Very smart way of checking SSL requirement!

fedbiomed/common/config.py
leonard2103: @srcansiz, I agree with your idea. I will be pushing changes for this.

fedbiomed/node/node.py
leonard2103: @srcansiz,  with mTLS off, `credentials.certificate` is overwritten by the live-fetched server cert at connect. Registering a certificate into the node's cert DB has no effect on this path — it only matters when `[mtls] enabled = True`.

fedbiomed/common/certificate_manager.py
srcansiz, line 792: Hi @leonard2103, 

After this change it fails when component path is given with relative path  `../node..` in the command `fedbiomed component create -c node -p ... `. It may be necessary to conver component path to absolute in `fedbiomed/cli.py", line 154, in create component.initiate(alias=args.name, root=component_path)`.

docs/user-guide/deployment/deployment.md
```diff
@@ -14,7 +14,7 @@ Fed-BioMed provides ready-to-use Docker images for each component (`fedbiomed/no
 - **Customized deployments** — use the provided images as a base and extend them with additional packages, custom datasets, or organisation-specific configuration (see [Extending Docker Images](./docker.md#extending-docker-images))
 - **Integration into larger systems** — embed Fed-BioMed nodes or researchers inside an existing infrastructure by building on top of the base images
 
-Plain Docker images do **not** provide VPN-based authentication between components. They are best suited for deployments on trusted networks or for testing purposes.
+Plain Docker images do **not** provide VPN-based authentication between components. They are best suited for deployments on trusted networks or for testing purposes. Where the parties must authenticate each other without a VPN, [mutual TLS](./mutual-tls.md) authenticates them at the gRPC level instead.
```
srcansiz: Hi @leonard2103, 

As discussed yesterday, it may be nice to replace Mutual TLS with Mutual Authentication. And explain that MTLS is to protocol used for mutual authentication. I think the term mutual authentication is more familiar than MTLS.

docs/user-guide/deployment/mutual-tls.md
```diff
@@ -0,0 +1,296 @@
+# Mutual TLS (mTLS) between nodes and researcher
```
srcansiz: As it is mentioned earlier: Mutual Authentication

docs/user-guide/researcher/configuration.md
```diff
@@ -85,7 +85,16 @@ This file contains settings for connecting to a server, managing security certif
 
 ### `[certificate]`
 - **private_key**: The path to the server's private key, which is used for secure communication.
-- **public_key**: The path to the server's public key, stored at `certs/server_certificate.pem`, which is used for encryption in secure communications.
+- **public_key**: The path to the server's certificate, stored at `certs/server_certificate.pem`, which nodes verify the researcher against when they connect.
+
+The certificate is generated when the component is created, issued for the `[server]
+host` configured then plus `localhost` and `127.0.0.1`. Editing `host` afterwards does
+not reissue it, so under [mutual TLS](../deployment/mutual-tls.md), where each node pins
+this certificate, regenerate it for the new host and have every node register it again:
```
srcansiz, line 93: I would vote for having this optional on the server and node side. A flag to pass `accept wildcard certicificate`.

docs/user-guide/secagg/certificate-registration.md
srcansiz, line 1: HI @leonard2103,

I am not sure if this certificate registration is relevant here.  I am not even sure if secagg still asks certificates to be registered. I think was mandatory for MPSPDZ but now this tool replaced by native implementation of secret sharing. So, I would advise to check again  secagg certificate registartion and remove this doc if it is not used. This page is not even active in the documentation (see `mkdocs.yml`) which may prove that it is not used for secagg.

docs/user-guide/deployment/mutual-tls.md
```diff
@@ -0,0 +1,296 @@
...
+
+Fed-BioMed encrypts all gRPC traffic between nodes and the researcher with TLS, and the
+recommended production deployment additionally tunnels that traffic through a
+[WireGuard VPN](./deployment-vpn.md), which authenticates the endpoints by key and
+isolates the instance from outsiders. Mutual TLS (mTLS) protects the same channel at the
+application layer rather than the network layer: the researcher requires and verifies
+each node's client certificate, and each node pins the researcher's certificate instead
+of trusting whatever certificate the endpoint presents.
```
srcansiz: We may remove VPN there. I think should not mention VPN. There can be a section in the end for VPN is it is necessary (depending on if we are going to keep VPN or not)

docs/user-guide/deployment/mutual-tls.md
```diff
@@ -0,0 +1,296 @@
...
+
+It is meant for two situations:
+
+- **as an alternative to the VPN**, when Fed-BioMed runs on an infrastructure providing
+  its own network layer — plain Docker images, or components embedded in an existing
+  system — where nothing else authenticates the parties;
+- **alongside the VPN**, because a VPN peer is authenticated as a machine, not as a
+  Fed-BioMed party: an authenticated insider can still spoof another party at the gRPC
+  level. mTLS binds each channel to a registered party identity.
```
srcansiz: Same comment about VPN, no need to mention. Main focus should be Mutual Authentication.

docs/user-guide/deployment/mutual-tls.md
```diff
@@ -0,0 +1,296 @@
...
+
+Either way it covers the "mutual verification of other party identity during gRPC setup"
+item of the [security model](./security-model.md) (mitigating a malicious insider
+man-in-the-middle that spoofs a party inside the network).
+
+!!! info "mTLS is optional and off by default"
```
srcansiz: Please add mandatory for deployment.

docs/user-guide/deployment/mutual-tls.md
```diff
@@ -0,0 +1,296 @@
...
+    When mTLS is disabled, the node fetches and trusts the researcher certificate on
+    connect, and the researcher accepts any node — the deployment then relies on the VPN,
+    or on the surrounding infrastructure, to authenticate the parties. Nothing below is
+    required unless you explicitly enable mTLS.
+
+## What mTLS changes
+
+| | Without mTLS (server-authenticated TLS) | Mutual TLS enabled |
```
srcansiz: Please use `Mutual Authentication`

docs/user-guide/deployment/mutual-tls.md
```diff
@@ -0,0 +1,296 @@
...
+A third-party certificate is registered as-is, so check that it names the researcher's
+deployment address before pinning it.
+
+A node verifies the address it dialled whenever the researcher certificate names it.
+Where it does not, the node verifies the researcher under the first name the certificate
+does carry, which is what lets a certificate issued before the deployment address is
+known still be pinned.
+
+For a researcher reachable under a name its configuration does not hold — a public DNS
+name, a second interface — issue the certificate for those names too:
```
srcansiz: I think this section is unnecessary long. This is not the main focus.

docs/user-guide/deployment/mutual-tls.md
```diff
@@ -0,0 +1,296 @@
...
+pin — delete the extras with `fedbiomed node certificate delete`.
+
+### 3. Turn mTLS on
+
+The config file (`etc/config.ini`) of **both** the researcher and every node must
+carry, adding the section if it is not already there:
+
+```ini
+[mtls]
+enabled = True
```
srcansiz: I think we can replace mtls enabled to 

```
[authentication]
enabled = True
```

docs/user-guide/deployment/security-model.md
srcansiz, line 1: Please also update network matrix documentation, you may be add a new table with mutual authentication over MTLS.

fedbiomed/common/dataset/_image_label_dataset.py
```diff
@@ -111,15 +111,16 @@ def load(
...
         Args:
             root: path to the dataset root
             to_format: format associated to expected return format
             train: if true then train files are used
-            download: if true then downloads and extracts the files if they do not exist
+            download: if true then downloads and extracts the files if they do not
+                exist. Off by default: a deployed dataset is already under `root`.
         """
         self.to_format = to_format
         self._init_controller(root=root, train=train, download=download)
```
srcansiz, line 126: Why this change was necessary ?

fedbiomed/common/dataset_controller/_mnist_controller.py
srcansiz, line 1: This change is not related to MTLS no ?

fedbiomed/common/utils/_config_utils.py
```diff
@@ -88,21 +91,28 @@ def get_component_certificate_from_config(config_path: str) -> Dict[str, str]:
         config_path: Path where config file is located.
 
     Returns:
-        Certificate object that contains  component type as `component`, party id `id`, public key content
-            (not path)  as `certificate`
+        Certificate object that contains  component type as `component` (uppercase),
+            party id `id`, public key content (not path)  as `certificate`
```
srcansiz: The term `party` is something remaining for SecAgg. I don't know if it make sense in the context of certificate and TLS. I would use `component id`

fedbiomed/common/certificate_manager.py
```diff
@@ -182,20 +654,102 @@ def register_certificate(
...
+                operation="certificate_registration_rejected",
+                status="failure",
+                party_id=party_id,
+                component=component,
+                registering_component=registering_component,
+                certificate_path=certificate_path,
+                error_type=type(exp).__name__,
+                error_message=str(exp),
+            )
+            raise
```
srcansiz: You may need to raise the error with proper message. This will not provide anything during the debug.

fedbiomed/transport/client.py
srcansiz, line 1: Hi @leonard2103, 

We also need to keep a record in the database to say that node is connected to a researcher. This can be done through a callback defined in the `Node.py` once client connect to researcher successfully it updates the record in the db. 

It can be either implemented here or in another branch.

fedbiomed/common/cli.py
```diff
@@ -349,26 +355,60 @@ def print_help(args):
...
+        ]
+
+        # One researcher because a node pins exactly one; two nodes because that is
+        # the smallest federation worth setting up. Nothing is written otherwise.
+        if len(researchers) != 1 or len(nodes) < 2:
+            CommonCLI.error(
+                "`certificate-dev-setup` sets up one researcher and at least two "
+                f"nodes. Found {len(researchers)} researcher(s) "
+                f"({', '.join(researchers) or 'none'}) and {len(nodes)} node(s) "
+                f"({', '.join(nodes) or 'none'})."
```
srcansiz: I tried executing `fedbiomed certificate-dev-setup` but it didn't function. It couldn't find researcher and node component files.

docs/user-guide/deployment/mutual-tls.md
```diff
@@ -0,0 +1,296 @@
...
+## Development / testing shortcut
+
+When several components run locally under the same Fed-BioMed installation, one command
+replaces steps 1 and 2: it registers in each of them the certificates it must trust — the
+node certificates on the researcher, the researcher's on each node. It expects one
+researcher and at least two nodes:
+
+```shell
+fedbiomed certificate-dev-setup
+```
```
srcansiz: `fedbiomed certificate-dev-setup` is not working. It is also required to indicate where to execute this. Where to put researcher and node component folders etc.

fedbiomed/common/dataset/_image_label_dataset.py
leonard2103: @srcansiz, this was to remove dataset_parameters from mint dataset as it is not in used and I simplified those calls. It is separated from the work in this branch, but still necessary.

fedbiomed/common/dataset_controller/_mnist_controller.py
leonard2103: @srcansiz, not related. I apologize. Still necessary.

fedbiomed/transport/client.py
leonard2103: This will be implemented in another branch.

fedbiomed/common/certificate_manager.py
leonard2103: @srcansiz, this error re-raises a `FedbiomedCertificateError` so it is logged here but raised in the right place. If you had a particular scenario to check let me know.

docs/user-guide/researcher/configuration.md
leonard2103: @srcansiz, this can derive in its own branch as no real control is in place.

docs/user-guide/deployment/security-model.md
srcansiz: Hi @leonard2103, 
Is this comment resolved ?

docs/user-guide/deployment/mutual-tls.md
srcansiz: Hi @leonard2103, 

Is this comment resolved ?

PR #1845, opened 2026-07-09
Feature/1837 remove node management tab and add force option

### LINKED ISSUE #1837: Remove Node Management Tab from GUI until containers are merged and add `--force` option when starting the node to update the node state database.

alitolga:
This PR disables the node management tab from the GUI (until the node and gui containers are merged). It also adds `--force` option when starting the node to override the node state in the database, in case of an inconsistency


alitolga:
I tested again with the containers. The `--force` option can be passed with the environment variable `FBM_NODE_START_OPTIONS` without a problem.

PR #1855, opened 2026-07-27
Add --force option by default on startup to docker containers

alitolga:
Add --force option by default on startup to docker containers to avoid potential bugs with node state database.


mvesin (approved):

mvesin:
@alitolga after merging I realized it is merged in `develop`.

Maybe it would be optimal to update the `release/v6.4.0` branch rather (push this as a commit to the branch, as it is a small fix it does not require review).

The change is then applied both in `develop` and `release/v6.4.0` but this should not create an issue when back-merging  `release/v6.4.0` to `develop`

Thank you !

PR #1885, opened 2026-08-27
fix(serialize): post declearn-2.9 json serialization compatibility

symdec:
# [Fix] Post declearn-2.9 JSON-serialization compatibility

**PR description**  

The goal of this PR is to allow the usage of `declearn>=2.9.0` as a `fedbiomed` dependency.  

DecLearn v2.9.0 introduces breaking changes regarding its serialization utils API.  
More specifically, the functions `json_pack` and `json_unpack` (used by fedbiomed in `fedbiomed.common.serializer.py` and `tests/test_optimizer.py`) have been removed, as the internal packing mechanisms are now private and not part of the public API anymore.
The recommended way to use DecLearn's serialization tools is now through the public API `declearn.utils.serialize.json_serialize` and `declearn.utils.serialize.json_deserialize`.  

Thus, some minor modifications were made in the two mentioned Fed-BioMed files, and proposed in this PR.

**Details on changes**  

1. `fedbiomed.common.serializer.py`:  
Declearn version is checked (using the `packaging.version` lib),  
if version is < v2.9.0:  
- we keep the same behavior as before (importing `json_(un)pack`)  
```
Object ---[json_pack]---> Dict (with `__type__` key)
```

if version is >= v2.9.0:  
- we define an equivalent version of `json_pack` and `json_unpack` behavior using the new public API `json_serialize` and `json_deserialize`.  
 
In order to have the same behavior as `json_pack`, it was needed to load the JSON-string to a dictionary, using `json.loads`. See the following diagram:
```
Object ---[json_serialize]---> str (JSON) ---[json.loads]---> Dict (with `__type__` key)
```

Idem, in a mirror way for `json_unpack` and `json_deserialize`.

Note: In post-v2.9.0, the implementation is less efficient (as we load/dump using `json` lib in an intermediary step), but this implementation seems to be the most readable and minimal, + the additional cost is really minor in my opinion.
Tell me if it is okay for you as well :)

EDIT: If you are not fan of the JSON-serialization trick that is made here (in >= 2.9), another (more efficient) option would be to use directly the new Declearn utils for msgpack serialization when the object that has to be serialized is imported from Declearn (e.g. AuxVar). Let me know if you are more interested by this option.

2. `tests/test_optimizer.py`:  `test_optimizer_15_get_state` only has been updated.  

Declearn version is checked (using the `packaging.version` lib),  
if version is < v2.9.0:  
- use the legacy code, wrapped in the `json_dumps` / `json_loads` functions.

if version is >= v2.9.0:  
- import the new declearn serialization utils `json_serialize` and `json_deserialize` as `json_dumps` and `json_loads`, respectively.

---

**License**

```
```

**Guidelines for PR review**

General:

Specific to some cases:


alitolga:
Hi @symdec ,

Thank you for the PR ! 

I tested it, but unfortunately it seems to break the current serializer logic in Fedbiomed.

The main reason seems to be that `Declearn` introduced the `dump` keyword, instead of `value`, which is confusing the `object_hook` in our (de)serializer logic...

I would suggest a trick like:

```
if Version(declearn.__version__) < Version("2.9"):
    from declearn.utils import json_pack, json_unpack

    def _pack_auxvar(obj: AuxVar) -> dict:
        """Serialize an AuxVar using DecLearn's legacy JSON hook."""
        return json_pack(obj)

    def _unpack_auxvar(obj: dict) -> AuxVar:
        """Deserialize an AuxVar using DecLearn's legacy JSON hook."""
        return json_unpack(obj)

else:
    from declearn.utils.serialize import json_deserialize, json_serialize

    def _pack_auxvar(obj: AuxVar) -> dict:
        """Serialize an AuxVar into an opaque Fed-BioMed envelope."""
        return {
            "__type__": _DECLEARN_AUXVAR_TYPE,
            "value": json_serialize(obj),
        }

    def _unpack_auxvar(obj: dict) -> AuxVar:
        """Deserialize an AuxVar from its opaque Fed-BioMed envelope."""
        return json_deserialize(obj["value"])
```

to modify the wrapper with the fedbiomed `value` keyword. 

Let me know what do you think :)

symdec:
Thanks ! Seems good to me, I have applied the trick :)
It works in my local test for both versions.

srcansiz:
Hi @symdec, 

Thank you for the PR, lately we have upgraded python versions and the versions of other packages. I see that the implementation is for Declearn 2.9 but in our `pyproject.toml` Declearn is limited to `<2.9` .  I was wondering if it is okay to upgrade DecLearn to `~2.9` ?   Are the changes are backward compatible?  @alitolga could you follow on this thread please?

symdec:
Hello @srcansiz @alitolga, 
Yes it is okay to upgrade to `~2.9`, FYI the current PR should allow compatibility with Declearn from 2.6.0 to 2.9.x included (I think 2.7.0 minimum if we consider the dependencies' versions match between Declearn and Fedbiomed, like scikit-learn).

So you can restrict to `2.9` minimum if needed / wanted.

alitolga:
Hi @srcansiz , 
I have tested with `"declearn[torch]  >=2.0,<2.10"` . For me it is working perfectly fine. I will update the pyproject file after merging the PR.

Thanks for the PR @symdec !
Looks good to me to merge. It is just can you update your fork with the current develop with rebase, and then push it. I don't think I have access to rebase and push to your branch.

symdec:
Hi @alitolga, thanks, it should be okay !

srcansiz:
> Hi @srcansiz , I have tested with `"declearn[torch] >=2.0,<2.10"` . For me it is working perfectly fine. I will update the pyproject file after merging the PR.
> 
> Thanks for the PR @symdec ! Looks good to me to merge. It is just can you update your fork with the current develop with rebase, and then push it. I don't think I have access to rebase and push to your branch.

Hi @alitolga, 

Can you please test with multi python version mode of e2e and unit test to make sure that dependencies do not conflict ?

alitolga:
Thanks again for the PR @symdec !

In the end, I have made a few changes, because we have decided to only support the latest version. I am merging it with yout contribution.

Issue #1536, opened 2025-12-10
Endurance tests are failing due to excessive disk usage.


srcansiz:
Endurance tests are failing due to excessive disk usage. These tests are design to force Fed-BioMed experiment using larger models and longer experiments. The test are failing because models on the node side and on the researcher side are saved in the disk before training or aggregation. 

The solution can be adding a testing or `non-artifact` mode in `Config` that can be activated for for endurance test to avoid creating artifact and filling the disk on the CI.

leonard2103:
I identified that there were processes still running from previous executions on the local Mac slave (most of them originating from endurance tests). I cleaned up those processes, which freed the slave. However, the issue persists. Further investigation is needed to determine why the processes are not being properly terminated.

leonard2103:
# Endurance test resource leaks — findings

Investigation into why `endurance-tests.yml` fails. Two independent leaks, one 
memory and one disk, both in product code.  **No fix is applied.** 

## Finding A — the researcher never releases node replies

The researcher process grows about 350 MB per round, linearly, and dies when the
runner is exhausted:

```
11:45:38  mem_avail=14043M  swap=0M     pytest=545M      test_01 starts
11:46:38  mem_avail=5802M   swap=0M     pytest=5234M
11:47:38  mem_avail=1821M   swap=570M   pytest=9896M     swapping begins
11:48:23  mem_avail=539M    swap=2940M  pytest=10389M
11:48:28  ##[error] exit 143 / The runner has received a shutdown signal
```

`test_01` completes all 30 rounds and consumes the entire budget; `test_02` dies
20 seconds in, before finishing round 0. Both tests share one pytest process, so
the ceiling is about 32 rounds for the module rather than 30 each.

The node processes stay flat at 0.5–2.5 GB throughout. This is researcher-side,
and it happens under `FedAverage`, so it is not specific to Scaffold.

### The owner

Allocation site, growing linearly across every sample:

```
152M → 912M → 1673M → 2281M → 2890M → 3650M → 4106M → 4563M
at fedbiomed/common/serializer.py:160
```

which is the numpy copy backing every deserialized torch tensor:

```python
if objtype == "torch.Tensor":
    data, dtype, shape = obj["value"]
    array = np.frombuffer(data, dtype=dtype).reshape(shape).copy()   # line 160
    return torch.from_numpy(array)
```

The reference chain from a retained 144 MB `fc1.weight`:

```
tensor (4096, 9216) 144M
[...]

PR #1893, opened 2026-09-11
Fix memory leak in node agent. Modify unit tests to the new implementation

alitolga:
Changes done:

- Merge `flushed_request_ids` array and `stopped_request_ids` array into one `closed_request_ids` array, in order to only track if a request is done or not. Done indicates any state of termination: Finished succesfully, timed out, interrupted...
- Adapt the unit tests according to the new behaviour.

The memory accumulation during endurance tests seem to have stopped, and all unit and e2e tests are passing on my local machine. I am not closing the issue #1536 yet, as I believe the disk usage should be investigated as well.


leonard2103 (approved):
LGTM

alitolga:
Thanks for the review @leonard2103 !

Nice addition for when `on_reply is None`. I am merging with your changes.

PR #1899, opened 2026-09-14
Feature/1536 disk usage endurance tests fix

alitolga:
Change model parameter saving logic so that the parameters are not saved to disk unless one of the `retain_full_history` or `save_breakpoints` flag is set to True. This PR aims to optimize disk usage.

**Should be merged after #1893**


srcansiz:
Hi @alitolga, 
It seems that there are conflicts in this branch with the one previously merged. Can you please fix them?

